PR #3, opened 2024-12-12
feat: get basic Sphinx docs working locally

benhoyt:
Just a very basic version of the rendered docs. Local for now; can always push up to readthedocs-hosted.com later.


docs/requirements.txt
```diff
@@ -0,0 +1,170 @@
+# This file was autogenerated by uv via the following command:
+#    uv export --group docs
```
dimaqq, line 2: I like this!

PR #23, opened 2025-01-27
fix: add additional params to deploy()

benhoyt:
This PR adds additional arguments for the `deploy` method -- all I think people will use (and possibly more). Design decisions:

- Make the parameter names match the CLI argument names 1:1, even when a plural might make more sense.
- In cases where a str is likely or natural, but a list can also be provided, allow `str | Iterable[str]`.

Also:

- Fix config values when the type is `bool` (should be Go-style lowercase "true" and "false").
- Add docstring "Args:" sections to most methods.
- Simplify `__repr__`.

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -72,7 +72,7 @@ class Juju:
             args.extend(['--controller', controller])
         if config is not None:
             for k, v in config.items():
-                args.extend(['--config', f'{k}={v}'])
+                args.extend(['--config', _format_config(k, v)])
 
         self.cli(*args, include_model=False)
         self.model = model
@@ -104,34 +104,56 @@ class Juju:
         charm: str | os.PathLike,
         app: str | None = None,
         *,
+        attach_storage: str | list[str] | None = None,
         base: str | None = None,
         channel: str | None = None,
         config: dict[str, bool | int | float | str] | None = None,
+        constraints: dict[str, str] | None = None,
+        force: bool = False,
         num_units: int = 1,
-        resources: dict[str, str] | None = None,
+        resource: dict[str, str] | None = None,
         revision: int | None = None,
+        storage: dict[str, str] | None = None,
+        to: str | list[str] | None = None,
         trust: bool = False,
-        # TODO: include all the arguments we think people will use
     ) -> None:
         """Deploy an application or bundle."""
         args = ['deploy', charm]
         if app is not None:
             args.append(app)
 
+        if attach_storage is not None:
+            if isinstance(attach_storage, str):
+                args.extend(['--attach-storage', attach_storage])
+            else:
+                args.extend(['--attach-storage', ','.join(attach_storage)])
         if base is not None:
             args.extend(['--base', base])
         if channel is not None:
             args.extend(['--channel', channel])
         if config is not None:
             for k, v in config.items():
-                args.extend(['--config', f'{k}={v}'])
+                args.extend(['--config', _format_config(k, v)])
+        if constraints is not None:
+            for k, v in constraints.items():
+                args.extend(['--constraints', f'{k}={v}'])
+        if force:
+            args.append('--force')
         if num_units != 1:
             args.extend(['--num-units', str(num_units)])
-        if resources is not None:
-            for k, v in resources.items():
+        if resource is not None:
+            for k, v in resource.items():
                 args.extend(['--resource', f'{k}={v}'])
         if revision is not None:
             args.extend(['--revision', str(revision)])
+        if storage is not None:
+            for k, v in storage.items():
+                args.extend(['--storage' f'{k}={v}'])
+        if to is not None:
+            if isinstance(to, str):
+                args.extend(['--to', to])
+            else:
+                args.extend(['--to', ','.join(to)])
         if trust:
             args.append('--trust')
 
@@ -226,3 +248,9 @@ def _exception_with_status(
         return exc
     else:
         return exc_type(msg + '\n' + str(status))
+
+
+def _format_config(k: str, v: bool | int | float | str) -> str:
+    if isinstance(v, bool):
+        v = 'true' if v else 'false'
+    return f'{k}={v}'
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -23,11 +23,23 @@ class Juju:
         juju.deploy('snappass-test')
 
     Args:
-        model: If specified, operate on this Juju model. If not specified, use the current model.
+        model: If specified, operate on this Juju model, otherwise use the current Juju model.
         wait_timeout: The default timeout for :meth:`wait` (in seconds) if that method's *timeout*
             parameter is not specified.
-        cli_binary: Path to the Juju CLI binary. If not specified, assumes "juju" is in the PATH.
+        cli_binary: Path to the Juju CLI binary. If not specified, uses "juju" and assumes it is
+            in the PATH.
     """
+
+    model: str | None
+    """If not None, operate on this Juju model, otherwise use the current Juju model."""
+
+    wait_timeout: float
+    """The default timeout for :meth:`wait` (in seconds) if that method's *timeout* parameter is
+    not specified.
+    """
+
+    cli_binary: str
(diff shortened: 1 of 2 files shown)


jubilant/_juju.py
```diff
@@ -104,34 +104,56 @@ def deploy(
...
         channel: str | None = None,
         config: dict[str, bool | int | float | str] | None = None,
+        constraints: dict[str, str] | None = None,
+        force: bool = False,
         num_units: int = 1,
-        resources: dict[str, str] | None = None,
+        resource: dict[str, str] | None = None,
         revision: int | None = None,
+        storage: dict[str, str] | None = None,
+        to: str | list[str] | None = None,
```
james-garner-canonical, line 117: Some contradictory typing thoughts

- `str | list[str]` could just be `list[str]` so that there's just one (`Optional`) input type for each argument
- It doesn't look like we need `list`/`dict` methods specifically, so `list[str]` could probably be `Iterable[str]` and then a `tuple` or `set` of names would work. `dict[str, str]` could be `Mapping[str, str]` though alternative `Mapping` types aren't so common
- Unfortunately if we had `to: Iterable[str]` then a `str` becomes a valid input again, since a `str` is an `Iterable[str]`. So we'd (still) have to handle the `str` case explicitly, otherwise `to='foo'` would give the cli arguments `['--to', 'f,o,o']`
- Even if the type was just `list[str] | None`, we'd still have to special case `str` inputs (to raise an error), to avoid this behaviour at runtime (though at least the type checker would complain)

So maybe they're best as they are now. At least the output type doesn't depend on whether `to` or `attach_storage` is a `str` or a `list[str]`

jubilant/_juju.py
```diff
@@ -104,34 +104,56 @@ def deploy(
         charm: str | os.PathLike,
         app: str | None = None,
         *,
+        attach_storage: str | list[str] | None = None,
```
tonyandrewmeyer, line 107: What's the documentation plan with these? Are people expected to know to just refer to the `juju` help and use jubilant as closely as possible? Or the docstring will have a `for details see `juju help deploy` type message? Or jubilant will end up with a full docstring that details the arguments and (where relevant) return types and possible raises?

jubilant/_juju.py
tonyandrewmeyer: > It doesn't look like we need list/dict methods specifically, so list[str] could probably be Iterable[str] and then a tuple or set of names would work. dict[str, str] could be Mapping[str, str] though alternative Mapping types aren't so common

+1 on `Iterable` and `Mapping` anywhere the "str is Iterable[str]" issue doesn't get in the way.

I'd also rather see `Iterable` than `list` because I think these arguments are really most correctly frozensets in most cases - there's no reason for them to be mutable, and duplicates are mistakes (from memory, ignored by `juju`). I haven't checked, but I assume a type checker will complain if I pass a `tuple` or `frozenset` and probably even a `set` because they don't offer the full `list` API, and I think that should be permissable.

I realise that "str is Iterable[str]" is then problematic as James also points out, but I'd rather we did deal with that than try to work around it.

jubilant/_juju.py
```diff
@@ -104,34 +104,56 @@ def deploy(
...
         """Deploy an application or bundle."""
         args = ['deploy', charm]
         if app is not None:
             args.append(app)
 
+        if attach_storage is not None:
+            if isinstance(attach_storage, str):
+                args.extend(['--attach-storage', attach_storage])
+            else:
+                args.extend(['--attach-storage', ','.join(attach_storage)])
```
tonyandrewmeyer, line 129: For these `str | list[str]` arguments, is the intention that I can provide a single storage as a `str` or multiple as `list[str]`, or am I also able to provide multiple as `str` (comma-separated, like with the `juju` CLI)? I think the latter, looking at the code. Are we allowing that deliberately, or even encouraging that style of usage, or is it just incidental that it will work?

jubilant/_juju.py
```diff
@@ -104,34 +104,56 @@ def deploy(
...
-        if resources is not None:
-            for k, v in resources.items():
+        if resource is not None:
+            for k, v in resource.items():
                 args.extend(['--resource', f'{k}={v}'])
         if revision is not None:
             args.extend(['--revision', str(revision)])
+        if storage is not None:
+            for k, v in storage.items():
+                args.extend(['--storage' f'{k}={v}'])
```
tonyandrewmeyer, line 151:
```suggestion
                args.extend(['--storage', f'{k}={v}'])
```

jubilant/_juju.py
james-garner-canonical: Yes, a type checker will complain about passing a `tuple`, `set`, etc when a `list` is specified.

The fact that permitting a `str` with the current implementation permits you to pass a comma-separated list values as a `str`, which I didn't notice until Tony pointed it out, has swayed me towards recommending that we use `Iterable[str]` for the type, and explicitly test for a bare `str` and raise a `TypeError` at runtime in that case.

jubilant/_juju.py
james-garner-canonical: Since we're already wrapping the `juju` cli, we could call `juju help` to generate the docstrings at runtime :smile:

benhoyt: I was thinking of using `Args:` in the docstring to document each briefly (one sentence), and people could look at Juju CLI help output for more. I'm not sure it'd be worth saying "For details see `juju help X`" on each method -- that could just be stated once in the docs or something.

So yeah, I think a "full docstring" including arguments, return types where relevant, and (interesting) exceptions raised.

jubilant/_juju.py
benhoyt: Yep, agreed this should be `Iterable[str]` -- thanks for the reminder. I've already done that for `apps: Iterable[str]` in the `_helpers.py` functions. So I think args like `to` should be `str | Iterable[str] | None` explicitly, so it's clear a single str is allowed (and then handled in the code as they are now).

I think it's probably okay and convenient to allow either `deploy(... to='3,lxd:5')` and `deploy(..., to=['3', 'lxd:5'])`.

jubilant/_juju.py
benhoyt: I was allowing that deliberately, as I think it's okay and convenient to allow either `deploy(... to='3,lxd:5')` and `deploy(..., to=['3', 'lxd:5'])`, and feels more "1:1 with the CLI". But I could be convinced otherwise -- let's discuss further voice.

benhoyt: Per voice discussion, decision was to use `str | Iterable[str]` for these for now, and see how people find it in the beta release. Can always adjust before the 1.0 final release.

james-garner-canonical: A question about the defaults and about the case of an empty input. Currently the default is `None`, which is explicitly tested for -- `if attach_storage is not None:` then `args` is extended.

If you have `attach_storage=""` or `attach_storage=()` then this will presumably ultimately get you a `CalledProcessError` of some description.

Would it be better to simplify the type to `str | Iterable[str]` and make the default `()`, and then just check for truthiness -- `if attach_storage:` ?

Whether or not the default + typing are changed, I guess it's probably better to test for truthiness rather than passing an empty 'list' of storage to Juju.

jubilant/_juju.py
benhoyt: I've added full docstrings with "Args:" now.

jubilant/_juju.py
benhoyt: I would think `attach_storage=""` would probably be a bug, but I can see where `storage=storages` where `storages` happens to be a empty list/tuple might make sense. Yeah, I think you're right -- I'll update.

benhoyt: I decided that `str | Iterable[str] | None = None` is clearly denotes "optional" better than `str | Iterable[str] = ()`, but I've changed the tests to just `if attach_storage` to handle empty sequences.

jubilant/_juju.py
benhoyt: Aha, good catch!

tonyandrewmeyer (approved):
I haven't actually run this code, but from a design and reading-the-code perspective, it looks great!

james-garner-canonical (approved):

PR #26, opened 2025-02-16
feat: add generated status dataclasses

### LINKED ISSUE #17: Flesh out status types in _types.py
Currently I've only implemented a minimal subset of fields in _types.py (application and unit status). We need to pull in all the other fields.

Ideally there's be a way to verify we've got the field names and types correct based on either the Juju Go code or some actual JSON returned from `juju status`.

benhoyt:
These come from the Go generator in https://github.com/benhoyt/juju/pull/2 (which now has the `is_*` methods added to `UnitStatus` and `AppStatus`).

Also renames `ApplicationStatus` to `AppStatus` for consistency (we're abbreviating "application" to "app" everywhere).

Fixes #17.

### DIFF
--- a/jubilant/_types.py
+++ b/jubilant/_types.py
@@ -159,10 +159,11 @@ class AppStatus:
             raise StatusError(d['status-error'])
         return cls(
             charm=d['charm'],
-            base=FormattedBase.from_dict(d['base']) if 'base' in d else None,
             charm_origin=d['charm-origin'],
             charm_name=d['charm-name'],
             charm_rev=d['charm-rev'],
+            exposed=d['exposed'],
+            base=FormattedBase.from_dict(d['base']) if 'base' in d else None,
             charm_channel=d.get('charm-channel') or '',
             charm_version=d.get('charm-version') or '',
             charm_profile=d.get('charm-profile') or '',
@@ -170,7 +171,6 @@ class AppStatus:
             scale=d.get('scale') or 0,
             provider_id=d.get('provider-id') or '',
             address=d.get('address') or '',
-            exposed=d['exposed'],
             life=d.get('life') or '',
             app_status=(
                 StatusInfoContents.from_dict(d['application-status'])
@@ -270,9 +270,9 @@ class StorageInfo:
     def from_dict(cls, d: dict[str, Any]) -> StorageInfo:
         return cls(
             kind=d['kind'],
-            life=d.get('life') or '',
             status=EntityStatus.from_dict(d['status']),
             persistent=d['persistent'],
+            life=d.get('life') or '',
             attachments=(
                 StorageAttachments.from_dict(d['attachments']) if 'attachments' in d else None
             ),
@@ -324,7 +324,7 @@ class FilesystemAttachments:
 
 @dataclasses.dataclass(frozen=True, kw_only=True)
 class FilesystemInfo:
-    Attachments: FilesystemAttachments
+    attachments: FilesystemAttachments
     size: int
 
     provider_id: str = ''
@@ -337,12 +337,12 @@ class FilesystemInfo:
     @classmethod
     def from_dict(cls, d: dict[str, Any]) -> FilesystemInfo:
         return cls(
+            attachments=FilesystemAttachments.from_dict(d['Attachments']),
+            size=d['size'],
             provider_id=d.get('provider-id') or '',
             volume=d.get('volume') or '',
             storage=d.get('storage') or '',
-            Attachments=FilesystemAttachments.from_dict(d['Attachments']),
             pool=d.get('pool') or '',
-            size=d['size'],
             life=d.get('life') or '',
             status=EntityStatus.from_dict(d['status']) if 'status' in d else EntityStatus(),
         )
@@ -360,10 +360,10 @@ class VolumeAttachment:
     @classmethod
     def from_dict(cls, d: dict[str, Any]) -> VolumeAttachment:
         return cls(
+            read_only=d['read-only'],
             device=d.get('device') or '',
             device_link=d.get('device-link') or '',
             bus_address=d.get('bus-address') or '',
-            read_only=d['read-only'],
             life=d.get('life') or '',
         )
 
@@ -412,6 +412,8 @@ class VolumeInfo:
     @classmethod
     def from_dict(cls, d: dict[str, Any]) -> VolumeInfo:
         return cls(
+            size=d['size'],
+            persistent=d['persistent'],
             provider_id=d.get('provider-id') or '',
             storage=d.get('storage') or '',
             attachments=(
@@ -422,8 +424,6 @@ class VolumeInfo:
             pool=d.get('pool') or '',
             hardware_id=d.get('hardware-id') or '',
             wwn=d.get('wwn') or '',
-            size=d['size'],
-            persistent=d['persistent'],
             life=d.get('life') or '',
             status=EntityStatus.from_dict(d['status']) if 'status' in d else EntityStatus(),
         )
@@ -497,10 +497,10 @@ class NetworkInterface:
         return cls(
             ip_addresses=d['ip-addresses'],
             mac_address=d['mac-address'],
+            is_up=d['is-up'],
             gateway=d.get('gateway') or '',
             dns_nameservers=d.get('dns-nameservers') or [],
             space=d.get('space') or '',
-            is_up=d['is-up'],
         )
 
 
@@ -590,8 +590,8 @@ class ModelStatus:
             type=d['type'],
             controller=d['controller'],
             cloud=d['cloud'],
-            region=d.get('region') or '',
             version=d['version'],
+            region=d.get('region') or '',
             upgrade_available=d.get('upgrade-available') or '',
             model_status=(
                 StatusInfoContents.from_dict(d['model-status'])
@@ -629,10 +629,10 @@ class OfferStatus:
             raise StatusError(d['status-error'])
         return cls(
             app=d['application'],
+            endpoints={k: RemoteEndpoint.from_dict(v) for k, v in d['endpoints'].items()},
             charm=d.get('charm') or '',
             total_connected_count=d.get('total-connected-count') or 0,
             active_connected_count=d.get('active-connected-count') or 0,
-            endpoints={k: RemoteEndpoint.from_dict(v) for k, v in d['endpoints'].items()},
         )
 
 
--- a/jubilant/_types.py
+++ b/jubilant/_types.py
@@ -16,7 +16,7 @@ class FormattedBase:
     channel: str
 
     @classmethod
-    def from_dict(cls, d: dict[str, Any]) -> FormattedBase:
+    def _from_dict(cls, d: dict[str, Any]) -> FormattedBase:
         return cls(
             name=d['name'],
             channel=d['channel'],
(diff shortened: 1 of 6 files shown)


benhoyt:
@tonyandrewmeyer and @james-garner-canonical Not sure the best way to review. Probably no need to review line by line, as this is all generated code, but maybe do a spot check of the different field types. You could also scan over the code generator in benhoyt/juju#2.

jubilant/_types.py
```diff
@@ -1,28 +1,107 @@
...
 import dataclasses
+from typing import Any
+
+
+class StatusError(Exception):
+    """Raised when "juju status" returns a status-error for certain types."""
 
 
-@dataclasses.dataclass
+@dataclasses.dataclass(frozen=True, kw_only=True)
```
james-garner-canonical, line 13: I'm still a little hesitant on `kw_only=True` purely because it makes this Python 3.10+ only. The main complication I think is that anyone who's using e.g. `pyright` with `pythonVersion="3.8"` (like we do in `ops`) may run into trouble with type checking their integration test suite -- though I'm not sure exactly how this shakes out with the various settings around type checking dependencies.

jubilant/_types.py
```diff
@@ -1,28 +1,107 @@
...
+class FormattedBase:
+    name: str
+    channel: str
+
+    @classmethod
+    def from_dict(cls, d: dict[str, Any]) -> FormattedBase:
```
james-garner-canonical, line 19: 1. If these classes will only be exposed for typing rather than for manual construction, should `from_dict` be public?
2. Can we do better than `dict[str, Any]`? In this case it could be as simple as `dict[str, str]`, but if we wanted this to be more public then maybe we should have a `TypedDict` for each class. That seems kind of overkill given the sheer number of classes we're providing (even though the `TypedDict`s could be automatically generated too) ... but `dict[str, str]` seems like a straightforward win here.

jubilant/_types.py
```diff
@@ -1,28 +1,107 @@
...
     workload_status: StatusInfoContents = dataclasses.field(default_factory=StatusInfoContents)
+    juju_status: StatusInfoContents = dataclasses.field(default_factory=StatusInfoContents)
+    leader: bool = False
+    upgrading_from: str = ''
+    machine: str = ''
+    open_ports: list[str] = dataclasses.field(default_factory=list)
+    public_address: str = ''
+    address: str = ''
+    provider_id: str = ''
+    subordinates: dict[str, UnitStatus] = dataclasses.field(default_factory=dict)
```
james-garner-canonical, line 75: We did already discuss this, but I just thought I'd note again that since we have `frozen=True` I think it would make sense to use the `Sequence` type instead of `list` and the `Mapping` type instead of `dict`, to help avoid unintentional mutation, since the type checker can warn you if you end up passing (e.g.) a `UnitStatus.open_ports` around. `list` and `dict` are less verbose though which is nice.

jubilant/_types.py
```diff
@@ -87,13 +216,489 @@ def is_waiting(self) -> bool:
...
+    persistent: bool
+
+    life: str = ''
+    attachments: StorageAttachments | None = None
+
+    @classmethod
+    def from_dict(cls, d: dict[str, Any]) -> StorageInfo:
+        return cls(
+            kind=d['kind'],
+            life=d.get('life') or '',
```
james-garner-canonical, line 273: This is probably an example where, if `from_dict` is public, it would be worth using a `TypedDict` for `d` to communicate that `'kind'` is a required key.

jubilant/_types.py
```diff
@@ -87,13 +216,489 @@ def is_waiting(self) -> bool:
...
+class FilesystemInfo:
+    Attachments: FilesystemAttachments
+    size: int
+
+    provider_id: str = ''
+    volume: str = ''
+    storage: str = ''
+    pool: str = ''
+    life: str = ''
+    status: EntityStatus = dataclasses.field(default_factory=EntityStatus)
```
james-garner-canonical, line 335: I'm curious why these fields are separated by a blank line, and (secondarily) why the `from_dict` code uses a different ordering for the arguments -- I'm guessing perhaps `from_dict` is using the default ordering and this version is more intentional? Maybe a comment as to the nature of the grouping could be helpful, but I guess that will all go in the docstrings.

james-garner-canonical (commented):
Ah few thoughts. I had a look at the codegen but not particularly deeply. If I haven't commented specifically on something here I likely haven't thought about it

jubilant/_types.py
tonyandrewmeyer: I'm assuming the `from_dict` order is arbitrary - I don't think it matters much, particularly for a file that people aren't likely to do much hand-editing in, or even reading of. However, maybe it would be nice to be sorted, alphabetically assuming that's a trivial change in the generator and the file is still in a state where it can be trivially regenerated?

I'm assuming the order in the class is the order in the Go struct and that there is (generally) some sort of reason for that order.

I'm also curious about the blank lines, not just here but in many of the classes. I haven't figured out any pattern there.

tonyandrewmeyer (approved):
Wow, this really is a lot. I definitely agree that hiding most of this away under `jubilant.types` or similar is a good idea.

It feels like it would be nice if it was more selective in the way that the calls only expose the most used methods/arguments and fall back to `juju.cli` for the rest, but dropping down to "and then the rest is a dict[str, any]" doesn't seem ideal. I guess being generated from the Go makes it worthwhile.

What's the plan with documenting these? Some of them are probably obvious enough if you know the Juju CLI, but it is nice if there's an actual docstring - but I assume you wouldn't want to notate every class and attribute by hand. If you add docstrings does that make it hard to re-generate?

I spot checked a couple, but didn't look at the Go generator. I'm trusting that's working correctly (or would otherwise be clearly wrong). I'm also assuming they're all needed - I didn't work through the chain to see that everything is linked to status.

jubilant/_types.py
james-garner-canonical: That's really interesting. I mean it works for us in this case, but doesn't it feel like *something* should complain that `import foo` will explode at runtime if we're checking with Python 3.8 explicitly? I guess it's not `pyright --pythonversion=3.8 bar.py`. I thought maybe `ruff check --target-version=py38 bar.py` would complain, but it doesn't either.

jubilant/_types.py
james-garner-canonical: Oh I think I've figured it out. The class definition has the required arguments (no defaults) first, separated by a blank line from the optional arguments. I'm guessing that `from_dict` has them in codegen order, since it doesn't need to put the required arguments first.

jubilant/_types.py
benhoyt: Per our discussion on Matrix, I think even if people have a charm which needs to deploy with Python 3.8, the integration tests can run with a newer version of Python and use Jubilant. So I think I'll keep it on Python 3.12 for now. If we really run into problems, we can restrict this further later.

jubilant/_types.py
benhoyt: Yeah, I replied to this [here](https://github.com/benhoyt/juju/pull/1#discussion_r1955505938):

> I've thought about this further, and I significantly prefer simple types for this use case (list and dict). They've been around forever and they're here to stay, they're very obvious and common, and they're 1:1 with the JSON types this data is coming from. They're also generated anew every time juju.status() is called, not kept around in a cache, so mutability is less of a concern. Keep It Stupid Simple!
>
> However, I agree about kw_only -- I'll add that. Currently I've got the min Python version set to 3.12 for this lib, though I'm not sure I can get away with that.

jubilant/_types.py
benhoyt: I'll think about whether `from_dict` should be public or not, but I'd rather not add the weight (in LoC) of TypedDicts for all of these. Because this is generated, and because it will be very rare to call it on something that didn't just come from JSON, I don't think the TypedDicts will pay off.

jubilant/_types.py
benhoyt: Yeah, that's right -- dataclasses require the required (no default) fields first, so they have to be first. I thought the blank line separating required from non-required was a nice visual cue. Within each group they're in the same order as the Go struct fields. I'll see if it's straight-forward to order the `from_dict` fields the same way, just to avoid confusion.

jubilant/_types.py
james-garner-canonical: Makes sense. If this will solely be called with dictionaries coming from JSON then I reckon that your choice of `dict[str, Any]` is the most user friendly one

jubilant/_types.py
james-garner-canonical: Yeah having a dev dependency require a higher Python version than the runtime requirements seems like it should be fine -- and `kw_only` is great so it definitely seems worth it given that. I was initially concerned that for jubilant specifically this restricted the Python version we could deploy charms with, but that's clearly incorrect, since that just depends on the charm's specified base(s).

By the way, is there some way to specify at the Python project level, the minimum Python version for your (e.g.) development dependencies, separate from the runtime Python version requirement? Or would users specify the python version in their `tox.ini` (e.g. `[testenv:integration]` sets `basepython=python3.12`)? Or maybe it should just get set in CI (defined in github workflows)?

jubilant/_types.py
benhoyt: 1. That's a good question. I think it's probably worth exposing at least `Status.from_dict` for the top-level one, in case someone wants to parse a status themself. But we could make the others `_from_dict`, at least for now. Maybe we can briefly discuss after daily.
2. I'd rather not go there (and I've replied about TypedDict elsewhere). These are going to be coming from JSON, so `dict[str, Any]` seems fine. We could go narrower in some cases, but I think the simplicity of `dict[str, Any]` for all of them beats the inconsistency of having different types, or complexity of figuring out a narrower type.

jubilant/_types.py
benhoyt: I've updated the code to put the required fields first (before the optional ones) in the `from_dict` methods too.

jubilant/_types.py
```diff
@@ -1,28 +1,107 @@
...
+    since: str = ''
+    version: str = ''
+    life: str = ''
+
+    @classmethod
+    def from_dict(cls, d: dict[str, Any]) -> StatusInfoContents:
+        if 'status-error' in d:
+            raise StatusError(d['status-error'])
+        return cls(
+            current=d.get('current') or '',
```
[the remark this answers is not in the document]
benhoyt: Yeah, that was the intention. Realistically that's not going to happen, as Go's `omitempty` will omit the field in that case, but I think it's still a reasonable "defensive programming" thing to do.

jubilant/_types.py
benhoyt: Per voice discussion, we've decided to make all the `from_dict` methods private for now (`_from_dict`).

benhoyt:
> It feels like it would be nice if it was more selective in the way that the calls only expose the most used methods/arguments and fall back to juju.cli for the rest, but dropping down to "and then the rest is a dict[str, any]" doesn't seem ideal. I guess being generated from the Go makes it worthwhile.

Yeah, that's my take too. I probably wouldn't have done all these if doing them by hand, but with generating them it seems reasonable.

> What's the plan with documenting these? Some of them are probably obvious enough if you know the Juju CLI, but it is nice if there's an actual docstring - but I assume you wouldn't want to notate every class and attribute by hand. If you add docstrings does that make it hard to re-generate?

The plan is to add docstrings "as needed". I've added a basic docstring to the top-level `Status` one for now. Unfortunately none of the fields have doc comments on the Go side, otherwise we could use those. So this isn't really documented anywhere. I suspect most people will find what they're looking for with the decent typing information, but we'll see.

PR #28, opened 2025-02-19
refactor: move types.py to statustypes.py to avoid confusion with stdlib 

benhoyt:
There's a (somewhat rarely-used) stdlib module named [types](https://docs.python.org/3/library/types.html), so it seems best to avoid confusion with this. It seems better to name it something related to status (the "domain") because that's what it is anyway.

Also updates the dev dependencies (it was the upgrade of ruff that complained about this).

We could also:

- Just add an ignore for this Ruff warning, as it's likely to be used as `jubilant.types` anyway, so probably won't actually cause issues.
- Call it something else like `status` (but that may conflict with `Juju.status` if we end up having top-level helpers).


benhoyt:
Yeah, on reflection I think it probably was fine as it was, because it'd never really conflict with the stdlib.

However, I also think it's better to have a name that's related to status, and `statustypes` is the shortest one I can think of that's not just `status`. Which has the problem of potential conflict with `Juju.status` if we later move functions to the top level -- but maybe we just won't do that? @tonyandrewmeyer what do you think?

james-garner-canonical:
How about `statuses`?

tonyandrewmeyer:
I agree it wasn't really an issue, partly because the standard library module is so rarely used, but also because it's in the jubilant namespace as well.

Do you think there are any other responses from Juju that are complex enough that you'd have a bunch of classes like this that get exported for typing but aren't generally used? I can't think of anything but you'd know the CLI better than I do. If so, that pushes me away from plain "types" so they aren't all in the same module.

I don't like "status" by itself - partly because of the name clash you mention but also because I think the name should imply that we expect these classes to be used for typing and not generally. I'm not a fan of "statuses" for the same reason.

"statustypes" is a bit long, but these won't be used often so I think that's ok. It's very clear, which I like. I like this the most, and "types" second, particularly if it seems unlikely that there will be more non-status ones not at top level.

benhoyt:
Cool, I'll go with `statustypes` for now then. I agree it's a bit long but that that probably doesn't matter. Can always change our minds again before v1.0.

PR #34, opened 2025-02-24
test: add unit tests for existing functionality

### LINKED ISSUE #30: Add unit tests of existing functionality
This should be done before too much longer!

Determine whether we want to 1) mock `subprocess.run`, or 2) inject a fake `juju` script and actually run that using the `cli_binary` parameter. Probably option 1 as it's simpler and faster, and option 2 doesn't really gain us much (and we'll have the integration tests).

benhoyt:
This adds unit tests for all existing functionality (except some of the generated `statustypes` dataclasses, though even many of them are tested when testing `Juju.status`).

I've added a basic mock and pytest fixture for mocking `subprocess.run`, extremely loosely modelled on `Harness.handle_exec`, which is used in most tests and works nicely. I've also added a very simplistic mock for `time.monotonic`  and `time.sleep`, which is used by the `Juju.wait` tests.

As much as I hate the implicitness of of pytest's "automatically imported" `conftest.py`, I've embraced pytest convention and used it...

Also some functional tweaks that came up while testing:

- Change the behaviour of `Juju.destroy_model` to only set `self.model` to None if destroying the current model.
- Seeing we're only supporting Python 3.12+, get rid of pre-3.12 code to support `Exception.add_node`.

Also adds `pytest-cov` with a basic coverage report when running `make unit`.

Fixes #30

### DIFF
--- a/test/unit/conftest.py
+++ b/test/unit/conftest.py
@@ -11,3 +11,4 @@ def run() -> Generator[mocks.Run, None, None]:
     run_mock = mocks.Run()
     with unittest.mock.patch('subprocess.run', run_mock):
         yield run_mock
+    assert run_mock.call_count >= 1, 'subprocess.run not called'
--- a/test/unit/conftest.py
+++ b/test/unit/conftest.py
@@ -7,6 +7,7 @@ import pytest
 
 @pytest.fixture
 def run() -> Generator[mocks.Run, None, None]:
+    """Pytest fixture that patches subprocess.run with a mocks.Run instance."""
     run_mock = mocks.Run()
     with unittest.mock.patch('subprocess.run', run_mock):
         yield run_mock
--- a/test/unit/conftest.py
+++ b/test/unit/conftest.py
@@ -1,5 +1,5 @@
-import unittest.mock
 from collections.abc import Generator
+from unittest import mock
 
 import mocks
 import pytest
@@ -7,8 +7,19 @@ import pytest
 
 @pytest.fixture
 def run() -> Generator[mocks.Run, None, None]:
-    """Pytest fixture that patches subprocess.run with a mocks.Run instance."""
+    """Pytest fixture that patches subprocess.run with mocks.Run."""
     run_mock = mocks.Run()
-    with unittest.mock.patch('subprocess.run', run_mock):
+    with mock.patch('subprocess.run', run_mock):
         yield run_mock
     assert run_mock.call_count >= 1, 'subprocess.run not called'
+
+
+@pytest.fixture
+def time() -> Generator[mocks.Time, None, None]:
+    """Pytest fixture that patches time.monotonic and time.sleep with mocks.Time."""
+    time_mock = mocks.Time()
+    with (
+        mock.patch('time.monotonic', time_mock.monotonic),
+        mock.patch('time.sleep', time_mock.sleep),
+    ):
+        yield time_mock
--- a/test/unit/conftest.py
+++ b/test/unit/conftest.py
@@ -1,0 +1,12 @@
+import unittest.mock
+from collections.abc import Generator
+
+import mocks
+import pytest
+
+
+@pytest.fixture
+def run() -> Generator[mocks.Run, None, None]:
+    run_mock = mocks.Run()
+    with unittest.mock.patch('subprocess.run', run_mock):
+        yield run_mock
--- a/test/unit/mocks.py
+++ b/test/unit/mocks.py
@@ -2,11 +2,22 @@ import subprocess
 
 
 class Run:
+    """Mock for subprocess.run.
+
+    When subprocess.run is called, the mock returns a subprocess.CompletedProcess
+    instance with data passed to :meth:`handle` for those command-line arguments.
+    Or, if returncode is nonzero, it raises a subprocess.CalledProcessError.
+
+    This also asserts that the correct keyword args are passed to subprocess.run,
+    for example check=True.
+    """
+
     def __init__(self):
-        self.commands = {}
+        self._commands = {}
 
     def handle(self, args: list[str], returncode: int = 0, stdout: str = '', stderr: str = ''):
-        self.commands[tuple(args)] = (returncode, stdout, stderr)
+        """Handle specified command-line args with the given return code, stdout, and stderr."""
+        self._commands[tuple(args)] = (returncode, stdout, stderr)
 
     def __call__(
         self,
@@ -18,8 +29,8 @@ class Run:
         assert check is True
         assert capture_output is True
         assert encoding == 'utf-8'
-        assert tuple(args) in self.commands, f'unhandled command {args}'
-        returncode, stdout, stderr = self.commands[tuple(args)]
+        assert tuple(args) in self._commands, f'unhandled command {args}'
+        returncode, stdout, stderr = self._commands[tuple(args)]
         if returncode != 0:
             raise subprocess.CalledProcessError(
                 returncode=returncode,
--- a/test/unit/mocks.py
+++ b/test/unit/mocks.py
@@ -14,6 +14,7 @@ class Run:
 
     def __init__(self):
         self._commands = {}
+        self.call_count = 0
 
     def handle(self, args: list[str], returncode: int = 0, stdout: str = '', stderr: str = ''):
         """Handle specified command-line args with the given return code, stdout, and stderr."""
@@ -30,6 +31,8 @@ class Run:
         assert capture_output is True
         assert encoding == 'utf-8'
         assert tuple(args) in self._commands, f'unhandled command {args}'
+
+        self.call_count += 1
         returncode, stdout, stderr = self._commands[tuple(args)]
         if returncode != 0:
             raise subprocess.CalledProcessError(
--- a/test/unit/mocks.py
+++ b/test/unit/mocks.py
@@ -47,3 +47,20 @@ class Run:
             stdout=stdout,
             stderr=stderr,
         )
+
+
+class Time:
+    """Mock for time.monotonic and time.sleep.
+
+    This is very simplistic: time.monotonic() starts out at 0, and every time
+    time.sleep(x) is called, it increases by x.
+    """
+
+    def __init__(self):
+        self._monotonic = 0
+
+    def monotonic(self) -> float:
+        return self._monotonic
+
+    def sleep(self, seconds: float):
+        self._monotonic += seconds
--- a/test/unit/mocks.py
+++ b/test/unit/mocks.py
@@ -1,0 +1,35 @@
+import subprocess
+
+
+class Run:
+    def __init__(self):
+        self.commands = {}
+
+    def handle(self, args: list[str], returncode: int = 0, stdout: str = '', stderr: str = ''):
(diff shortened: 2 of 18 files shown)


test/unit/test_helpers.py
```diff
@@ -0,0 +1,248 @@
...
+        (jubilant.any_waiting, 'waiting', 'error'),
+    ],
+)
+def test_any_two_apps(any_func: Callable, expected: str, unexpected: str):
+    status_dict = copy.deepcopy(json.loads(SNAPPASS_JSON))
+    status_dict['applications']['snappass-test']['application-status']['current'] = expected
+    status_dict['applications']['snappass-test']['units']['snappass-test/0']['workload-status'][
+        'current'
+    ] = expected
+    # Add another app named "app2" unit one unit named "app2/0"
```
james-garner-canonical, line 182:
```suggestion
    # Add another app named "app2" with one unit named "app2/0"
```

test/unit/test_helpers.py
```diff
@@ -0,0 +1,248 @@
...
+    'all_func',
+    [
+        jubilant.all_active,
+        jubilant.all_blocked,
+        jubilant.all_error,
+        jubilant.all_maintenance,
+        jubilant.all_waiting,
+    ],
+)
+def test_all_type_errors(all_func: Callable):
```
james-garner-canonical, line 227: Suggestion: combine `test_all_type_errors` and `test_any_type_errors`. Rename to `test_type_error_when_apps_str_or_bytes`. Maybe use keyword arguments like `helper(status=None, apps='str')`.

test/unit/test_juju_class.py
```diff
@@ -0,0 +1,29 @@
...
+def test_repr_defaults():
+    juju = jubilant.Juju()
+
+    assert repr(juju) == "Juju(model=None, wait_timeout=180.0, cli_binary='juju')"
+
+
+def test_repr_args():
+    juju = jubilant.Juju(model='m', wait_timeout=7, cli_binary='/bin/juju3')
+
+    assert repr(juju) == "Juju(model='m', wait_timeout=7, cli_binary='/bin/juju3')"
```
james-garner-canonical, line 29: Could be a bit fragile. Do we need to test the `repr`? If so, how about just one test with:
```py
assert repr(juju) == f"Juju(model={juju.model!r}, wait_timeout={juju.wait_timeout!r}, cli_binary={juju.cli_binary!r})"
```

test/unit/test_status.py
```diff
@@ -0,0 +1,122 @@
+import mocks
+
+import jubilant
+
+MINIMAL_JSON = """
```
james-garner-canonical, line 5: Since these are imported elsewhere, maybe they should live in a separate module like `test/unit/statuses`?

test/unit/test_helpers.py
```diff
@@ -0,0 +1,248 @@
...
+        (jubilant.all_maintenance, 'maintenance', 'error'),
+        (jubilant.all_waiting, 'waiting', 'error'),
+    ],
+)
+def test_all_one_app(all_func: Callable, expected: str, unexpected: str):
+    status_dict = copy.deepcopy(json.loads(SNAPPASS_JSON))
+    status_dict['applications']['snappass-test']['application-status']['current'] = expected
+    status_dict['applications']['snappass-test']['units']['snappass-test/0']['workload-status'][
+        'current'
+    ] = expected
```
james-garner-canonical, line 58: What if `SNAPASS_JSON` was a template string, and you called something like:
```py
status_dict = json.loads(SNAPPASS_JSON.format(app_status=expected, unit_status=expected)
```
Also, is it necessary to `copy.deepcopy` here? Does `json.loads` do some funky caching stuff with the objects internally?

test/unit/mocks.py
```diff
@@ -0,0 +1,66 @@
...
+    for example check=True.
+    """
+
+    def __init__(self):
+        self._commands = {}
+        self.call_count = 0
+
+    def handle(self, args: list[str], returncode: int = 0, stdout: str = '', stderr: str = ''):
+        """Handle specified command-line args with the given return code, stdout, and stderr."""
+        self._commands[tuple(args)] = (returncode, stdout, stderr)
```
james-garner-canonical, line 21: I like how explicit this is. I wonder if it would be nice as a context manager so that it's linked to the specific calls that use `subprocess.run` rather than just the test function. Probably fine as is though, I don't think any of the tests ever even call `handle` more than once. But just for fun ...
```suggestion
    @contextmanager
    def handle(self, args: list[str], returncode: int = 0, stdout: str = '', stderr: str = '') -> None:
        """Handle specified command-line args with the given return code, stdout, and stderr."""
        key = tuple(args)
        assert key not in self._commands
        self._commands[key] = (returncode, stdout, stderr)
        yield None
        del self._commands[key]
```
Used like:
```py
with run.handle(['foo', 'bar']):
    thing.foo('bar')
with (
    run.handle(['foo', 'bar']),
    run.handle(['foo' '--cleanup'], returncode=9000),
    pytest.raises(CalledProcessError) as excinfo,
):
    thing.foo('bar', cleanup=True)
assert run.call_count == 3
```

james-garner-canonical (approved):
Nice test coverage -- just some minor comments which may or may not be helpful.

test/unit/test_cli.py
```diff
@@ -0,0 +1,74 @@
...
+    assert exc.cmd == ['juju', 'error']
+    assert exc.stdout == 'OUT'
+    assert exc.stderr == 'ERR'
+    assert (
+        str(exc)
+        == """Command '['juju', 'error']' returned non-zero exit status 3.
+Stdout:
+OUT
+Stderr:
+ERR"""
```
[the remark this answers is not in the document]
benhoyt: Good call -- I went with the `endswith`.

test/unit/test_helpers.py
benhoyt: Thanks, fixed.

test/unit/test_juju_class.py
benhoyt: I think this is fair for `test_repr_defaults`, as the default might change and we're not testing that here, so I've just removed `test_repr_defaults`. But I think `test_repr_args` is okay. It will be fragile in the sense that if we add more (args) to the repr, it'll need to change, but I think that's the idea.

test/unit/test_status.py
benhoyt: Done: `fake_statuses.py`

test/unit/conftest.py
```diff
@@ -0,0 +1,25 @@
+from collections.abc import Generator
+from unittest import mock
+
+import mocks
```
tonyandrewmeyer, line 4: I think this would be better as `from . import mocks` - it would get sorted properly (below `pytest`) and it's clearer that it's something that's in this project and not some third-party "mocks" package.

(Same comment for everywhere that has the import).

test/unit/test_helpers.py
benhoyt: I toyed with the idea of a template string, but didn't love that. It's nice to just have something concrete in most uses of `SNAPPASS_JSON`, and we need to copy apps/units in the two-app tests anyway, so we'd need something else for that.

Instead, I've added two helper functions that tidy this up significantly: `set_app_and_unit_status`, and `add_app`.

You're right about the not needing the `deepcopy` here -- thanks.

test/unit/mocks.py
benhoyt: Yeah, I don't think this is necessary here (if we were exporting it as a more general mock, maybe), and it's better as a test-level pytest fixture, as it fits with pytest and avoids the indentation. You can actually call `handle` multiple times in the test if you want, to update it (you could even call it in wait's `ready` func if you wanted to get clever, but I didn't need to).

test/unit/test_wait.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def test_modified_delay_and_successes(run: mocks.Run, time: mocks.Time):
+    run.handle(['juju', 'status', '--format', 'json'], stdout=MINIMAL_JSON)
+    juju = jubilant.Juju()
+
+    status = juju.wait(lambda _: True, delay=0.5, successes=5)
+
+    assert run.call_count == 5
+    assert time.monotonic() == 2.0
```
tonyandrewmeyer, line 52: Nit: I think it's a little confusing that this is the same as in the `ready_normal` test, except that it's a float. I get that it's because `(3-1)*1 == (5-1)*0.5`, but it took a minute to understand that. Maybe the delay could be 0.75 or something like that instead, so that the result is != 2?

test/unit/mocks.py
james-garner-canonical: I agree that it's not necessary here. I just want to clarify that I think you'd still use the version I suggested as a pytest fixture rather than (e.g.) doing the monkeypatching in the context manager, as that forces you to handle *all* calls to `run` (like you do currently). I just like the idea of specific `handle` setups only existing as long as needed, and automatically un-handling themselves so that if you do need to handle multiple calls in one test, it's clear which call you're expecting where, if that makes sense. But yeah, agreed that it's not necessary here.

test/unit/test_status.py
```diff
@@ -0,0 +1,34 @@
+import mocks
+from fake_statuses import MINIMAL_JSON, MINIMAL_STATUS, SNAPPASS_JSON
```
tonyandrewmeyer, line 2: `fake_statuses` is more obviously something local than `mocks` but I suggest the same thing here: `from .fake_statuses import ...`.

test/unit/conftest.py
```diff
@@ -0,0 +1,25 @@
+from collections.abc import Generator
+from unittest import mock
```
tonyandrewmeyer, line 2: My inclination is to use pytest.monkeypatch when using pytest, rather than using unittest.mock. I'm not sure if that's a general practice or not, though.

test/unit/test_helpers.py
```diff
@@ -0,0 +1,206 @@
+import copy
```
tonyandrewmeyer, line 1: This might be unavoidable, but when I saw the name of this file, I assumed that it was "helpers for the tests", rather than "tests for the _helpers module".

tonyandrewmeyer (approved):
I haven't done a line-by-line review (or actually run the tests locally) but a few small comments.

In terms of the general approach - I really like this. The tests are very clean and readable, and seem like they will easily expand to cover new functionality. I assume these are also very fast since all the subprocessing (and time) is mocked out.

test/unit/conftest.py
benhoyt: Good call, done.

test/unit/test_wait.py
benhoyt: `2 == 2.0`, so I didn't actually need the `.0` there. But I've modified delay to 0.75 to avoid confusion.

test/unit/conftest.py
benhoyt: Fair - done.

test/unit/test_helpers.py
benhoyt: I decided to rename `_helpers.py` to the more specific `_all_any.py`, and `test_helpers.py` similarly. If there are other helpers they can go in specifically-named files too.

PR #35, opened 2025-02-25
feat: add compact Status.__str__ like "juju status" tabular output

### LINKED ISSUE #24: Add compact __str__ for the Status object
This would be used for debugging and logging, rather than printing out the whole JSON object or dataclass dump. Should be compact and similar to or reminiscent of the `juju status` textual output, one line per unit and application, something like this:

```
App  Version  Status  Scale  Charm  Channel  Rev  Exposed  Message
foo                       2                    0  no       

Unit   Workload     Agent      Machine  Public address  Ports  Message
foo/0  maintenance  executing                                  (config-changed) doing some work
foo/1  maintenance  executing                                  (backup database) doing some work
```

benhoyt:
NOTE: this branches off #34, so until that's merged the PR diff includes those changes too.

This adds a custom `Status.__str__`, as the default `dataclasses` one is massive and impossible to read for a structure this large. `str(status)` is logged by `Juju.wait` the first iteration or whenever the status changes after that, and it will be important for debugging CI to a clear, human-readable representation of the current status.

It seems best to model this after `juju status` output (the default, tabular output). It's a little simplified in some cases, but the structure is the same, for example:

```
Model  Controller          Cloud/Region        Version  Timestamp       Notes
tt     microk8s-localhost  microk8s/localhost  3.6.1    12:04:55+13:00

App            Version  Status  Scale  Charm          Channel        Rev  Address         Exposed  Message
snappass-test           active  1      snappass-test  latest/stable  9    10.152.183.248  no       snappass started

Unit              Workload  Agent  Machine  Address       Ports  Message
snappass-test/0*  active    idle            10.1.164.138
```

Fixes #24

### DIFF
--- a/jubilant/statustypes.py
+++ b/jubilant/statustypes.py
@@ -5,6 +5,15 @@ from __future__ import annotations
 import dataclasses
 from typing import Any
 
+try:
+    # Real implementation lives in another file to avoid having to include and
+    # maintain this in the Juju code generator.
+    from . import _status
+except ImportError:
+    # So that this module can be imported outside this package (in the context
+    # of the Juju code generator).
+    _status = None
+
 __all__ = [
     'AppStatus',
     'AppStatusRelation',
@@ -742,3 +751,9 @@ class Status:
                 else ControllerStatus()
             ),
         )
+
+    def __str__(self):
+        """Return a "juju status"-style tabular representation of this status."""
+        if _status is None:
+            return super().__str__()
+        return _status._status_str(self)
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ jubilant.egg-info
 /docs/_build
 /.idea
 /t.py
+/.coverage
--- a/Makefile
+++ b/Makefile
@@ -43,4 +43,4 @@ static:
 # Run quick unit tests
 .PHONY: unit
 unit:
-	uv run pytest test/unit
+	uv run pytest test/unit --cov=jubilant
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ help:
 
 # Run all quick, local commands
 .PHONY: all
-all: fmt lint static unit
+all: format lint static unit
 
 # Build documentation
 .PHONY: docs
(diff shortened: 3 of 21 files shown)


jubilant/statustypes.py
```diff
@@ -5,6 +5,15 @@
...
 
+try:
+    # Real implementation lives in another file to avoid having to include and
+    # maintain this in the Juju code generator.
+    from . import _status
+except ImportError:
+    # So that this module can be imported outside this package (in the context
+    # of the Juju code generator).
+    _status = None
+
```
james-garner-canonical, line 16: Bigger picture alternative, would it make sense to make the generated dataclasses a dependency of the handmade stuff instead? e.g. (I *think* this would work)

private `_statustypes.py`
```py
_status = None
# generated dataclasses with current conditional __str__ implementation
__all__ = [
    # generated dataclass names
]
```
public `statustypes.py`
```py
import _status
import _statustypes
from _statustypes import *  # re-exporting for public use

_statustypes._status = _status
```

james-garner-canonical: A couple of low level ideas for the current approach, not sure if they're better but just sharing.

1. conditionally define `_status_str` here so that the `__str__` function doesn't also require branching.
```py
try:
    from .status import _status_str
   # or from . import _status; _status_str = _status._status_str
except ImportError:
    _status_str = lambda x: super(type(x), x).__str__()
```
2. Instead of `try ... except` you could branch on an environment variable like `CODEGEN_ONLY`.

benhoyt: Those are good ideas -- thanks. However, I've decided we just need to bite the bullet and start modifying `statustypes.py` locally. We'll need to start adding docstrings anyway. That avoids the problem entirely.

benhoyt:
Per Dima's push-back, we've decided against this approach. It's very tricky to re-implement `juju status` tabular output correctly, and we shouldn't try. I'm going to take a "dataclasses repr, but compact and nicely indented" approach in a subsequent PR.

PR #36, opened 2025-02-26
docs: use Canonical documentation starter pack

dwilding:
This PR adds the latest version of [Canonical's documentation starter pack](https://github.com/canonical/sphinx-docs-starter-pack).

I've customised the starter pack with Jubilant-specific config. A few things to note:

- To build the docs, use `make run` in the `docs` directory. I changed the `docs` target in the top-level makefile accordingly.

- The auto-generated docs are located at http://127.0.0.1:8000/reference/jubilant/

- The starter pack brings in a spelling check, which I think we should keep enabled if possible. We can run it using `make spelling` in the `docs` directory. It also runs as a GitHub action. I added some [spelling exceptions](https://github.com/canonical/jubilant/pull/36/files#diff-4d576131bbf34be4af7449c912e3cb8c436fb2efa021fc3a6d39aeb8c2cc46ae). I also had to add some code formatting in the Jubilant source - see [_juju.py](https://github.com/canonical/jubilant/pull/36/files#diff-fdac848d1b388cc3783af07c86ebe9c457693f9448d29cc1c45ea6ece78484be) and [statustypes.py](https://github.com/canonical/jubilant/pull/36/files#diff-6ebcdb1b840a8bdc7045be62cb55e536bd7efb3917afb04b2c7024157a1be293).

**Next steps after merging**

1. Publish the docs to Read the Docs
2. Improve the homepage


dwilding:
It looks like some checks are failing. I'm looking into that now.

dwilding:
All checks passing. See the PR description for details about the spelling check.

.github/workflows/automatic-doc-checks.yml
```diff
@@ -0,0 +1,22 @@
+#
+name: Automatic doc checks
```
benhoyt, line 2: These names / this hierarchy shows up as very long in the GitHub Actions run:

> Automatic doc checks / documentation-checks / Run documentation checks

The others are short like "CI / lint". Can we do a couple of things:

- Move all the docs-related workflows to a single workflow file, probably `docs.yaml` (note that `ci.yaml` is `.yaml` so let's be consistent with that file extension). Then have different jobs in the one file is appropriate.
- Shorten the names, eg "Docs / spelling" (all of these are automatic checks).

.github/workflows/automatic-doc-checks.yml
```diff
@@ -0,0 +1,22 @@
...
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+
+  workflow_dispatch:
+    # Manual trigger
+
+
+concurrency:
```
benhoyt, line 13: Is this section needed? What does it do?

.github/workflows/markdown-style-checks.yml
```diff
@@ -0,0 +1,20 @@
+name: Markdown style checks
```
benhoyt, line 1: What kinds of things does this detect?

.github/workflows/sphinx-python-dependency-build-checks.yml
```diff
@@ -0,0 +1,50 @@
+# The purpose of this workflow file is to confirm that the Sphinx
+# virtual environment can be built from source, consequently documenting
+# the packages required in the build environment to do that.
```
benhoyt, line 3: I'd prefer to drop this workflow until/unless we need it. I think once we have the docs site building on ReadTheDocs, that'll essentially achieve the same thing (ensuring it can build).

.readthedocs.yaml
```diff
@@ -1,13 +1,30 @@
...
 
+# Build documentation in the docs/ directory with Sphinx
 sphinx:
+  builder: dirhtml
   configuration: docs/conf.py
+  fail_on_warning: true
+
+# If using Sphinx, optionally build your docs in additional formats such as PDF
+formats:
+- pdf
```
benhoyt, line 25: What's PDF generation needed for? Is that just part of the starter park that we can't turn off?

.readthedocs.yaml
```diff
@@ -1,13 +1,30 @@
...
 build:
-  os: "ubuntu-24.04"
+  os: ubuntu-22.04
   tools:
-    python: "3.12"
-
-python:
-  install:
-    - requirements: docs/requirements.txt
+    python: "3.11"
```
benhoyt, line 12: Is there a reason 3.11 is needed here? The rest of Jubilant requires 3.12, so I was thinking this should match.

.wokeignore
```diff
@@ -0,0 +1,4 @@
+# the cheat sheets contain a link to a repository with a block word which we
```
benhoyt, line 1: I'm fairly strongly opposed to the use of this tool. My grandfather is [legacy](https://github.com/get-woke/woke/blob/5d52c1541ab7b5823531dd48144197dda342042e/pkg/rule/default.yaml#L45-L49) at this point, but why rub it in? ;-) More seriously, I think it's harmful, even Orwellian, to force change on language in this way. I believe it's easy to see the silliness of it all when the term "grandfathered in" is considered "exclusive".

But I'm fine with the other checks -- spelling and link check. Instead of running [documentation-checks](https://github.com/canonical/documentation-workflows/blob/main/.github/workflows/documentation-checks.yaml), can we run `canonical/documentation-workflows/spellcheck@main` and `canonical/documentation-workflows/linkcheck@main` directly?

benhoyt (commented):
Nice work! I've run it locally and it looks really good. A few comments, mainly about the structure of the GitHub Actions and what's included there.

benhoyt:
One other thing I noticed: `uv` is so much faster than `venv` / `pip` (the starter pack Makefile uses the latter). Be cool to update the starter pack to use `uv` at some point. :-)

.github/workflows/sphinx-python-dependency-build-checks.yml
dwilding: I agree. Done in 464a189

.github/workflows/automatic-doc-checks.yml
dwilding: I agree. I've combined and shortened what I can in our repo, in 464a189. I think it's reasonable now.

.github/workflows/markdown-style-checks.yml
dwilding: This checks things like heading style (eg spacing, punctuation, consistent style within a doc), spacing around code blocks and lists, no undefined links, and alt text for images. I don't have a problem if we decide to remove any rules that get in our way later.

The enabled rules are defined in [.sphinx/.markdownlint.json](https://github.com/canonical/jubilant/pull/36/files#diff-bac1d7da1605a7f4ab558458744681babf3e2a96731f26c7804abc19101fbbe4) - and [here's what each rule means](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md).

.readthedocs.yaml
dwilding: We can turn that off if we don't want to provide PDFs. I've removed it for now in 77a2da0.

.github/workflows/automatic-doc-checks.yml
dwilding: I changed the file, but was your question about the `concurrency` part? That part prevents jobs from different commits running concurrently. So that if we push a new commit to a PR while jobs are still running from the previous commit, the jobs will be stopped and fresh jobs will be started for the new commit.

benhoyt: Ah, that makes sense -- for some reason I thought that was the default.

.wokeignore
dwilding: From discussion: As part of the Documentation objectives, we should consider inclusive language. As an alternative to the automated check, we'll review language as part of the normal PR review process.

benhoyt: Thanks. Can we remove this config file from the PR?

docs/.custom_wordlist.txt
```diff
@@ -0,0 +1,7 @@
+# Leave a blank line at the end of this file to support concatenation
```
benhoyt, line 1: It looks like comment isn't being followed, as there's no blank line at the end of this file. Is that needed, though?

docs/Makefile
```diff
@@ -0,0 +1,179 @@
...
+        allmetrics pa11y pdf-prep-force pdf-prep pdf Makefile.sp vale-install vale
+
+full-help: $(VENVDIR)
+	@. $(VENV); $(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+	@echo "\n\033[1;31mNOTE: This help texts shows unsupported targets!\033[0m"
+	@echo "Run 'make help' to see supported targets."
+
+# If requirements are updated, venv should be rebuilt and timestamped.
+$(VENVDIR):
+	python3 -c "import venv" || \
```
benhoyt, line 53: Would be fun to fix this Makefile to use `uv` (and see how much it speeds up!). And then see if upstream wants to take the changes. :-) I've opened https://github.com/canonical/jubilant/issues/43 to rack.

benhoyt (approved):
Thanks -- approved. Just a couple of very minor comments before we merge (mainly removing the now-unneeded config file from the repo top level).

.wokeignore
dwilding: Ah yes, removed in 86926bc.

docs/.custom_wordlist.txt
dwilding: I see a blank line in the file (GitHub would show an absence of blank line with :no_entry_sign: or similar). Looking at the makefile, the various word lists do get cat'ed and appended, so probably best to keep the blank line.

PR #37, opened 2025-02-26
feat: pretty-print Status dataclass

### LINKED ISSUE #24: Add compact __str__ for the Status object

benhoyt:
This adds a function that pretty-prints a value with special cases for dataclasses, lists, and dicts. The result is still Python code (as `__repr__` should be), so will `eval` back to itself.

Per the docstring:

> If *value*'s fields or sub-values are simple scalar types and the result fits in a limited number of characters, use a single-line format. Otherwise, split onto multiple lines and indent each line.

Use this for `Status.__repr__` and `Status.__str__`, so that when status values are logged, it's nice and readable, but not too verbose. It looks pretty decent to me.

This PR also renames `StatusInfoContents` to `StatusInfo` for brevity. It's used a lot and deserves a shorter name, despite the long name in the Go struct definition. `StatusInfoContents` is an unnecessarily long name, and this
will just make the output a bit nicer to read.

Fixes #24.


test/unit/test_pretty.py
```diff
@@ -0,0 +1,144 @@
...
+          leader=True,
+          address='10.1.164.138',
+          provider_id='snappass-test-0',
+        ),
+      },
+    ),
+  },
+  controller=ControllerStatus(timestamp='12:04:55+13:00'),
+)
+"""[1:-1]
```
james-garner-canonical, line 63: I always wonder how to handle this kind of thing. Is this particular solution a `go` idiom? I think the first time I recall seeing it used like this is in some `go` code. An alternative is to have the first line be `"""\` with the trailing backslash eating the first newline, and then just put the closing `"""` on the final line -- `)"""` in this case.

jubilant/_pretty.py
```diff
@@ -0,0 +1,59 @@
+import dataclasses
+from typing import Any
+
+_MAX_VALUE = 150
+_INDENT = 2
+
+
+def _dump(value: Any, indent='') -> str:
```
james-garner-canonical, line 8: Does `object` work instead of `Any` here, since all the specific stuff is behind type checks?

jubilant/statustypes.py
```diff
@@ -5,6 +5,8 @@
 import dataclasses
 from typing import Any
 
+from . import _pretty
```
james-garner-canonical, line 8: No concerns with this local import preventing the easy use of `statustypes` in codegen?

james-garner-canonical (approved):
Looks good to me, just a few questions.

test/unit/test_pretty.py
benhoyt: Good call -- I'd forgotten you could use `\` in `"""` string. I don't know that it's a particular Go idiom, but yeah, your suggestion is better here.

jubilant/_pretty.py
benhoyt: It does, thanks!

jubilant/statustypes.py
benhoyt: Nah, I decided we could just bit the bullet and start modifying that file locally. We'll be adding docstrings and other things anyway. If we update the codegen (which is pretty stable now) we can patch it with a diff (manually or automatically) rather than copying the file.

jubilant/statustypes.py
```diff
@@ -742,3 +744,11 @@ def _from_dict(cls, d: dict[str, Any]) -> Status:
...
             ),
         )
+
+    def __repr__(self):
+        """Return a pretty-printed version of the status."""
+        return _pretty._dump(self)
+
+    def __str__(self):
+        """Return a pretty-printed version of the status."""
+        return repr(self)
```
dimaqq, line 754: I'm not sure about this one, but don't have a strong opinion.

My concern is that pretty-print is multiline, while str format should be concise, so ideally not span multiple lines.

test/unit/test_pretty.py
```diff
@@ -0,0 +1,142 @@
...
+def test_small():
+    status_repr = repr(MINIMAL_STATUS)
+    assert (
+        status_repr
+        == "Status(model=ModelStatus(name='mdl', type='typ', controller='ctl', cloud='aws', version='3.0.0'), machines={}, apps={})"
+    )
+    assert str(MINIMAL_STATUS) == status_repr
+
+    # Ensure it's actually a Python repr: it can roundtrip via eval()
+    assert eval(status_repr, jubilant.statustypes.__dict__) == MINIMAL_STATUS
```
dimaqq, line 17: This is an interesting approach.
I dunno if we want to give this guarantee to our users, but it sure is convenient :)

jubilant/_pretty.py
```diff
@@ -0,0 +1,58 @@
...
+
+def _dump(value: object, indent='') -> str:
+    """Pretty-print a value with special cases for dataclasses, lists, and dicts.
+
+    If the value fits on a single (long) line, use that. Otherwise, split onto
+    multiple lines and indent each line.
+
+    For dataclasses, omit fields that are set to their default (or
+    default_factory return value).
+    """
```
dimaqq, line 15: Are nested values allowed on the same line, like `{"a": {"b": {"c": 42}}}` ?

If so, then the function should somehow receive the amount of space already consumed to be used instead of indent if this is a sub-value.

Or, at least I can't see how this would work otherwise.

Example:
```
FirstLevel:
  SubLevelOne: a=23, b=45
  SubLevelTwo: a=88, b=99
```

When SubLevelTwo is rendered, and we're trying to see if the entire FirstLevel+subs would fit on a single line, that really depends on SubLevelOne length.

For another way to approach this, see the comment below.

jubilant/_pretty.py
```diff
@@ -0,0 +1,58 @@
...
+                continue
+            if field.default_factory is not dataclasses.MISSING and v == field.default_factory():
+                continue
+            v_str = _dump(v, sub_indent)
+            fields.append(f'{field.name}={v_str}')
+
+        class_name = value.__class__.__name__  # type: ignore
+        single_line = f'{class_name}({", ".join(fields)})'
+        if len(indent) + len(single_line) <= _MAX_VALUE:
+            return single_line
```
dimaqq, line 32: Here, if the sub-`_dump()` returned a multi-line string, then we shouldn't ever return this level as a single string.

Actually I think that the code does work as is, albeit via a side-effect.

We're relying that the multi-line representation is always longer than a single-line representation. Which is true due to non-zero indent, I think.

If that didn't hold, we could end up with output like this:

```
TopLevelThing(field1=SubOne(
  sub_field1="something long",
  sub_field2="also long, causing a split"), field2=SubTwo(a=42))
```

If I read the code correctly, either document that, or propagate multi-lini-ness explicitly from sublevel to parent.

jubilant/_pretty.py
```diff
@@ -0,0 +1,58 @@
...
+
+        lines_str = '\n'.join(sub_indent + f + ',' for f in fields)
+        return f'{class_name}(\n{lines_str}\n{indent})'
+
+    elif isinstance(value, list):
+        items = [_dump(v, sub_indent) for v in value]
```
dimaqq, line 38: I think that sub-indent is only correct here under the multi-line assumption.

It doesn't compute max single line length correctly for input like `[[[[[[[]]]]]]]`.

Specifically, the `_dump` will be called with indent like `14` for the inner-most list, while fr the purposes of single-line output, it should be `7`.

test/unit/test_pretty.py
```diff
@@ -0,0 +1,142 @@
...
+      charm='local:database-0',
+      charm_origin='local',
+      charm_name='database',
+      charm_rev=0,
+      exposed=False,
+      base=FormattedBase(name='ubuntu', channel='22.04'),
+      scale=1,
+      provider_id='fa764a56-2b71-4f7e-a6eb-b265f13adc4c',
+      address='10.152.183.228',
+      app_status=StatusInfoContents(current='active', message='relation-created: added new secret', since='24 Feb 2025 16:59:43+13:00'),
```
dimaqq, line 96: I suppose that not parsing Juju timestamp is a deliberate choice.

FWIW I prefer strings to datetimes.

dimaqq (commented):
Looks good!

test/unit/test_pretty.py
benhoyt: Yeah, my intent is not to guarantee it to our users, but to make sure we've got it right ourselves. For our users, all the docstring says is "Return a pretty-printed version of the status".

jubilant/_pretty.py
benhoyt: Ah yes, I think you're right there are some issues here. I'll take a closer look.

benhoyt: I've taken a slightly different approach now, and added a comment in the code to clarify the part I think you're concerned about. Here's the docstring:

> If *value*'s fields or sub-values are simple scalar types and the result fits in a limited number of characters, use a single-line format. Otherwise, split onto multiple lines and indent each line.

jubilant/_pretty.py
benhoyt: See https://github.com/canonical/jubilant/pull/37#discussion_r1982346536

test/unit/test_pretty.py
benhoyt: Yep, datetime objects are a mixed blessing, so I decided to KISS for now. I don't think these will be treated/compared as datetimes much, if at all, so the user can do it if they need to.

benhoyt:
@dimaqq I've made some tweaks to the algorithm and added a comment. Hopefully this satisfied your concerns. I'm going to merge this now, and we can always tweak it later if needed (we're not guaranteeing certain formatting to users).

Issue #4, opened 2025-01-22
Add add-secret command


benhoyt:
Probably the signature would be:

```python
def add_secret(self, secret_name: str, content: Mapping[str, str], *, info: str | None = None) -> str:
    """Add a new named secret with the given content, and return the secret URI."""
```

benhoyt:
FWIW, I mapped out what (some of?) the other secrets commands might look like. We probably don't need to add these, as I *think* `add_secret` is the only one that's likely to be used in integration tests, but copying here for the record:

```
@dataclasses.dataclass
class Secret:
    name: str
    revision: int
    content: dict[str, str] | None

class Juju:
    ...

    def add_secret(self, name: str, content: Mapping[str, str], *, info: str | None = None) -> SecretURI:
        ...

    def show_secret(self, name_or_uri: str | SecretURI, *, reveal: bool = False, revision: int = 0) -> Secret:
        ...

    def update_secret(self, name_or_uri: str | SecretURI, content: Mapping[str, str], *, name: str | None = None) -> None:
        pass

    def secrets(self, *, owner: str | None = None) -> list[Secret]:
        ...
```

PR #39, opened 2025-02-27
feat: add "config" method and tests

### LINKED ISSUE #6: Add config command(s)
This command's a bit more tricky, because it both gets and sets config, which makes for an awkward function signature. It's almost always a mistake to overload that kind of thing in Python functions. So I propose something similar to what [Pietro had in his juju.py](https://github.com/canonical/tempo-coordinator-k8s-operator/blob/a487211b1afe6f2370af974089cc1c26ba2b9922/tests/integration/juju.py#L261-L280):

```python
def config_get(self, app: str) -> dict[str, bool | int | float | str]:
    """Gets the configuration of a deployed application."""

def config_set(self, app: str, values: dict[str, bool | int | float | str], *, reset: list[str] | None = None):
    """Sets or resets configuration values of a deployed application."""
```

This would mean to only reset values you'd have to do `config_set('app1', {}, reset=['foo', 'bar'])`, but I think that's okay. It allows you to do sets and resets in one operation, like the underlying command.

Note: the reason to name these `config_get` and `config_set` rather than the more natural `get_config` and `set_config` is for better autocompletion (and findability in an alphabetical list of methods).

benhoyt:
This PR adds a `config` method for the `juju config` CLI command. After discussion, we've decide to use a single method for get and set as it's more 1:1 with the CLI command, and is pretty straight-forward to type (with `@overload`). 

Also:

- Change `dict[K, V]` arguments to `Mapping[K, V]`
- Bump Pyright to latest version
- In docstrings, use *arg* instead of "arg" for mentioning argument names, like the Python docs do
- Add test_wait.test_with_model to cover that case
- For consistency, remove `_no_model` from test name that don't use model (just `_with_model` for those that do)

Fixes #6

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -3,7 +3,8 @@ import logging
 import os
 import subprocess
 import time
-from collections.abc import Callable, Iterable
+from collections.abc import Callable, Iterable, Mapping
+from typing import overload
 
 from .statustypes import Status
 
@@ -23,7 +24,7 @@ class CLIError(subprocess.CalledProcessError):
 
 
 class WaitError(Exception):
-    """Raised when :meth:`Juju.wait`'s "error" callable returns False."""
+    """Raised when :meth:`Juju.wait`'s *error* callable returns False."""
 
 
 class Juju:
@@ -98,7 +99,7 @@ class Juju:
         model: str,
         *,
         controller: str | None = None,
-        config: dict[str, bool | int | float | str] | None = None,
+        config: Mapping[str, bool | int | float | str] | None = None,
     ) -> None:
         """Add a named model and set this instance's model to it.
 
@@ -160,13 +161,13 @@ class Juju:
         attach_storage: str | Iterable[str] | None = None,
         base: str | None = None,
         channel: str | None = None,
-        config: dict[str, bool | int | float | str] | None = None,
-        constraints: dict[str, str] | None = None,
+        config: Mapping[str, bool | int | float | str] | None = None,
+        constraints: Mapping[str, str] | None = None,
         force: bool = False,
         num_units: int = 1,
-        resource: dict[str, str] | None = None,
+        resource: Mapping[str, str] | None = None,
         revision: int | None = None,
-        storage: dict[str, str] | None = None,
+        storage: Mapping[str, str] | None = None,
         to: str | Iterable[str] | None = None,
         trust: bool = False,
     ) -> None:
@@ -231,6 +232,43 @@ class Juju:
                 args.extend(['--to', ','.join(to)])
         if trust:
             args.append('--trust')
+
+        self.cli(*args)
+
+    @overload
+    def config(self, app: str) -> Mapping[str, bool | int | float | str]: ...
+
+    @overload
+    def config(self, app: str, values: Mapping[str, bool | int | float | str | None]) -> None: ...
+
+    def config(
+        self, app: str, values: Mapping[str, bool | int | float | str | None] | None = None
+    ) -> Mapping[str, bool | int | float | str] | None:
+        """Get or set the configuration of a deployed application.
+
+        If called with only the *app* argument, get the config and return it.
+        If called with the *values* argument, set the config values and return
+        ``None``.
+
+        Args:
+            app: Application name to get or set config for.
+            values: Mapping of config names to values. Reset values that are
+                ``None``.
+        """
+        if values is None:
+            stdout = self.cli('config', '--format', 'json', app)
+            result = json.loads(stdout)
+            return {k: v['value'] for k, v in result['settings'].items()}
+
+        reset = []
+        args = ['config', app]
+        for k, v in values.items():
+            if v is None:
+                reset.append(k)
+            else:
+                args.append(_format_config(k, v))
+        if reset:
+            args.extend(['--reset', ','.join(reset)])
 
         self.cli(*args)
 
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ jubilant.egg-info
 /dist
 /SCRATCH.md
 /docs/_build
+/docs/.sphinx
 /.idea
 /t.py
 /.coverage
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ dynamic = ["version"]
 
 [dependency-groups]
 dev = [
-    "pyright==1.1.394",
+    "pyright==1.1.395",
     "pytest==8.3.4",
     "pytest-cov==6.0.0",
     "ruff==0.9.6",
--- a/test/unit/mocks.py
+++ b/test/unit/mocks.py
@@ -16,7 +16,7 @@ class Run:
         self._commands = {}
         self.call_count = 0
 
-    def handle(self, args: list[str], returncode: int = 0, stdout: str = '', stderr: str = ''):
+    def handle(self, args: list[str], *, returncode: int = 0, stdout: str = '', stderr: str = ''):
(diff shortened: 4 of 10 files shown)


tonyandrewmeyer (approved):
Looks good!

Two thoughts:

1. What's the plan when you get to the various secret CLI commands? Will a secret ID be a plain string, or will you have a string alias or subtype so that tests need to be explicit about what is a secret ID, and maybe about its form? If it's not going to be a plain string, then it might be worth doing that now, and having config also accept that type. (I lean towards making it an alias or subclass since Juju doesn't treat it as a plain string).
2. I assume you've deliberately left out the `--file` version of config to keep things simpler and because it seems less likely that you'd pass the config as a file from an integration test. That seems reasonable to me -- it's a pretty simple `juju.cli` if needed (with no response to parse), and it's already overloaded.
3. What's the plan for `application-config`? With the CLI, you get this from `config` as well. I don't think it's worth returning a pair of mappings here, but you probably do want to be able to retrieve that. Maybe an `application_config()` method, even though that does break away from 1:1 CLI mapping? As far as I know, you can't *set* values in there via `config`, so it is simpler.
4. The docs have a `--branch` argument for config as well, and it seems like it would be unobtrusive to have `branch:str='master'` as an argument. However, my Juju 3.6 doesn't seem to accept `--branch` even though it's in the `--help`, and I'm not really familiar with branches. Do you know the story here?

benhoyt:
> 1. What's the plan when you get to the various secret CLI commands? Will a secret ID be a plain string, or will you have a string alias or subtype so that tests need to be explicit about what is a secret ID, and maybe about its form? If it's not going to be a plain string, then it might be worth doing that now, and having config also accept that type. (I lean towards making it an alias or subclass since Juju doesn't treat it as a plain string).

Interesting idea. I'm planning to only add the `add_secret` command for v1, as I think that's the main/only one that'll be used in integration tests. See my [comment here](https://github.com/canonical/jubilant/issues/4#issuecomment-2689861405) for what that signatures (and the others) might look like.

Yeah, on balance I think it'd be useful to add a `SecretURI` type now, to avoid type churn later. It may also be helpful in preventing logging sensitive data to logs (though hopefully in integration tests you wouldn't use real secrets).

You'll still be able to call `juju.config('app', {"secr": "secret:abcd1234"}) with the secret URI as a plain string though ... unless we do smarts where we fetch the types from Juju first -- but I think that's a bit too clever.

So yes, I think I'll make this change with a `class SecretURI(str): pass`, and then convert string values to `SecretURI` when fetching config for `type: secret` ones.

> 2. I assume you've deliberately left out the --file version of config to keep things simpler and because it seems less likely that you'd pass the config as a file from an integration test. That seems reasonable to me -- it's a pretty simple juju.cli if needed (with no response to parse), and it's already overloaded.

Yeah, that's right. It just seems simpler for now. The integration test could always read from a file and pass as args.

> 3. What's the plan for application-config? With the CLI, you get this from config as well. I don't think it's worth returning a pair of mappings here, but you probably do want to be able to retrieve that. Maybe an application_config() method, even though that does break away from 1:1 CLI mapping? As far as I know, you can't set values in there via config, so it is simpler.

Yeah, I saw it stuffed in there. I was planning to just ignore it, thinking that integration tests don't need that. Do you think they will? Anyway, if so, I think maybe an `application_config=True` keyword-only arg. It's not quite 1:1 with the CLI, but close, as the application config is in an `'application-config'` key of the JSON/YAML.

> 4. The docs have a --branch argument for config as well, and it seems like it would be unobtrusive to have branch:str='master' as an argument. However, my Juju 3.6 doesn't seem to accept --branch even though it's in the --help, and I'm not really familiar with branches. Do you know the story here?

I noticed that too. Branches are an incomplete feature behind a feature flag, which is turned off by default. Though somewhat buggily, the help for it shows regardless of whether the feature flag is on or off.

jubilant/_juju.py
```diff
@@ -23,7 +26,15 @@ def __str__(self):
...
-    """Raised when :meth:`Juju.wait`'s "error" callable returns False."""
+    """Raised when :meth:`Juju.wait`'s *error* callable returns False."""
+
+
+class SecretURI(str):
+    """A string subclass that represents a secret URI ("secret:...")."""
+
+
+type ConfigValue = bool | int | float | str | SecretURI
+"""The possible types a charm config value can be."""
```
benhoyt, line 37: I couldn't for the life of my get Sphinx showing these how I wanted them, or to get ConfigValue to show up in the docs at all. I tried using [`autodoc_type_aliases`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_type_aliases), but it was acting up. I think I'll wait for David's Sphinx/docs overhaul and then play with it a bit more.

benhoyt:
@tonyandrewmeyer Ignore this till next week (I'm working late tonight to make up some time from an appointment today). But if you could please review the changes before I merge, that would be great.

tonyandrewmeyer (approved):

tonyandrewmeyer:
> Yeah, on balance I think it'd be useful to add a `SecretURI` type now, to avoid type churn later. It may also be helpful in preventing logging sensitive data to logs (though hopefully in integration tests you wouldn't use real secrets).

This looks good, thanks!

> You'll still be able to call `juju.config('app', {"secr": "secret:abcd1234"}) with the secret URI as a plain string though ... unless we do smarts where we fetch the types from Juju first -- but I think that's a bit too clever.

Agreed.

> > 3. What's the plan for application-config?
[...]
> Yeah, I saw it stuffed in there. I was planning to just ignore it, thinking that integration tests don't need that. Do you think they will?

Is `trust` the only field that can be in there at the moment? (The docs don't seem to say, and I haven't looked at the source). If so then they probably don't, since it seems like the test(s) would be managing that value themselves, so not need to read it. On the other hand, maybe it is convenient to do `juju.config('foo', app_config=True)['trust']` rather than have to keep track of the trust status somewhere.

> I've added an `app_config=True` keyword-only arg. It's not quite 1:1 with the CLI, but close, as the application config is in an `'application-config'` key of the JSON/YAML. Let me know what you think.

This looks good to me. I feel it's small enough that including it even though it's not clear if it's needed is ok. However, if you'd rather keep v1 more minimal, I think it'd be ok to back this out and only add it if the need becomes clear.

> I noticed that too. Branches are an incomplete feature behind a feature flag, which is turned off by default. Though somewhat buggily, the help for it shows regardless of whether the feature flag is on or off.

I saw the Matrix thread, thanks! I agree it seems buggy that it's there in the docs even when unusable (at least the live docs, but I would say that it should not be there in the static ones, or mention the flag, also).

benhoyt:
> Is trust the only field that can be in there at the moment? (The docs don't seem to say, and I haven't looked at the source).

No, there are a couple of `juju-*` fields and, for K8s models, tons of `kubernetes-*` ones. [See output from my test "database" charm.](https://gist.github.com/benhoyt/42fc05acd604c12c26fb90c6a745ab28)

> This looks good to me. I feel it's small enough that including it even though it's not clear if it's needed is ok. However, if you'd rather keep v1 more minimal, I think it'd be ok to back this out and only add it if the need becomes clear.

I think I'll keep it. It's not really getting in the way, and it could be useful to either read "trust" or one of these special Juju/K8s values.

tonyandrewmeyer:
> No, there are a couple of `juju-*` fields and, for K8s models, tons of `kubernetes-*` ones. [See output from my test "database" charm.](https://gist.github.com/benhoyt/42fc05acd604c12c26fb90c6a745ab28)

Ah, nice to know, thanks! I only checked it against an untrusted machine charm, so I guess that's why I only saw trust. I agree, with all those extra fields, it's good to have.

PR #40, opened 2025-02-28
chore: turn on Pyright strict typing

benhoyt:
This is probably best for a library, and wasn't actually too annoying.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -119,6 +119,8 @@ builtins-ignorelist = ["id", "min", "map", "range", "type", "input", "format"]
 include = ["jubilant/*.py", "test/unit/*.py", "test/integration/*.py"]
 pythonVersion = "3.12"
 pythonPlatform = "All"
+typeCheckingMode = "strict"
+reportPrivateUsage = false
 
 [tool.pytest.ini_options]
 pythonpath = [
--- a/test/unit/test_all_any.py
+++ b/test/unit/test_all_any.py
@@ -1,6 +1,6 @@
 import copy
 import json
-from collections.abc import Callable
+from typing import Any
 
 import pytest
 
@@ -19,7 +19,7 @@ from .fake_statuses import MINIMAL_STATUS, SNAPPASS_JSON
         jubilant.all_waiting,
     ],
 )
-def test_all_no_apps(all_func: Callable):
+def test_all_no_apps(all_func: Any):
     # Just like Python's all(), all_* helpers return True if no apps
     assert all_func(MINIMAL_STATUS)
     assert all_func(MINIMAL_STATUS, [])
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -119,6 +119,8 @@ builtins-ignorelist = ["id", "min", "map", "range", "type", "input", "format"]
 include = ["jubilant/*.py", "test/unit/*.py", "test/integration/*.py"]
 pythonVersion = "3.12"
 pythonPlatform = "All"
+typeCheckingMode = "strict"
+reportPrivateUsage = false
```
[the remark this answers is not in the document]
james-garner-canonical: I think this is fine for us as a library creator rather than consumer. If we wanted to be more strict we could instead disable this in the relevant test files to catch us cheekily using private contents of other libraries. Unfortunately you can only do this by adding a comment to the file rather than via `pyproject.toml`.
```py
# pyright: reportPrivateUsage = false
```
Either way seems fine to me.

test/unit/test_all_any.py
```diff
@@ -19,7 +19,7 @@
         jubilant.all_waiting,
     ],
 )
-def test_all_no_apps(all_func: Callable):
+def test_all_no_apps(all_func: Any):
```
james-garner-canonical, line 22: I think this use of `Any` is fine too, otherwise it looks like you'd need a protocol to capture the optional argument here:
```py
class AllAnyFunc(Protocol):
    def __call__(self, status: Status, apps: Iterable[str] | None = None) -> bool: ...
```
Which probably doesn't add enough value really.

test/unit/test_all_any.py
```diff
@@ -182,15 +182,15 @@ def test_any_two_apps(any_func: Callable, expected: str, unexpected: str):
...
+def test_all_and_any_type_errors(func: Any):
     with pytest.raises(TypeError):
         func(MINIMAL_STATUS, apps='app')
     with pytest.raises(TypeError):
         func(MINIMAL_STATUS, apps=b'app')
 
 
 def set_app_and_unit_status(
-    status_dict: dict, app: str, app_status: str, unit_status: str, *, unit: int = 0
+    status_dict: dict[str, Any], app: str, app_status: str, unit_status: str, *, unit: int = 0
```
james-garner-canonical, line 193: And definitely too painful to try and type this more than `Any`!

james-garner-canonical (approved):
Looks great, the only other note I had was that we need to add a `py.typed` file (just an empty file) which declares that users can use our type annotations. Otherwise type checkers want to look for stubs instead -- they'll fall back to using our type annotations in the absence of stubs, but will complain about it to the library user.

pyproject.toml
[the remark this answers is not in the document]
benhoyt: I played with this, but we use `_from_dict` in `_juju.py` and throughout `statustypes.py` too. It seems weird that it complains when it's in the same library -- I think this check is too sweeping, so I think I'm just going to leave it off globally.

test/unit/test_all_any.py
benhoyt: Agreed -- too much effort for a test file.

benhoyt:
Added `py.typed` -- thanks for the note, I would have forgotten that!

PR #42, opened 2025-03-01
feat: add run method

### LINKED ISSUE #13: Add run command
Signature something like:

```python
def run(self, units: list[str], action: str, *, params: dict[str, Any]) -> ???:
```

TODO:

- Need to figure out the result/output
- Units being a list is awkward for the common use case (`run(['foo/0'], 'backup')`)
- Should the type of params values be narrower?
- Do we need support for `--background` and `show-operation`? Probably can wait for v2.

benhoyt:
Adds support for the "juju run" command. Don't support `--background` mode or `juju show-operation` for now (likely not needed for integration testing).

This PR adds a dependency on PyYAML, which is kind of inevitable at some point. All params are passed as YAML, to ensure they round trip with the correct types (and to avoid ad-hoc foo.bar=baz serialization on the command line).

I played around with the API design / signature of this for a while. Initially I had an `@overload` that supported the `unit` parameter being either a single unit name a list/tuple of unit names, but the return value was too complex (`ActionResult` in single case, `dict[str, ActionResult]` in the other). We decided to stick with the single-unit version for now, and we can add `run_multiple` or something later.

Fixes #13

### DIFF
--- a/jubilant/_actions.py
+++ b/jubilant/_actions.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import dataclasses
-from typing import Any
+from typing import Any, Literal
 
 
 @dataclasses.dataclass(frozen=True, kw_only=True)
@@ -11,8 +11,8 @@ class ActionResult:
     success: bool
     """Whether the action was successful."""
 
-    status: str
-    """Status of the action, for example "completed" or "failed"."""
+    status: Literal['aborted', 'cancelled', 'completed', 'error', 'failed']
+    """Status of the action (Juju operation). Typically "completed" or "failed"."""
 
     task_id: str
     """Task ID of the action, for use with "juju show-task"."""
@@ -33,6 +33,9 @@ class ActionResult:
     stderr: str = ''
     """Stderr printed by the action hook."""
 
+    message: str = ''
+    """Failure message, if the charm provided a message when it failed the action."""
+
     log: list[str] = dataclasses.field(default_factory=list)
     """List of messages logged by the action hook."""
 
@@ -50,6 +53,7 @@ class ActionResult:
             return_code=return_code,
             stdout=stdout,
             stderr=stderr,
+            message=d.get('message') or '',
             log=d.get('log') or [],
             task_id=d['id'],
         )
--- a/jubilant/_actions.py
+++ b/jubilant/_actions.py
@@ -1,0 +1,65 @@
+from __future__ import annotations
+
+import dataclasses
+from typing import Any
+
+
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class ActionResult:
+    """Holds the results of Juju running an action on a single unit."""
+
+    success: bool
+    """Whether the action was successful."""
+
+    status: str
+    """Status of the action, for example "completed" or "failed"."""
+
+    task_id: str
+    """Task ID of the action, for use with "juju show-task"."""
+
+    results: dict[str, Any] = dataclasses.field(default_factory=dict)
+    """Results of the action provided by the charm.
+
+    This excludes the special "return-code", "stdout", and "stderr" keys
+    inserted by Juju; those values are provided by separate attributes.
+    """
+
+    return_code: int = 0
+    """Return code from executing the charm action hook."""
+
+    stdout: str = ''
+    """Stdout printed by the action hook."""
+
+    stderr: str = ''
+    """Stderr printed by the action hook."""
+
+    log: list[str] = dataclasses.field(default_factory=list)
+    """List of messages logged by the action hook."""
+
+    @classmethod
+    def _from_dict(cls, d: dict[str, Any]) -> ActionResult:
+        status = d['status']
+        results: dict[str, Any] = d.get('results') or {}
+        return_code = results.pop('return-code', 0)
+        stdout = results.pop('stdout', '')
+        stderr = results.pop('stderr', '')
+        return cls(
+            success=status == 'completed',
+            status=status,
+            results=results,
+            return_code=return_code,
+            stdout=stdout,
+            stderr=stderr,
+            log=d.get('log') or [],
+            task_id=d['id'],
+        )
+
+
+class ActionError(Exception):
+    """Exception raised when an action run on a single unit fails."""
+
+    result: ActionResult
+    """Results of running the action."""
+
+    def __init__(self, result: ActionResult):
+        self.result = result
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -2,9 +2,15 @@ import json
 import logging
 import os
 import subprocess
+import tempfile
 import time
-from collections.abc import Callable, Iterable
+from collections.abc import Callable, Iterable, Mapping
+from typing import Any, overload
 
+import yaml
+
+from . import ActionError
+from ._actions import ActionResult
 from .statustypes import Status
 
 logger = logging.getLogger('jubilant')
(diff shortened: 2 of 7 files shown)


jubilant/_actions.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class ActionResult:
+    """Holds the results of Juju running an action on a single unit."""
+
+    success: bool
+    """Whether the action was successful."""
+
+    status: str
```
tonyandrewmeyer, line 14: What do you think about making this either `Literal[...]` or a `StrEnum`? 8 possible values, 3 of which seem likely in integration tests. I don't think you'll get regular autocomplete, or static type checkers indicating comparison errors, but I think it is able to help LLM autocomplete be more accurate.

jubilant/_juju.py
```diff
@@ -234,10 +240,93 @@ def deploy(
...
+        if isinstance(unit, str):
+            args.append(unit)
+        else:
+            args.extend(unit)
+        args.append(action)
+
+        params_file = None
+        if params is not None:
+            with tempfile.NamedTemporaryFile('w+', delete=False) as params_file:
+                yaml.safe_dump(params, params_file)
```
tonyandrewmeyer, line 305:
```
$ python -m timeit --setup "import yaml;x={'arg': 'value'}" "yaml.safe_dump(x)"
5000 loops, best of 5: 72.4 usec per loop
$ python -m timeit --setup "import yaml;x={'arg': 'value'}" "yaml.dump(x, Dumper=yaml.CSafeDumper)"
10000 loops, best of 5: 20 usec per loop
```

I wish pyyaml would automatically use the C version if it was available. Is your thinking here that tens of microseconds saved per test isn't worth the hassle of having a file like `ops._private.yaml`? That seems reasonable for integration tests where you're probably waiting *minutes* for a `deploy` command anyway.

jubilant/_juju.py
```diff
@@ -234,10 +240,93 @@ def deploy(
...
+            args.extend(['--params', params_file.name])
+
+        try:
+            stdout = self.cli(*args)
+            # Command doesn't return any stdout if no units exist
```
tonyandrewmeyer, line 310: Nit:
```suggestion
            # Command doesn't return any stdout if no units exist.
```

tonyandrewmeyer (approved):
> This PR adds a dependency on PyYAML, which is kind of inevitable at some point.

I think you could avoid it here by `json.dump`ing instead, since the action args have to be defined with JSONSchema, so presumably are limited to what can be expressed in JSON, and the dumped JSON would be valid YAML.

However, I suspect that you're right that it'll be needed anyway at some point, and it's a well-known dependency that's already used in other Charm-Tech projects, so it's a low-weight addition.

> I played around with the API design / signature of this for a while. In particular, I'm not sure if it's worth the multiple-units variant for v1.

I lean towards 'not'. I think your design here is good, but it seems likely that integration tests would be running against a single unit most of the time and putting a `juju.run` in a loop is simple enough to do (and only loses the concurrency). 

(But 'approving' anyway, because I don't object to having it if that's how the decision plays out).

> I think you want the single-unit version to raise errors on failures.

Agreed.

jubilant/_juju.py
```diff
@@ -234,10 +240,93 @@ def deploy(
...
+            If a list of units is specified, return a dict where each key is a
+            unit name and each value contains the result for that unit. Only
+            units that exist will be included in the dict (but no exception is
+            raised if some don't exist).
+
+        Raises:
+            ValueError: if a single unit is specified but the unit doesn't exist.
+            ActionError: if a single unit is specified and the action failed.
+
+        Example::
```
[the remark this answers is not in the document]
james-garner-canonical: Hmm I  think I kind of prefer the opposite (current approach) -- signature all in one place (and ideally before the rest of the docstring).

jubilant/_juju.py
james-garner-canonical: If we're thinking about efficiency, is it worth avoiding the tempfile altogether and passing the arguments via the commandline if they're short enough?

jubilant/_juju.py
```diff
@@ -234,10 +240,93 @@ def deploy(
...
+        self, unit: str, action: str, params: Mapping[str, Any] | None = None
+    ) -> ActionResult: ...
+
+    @overload
+    def run(
+        self,
+        unit: list[str] | tuple[str, ...],
+        action: str,
+        params: Mapping[str, Any] | None = None,
+    ) -> dict[str, ActionResult]: ...
```
james-garner-canonical, line 254: I've gotta say, I really don't like the shape of the output changing based on the type of one of the arguments like this. It's the reasonable approach here, but it pushes me in favour of `run` taking a single `unit: str` argument in `v1`, with an eventual `run_multiple` taking `units: Iterable[str]` if needed.

james-garner-canonical (commented):
I like the idea of separate `run` and `run_multiple` a lot more than the `overload`.

The `PyYAML` dependency seems completely fine given that it's already an `ops` dependency -- of course this doesn't mean it's already installed in the integration testing environment, but it doesn't add increase the dependency surface we need to worry about.

jubilant/_juju.py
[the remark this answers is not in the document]
benhoyt: I'm with Tony here: I significantly prefer the paragraphs and example first, then the "metadata" (Args/Returns/Raises). I think it looks better in the rendered docs too. So I've gone with that.

jubilant/_actions.py
benhoyt: Done as a Literal. I've added the 5 that I believe are possible here.

jubilant/_juju.py
benhoyt: Ugh, I'd forgotten about this. It sucks that PyYAML doesn't opt you in to either safety or the C version by default. I've added a `_yaml.py` stub, copied from Ops's `_private/yaml.py` but with types copied from typeshed.

benhoyt:
@james-garner-canonical:

> I like the idea of separate run and run_multiple a lot more than the overload.
>
> I've gotta say, I really don't like the shape of the output changing based on the type of one of the arguments like this. It's the reasonable approach here, but it pushes me in favour of run taking a single unit: str argument in v1, with an eventual run_multiple taking units: Iterable[str] if needed.

Yeah, I agree. I think it's okay in `config` where the switch was "return something" vs "return nothing", but I think two functions is clearer and simpler.

As such, I've removed the support for multiple units for now. We can always add `run_multiple` in future. I think the drawbacks of it not matching the Juju CLI exactly -- single command = single method -- outweigh the positives.

jubilant/_juju.py
```diff
@@ -2,9 +2,13 @@
...
 import os
 import subprocess
+import tempfile
 import time
-from collections.abc import Callable, Iterable
+from collections.abc import Callable, Iterable, Mapping
+from typing import Any
 
+from . import ActionError, _yaml
+from ._actions import ActionResult
```
tonyandrewmeyer, line 11: This feels a little weird. Wouldn't it be cleaner to import `ActionError` and `ActionResult` from the same place? And that place be `._actions`? I don't love a `from .` that imports a class and a module.

benhoyt: Yep, good call -- just an oversight. Fixed.

jubilant/_juju.py
benhoyt: Agreed -- I've removed the multiple-units overload for now.

jubilant/_juju.py
benhoyt: Re @james-garner-canonical comment, it's not *really* about efficiency, as efficiency probably doesn't matter a great deal here. And you're right that the overhead of creating a file probably outweighs that of encoding YAML in Python, however, it seems a real waste to not use the C version if it's installed, so I've done that. (And it'd be a slight pain to have to write a function that turns a recursive structure into `foo.bar.bazz=whateve`.)

jubilant/_juju.py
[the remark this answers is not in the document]
james-garner-canonical: Oh yep, that's good too -- I should have checked the `wait` example before commenting, I was worried we were going to have args then example then returns/raises.

PR #45, opened 2025-03-03
feat: add integrate method

### LINKED ISSUE #8: Add integrate command
Signature something like:

```python
def integrate(self, app1: str, app2: str, *, vias: list[str] | None = None):
```

benhoyt:
Very straight-forward.

Fixes #8

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -306,7 +306,7 @@ class Juju:
             app2: The other of the applications (and endpoints) to integrate.
             via: Inform the offering side (the remote application) of the
                 source of traffic, to enable network ports to be opened. This
-                is in the format ``<cidr-subnet>``.
+                is in CIDR notation, for example ``192.0.2.0/24``.
         """
         args = ['integrate', app1, app2]
         if via:
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -300,6 +300,8 @@ class Juju:
 
         To integrate an application in the current model with an application in
         another model (cross-model), prefix *app1* or *app2* with ``<model>.``.
+        To integrate with an application on another controller, *app1* or *app2* must
+        be an offer endpoint. See ``juju integrate --help`` for details.
 
         Args:
             app1: One of the applications (and endpoints) to integrate.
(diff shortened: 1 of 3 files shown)


jubilant/_juju.py
```diff
@@ -290,6 +290,32 @@ def destroy_model(
         if model == self.model:
             self.model = None
 
+    def integrate(self, app1: str, app2: str, *, via: str | Iterable[str] | None = None) -> None:
+        """Integrate two applications, creating a relation between them.
+
+        The order of *app1* and *app2* is not significant. Each of them should
+        be in the format ``<application>[:<endpoint>]``. The endpoint is only
```
tonyandrewmeyer, line 297: Is it worth saying that one of them can also be an offer URL or alias, either here or down in the CMR paragraph?

jubilant/_juju.py
```diff
@@ -290,6 +290,32 @@ def destroy_model(
         if model == self.model:
             self.model = None
 
+    def integrate(self, app1: str, app2: str, *, via: str | Iterable[str] | None = None) -> None:
```
tonyandrewmeyer, line 293: How would you feel about supporting actual IPvXNetwork objects, like:

```suggestion
    def integrate(self, app1: str, app2: str, *, via: str | ipaddress.IPv4Network | ipaddress.IPv6Network | Iterable[str | ipaddress.IPv4Network | ipaddress.IPv6Network ] | None = None) -> None:
```

(You could have a private alias to make that shorter and more readable).

It should only require changing the via to be:

```python
        if via:
            if isinstance(via, (str, ipaddress.IPv4Network, ipaddress.IPv6Network)):
                args.extend(['--via', str(via)])
            else:
                args.extend(['--via', ','.join(str(v) for v in via))])
```

tonyandrewmeyer (approved):

tonyandrewmeyer:
Re: the docs failure, does it need `CIDR` added to a list somewhere? Or the spellcheck settings tweaked somehow?

jubilant/_juju.py
benhoyt: It's a reasonable suggestion. However, I'm inclined to keep them simple types for v1 at least. I'm skeptical that `IPvXNetwork` wouldn't be used instead of just `str`, in this context at any rate, and it doesn't seem worth adding type complexity for until we see that it is desired.

jubilant/_juju.py
benhoyt: Yeah, good idea. I've added a sentence about that, with "See ``juju integrate --help`` for details." as we don't want to repeat that whole thing here.

PR #46, opened 2025-03-03
feat: add add_unit method

### LINKED ISSUE #5: Add add-unit command
Signature should likely be:

```python
def add_unit(self, app: str, *, num_units: int = 1, to: str | None = None, attach_storages: list[str] | None = None):
   """Add one or more units to a deployed application."""
```

Debatable whether `attach_storage(s)` should be a list (or a comma-separated string like the CLI), because a single storage is the most common use case. However, it seems most consistent and Pythonic to do so, and then it's clear from the type signature that you can provide multiple.

benhoyt:
The `attach_storage` and `to` and `num_units` parameters are the same as for `deploy`.

Fixes #5

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -119,6 +119,41 @@ class Juju:
         self.cli(*args, include_model=False)
         self.model = model
 
+    def add_unit(
+        self,
+        app: str,
+        *,
+        attach_storage: str | Iterable[str] | None = None,
+        num_units: int = 1,
+        to: str | Iterable[str] | None = None,
+    ):
+        """Add one or more units to a deployed application.
+
+        Args:
+            app: Name of application to add units to.
+            attach_storage: Existing storage(s) to attach to the deployed unit, for example,
+                ``foo/0`` or ``mydisk/1``. Not available for Kubernetes models.
+            num_units: Number of units to add.
+            to: Machine or container to deploy the unit in (bypasses constraints). For example,
+                to deploy to a new LXD container on machine 25, use ``lxd:25``.
+        """
+        args = ['add-unit', app]
+
+        if attach_storage:
+            if isinstance(attach_storage, str):
+                args.extend(['--attach-storage', attach_storage])
+            else:
+                args.extend(['--attach-storage', ','.join(attach_storage)])
+        if num_units != 1:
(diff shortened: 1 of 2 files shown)


test/unit/test_add_unit.py
```diff
@@ -0,0 +1,43 @@
...
+            'stg',
+            '--num-units',
+            '3',
+            '--to',
+            'lxd:25',
+        ]
+    )
+    juju = jubilant.Juju()
+
+    juju.add_unit('app2', attach_storage='stg', num_units=3, to='lxd:25')
```
dimaqq, line 36: Would I be correct to assume that the run mock requires exactly one call with exact same arguments?

jubilant/_juju.py
```diff
@@ -119,6 +119,41 @@ def add_model(
...
+        """
+        args = ['add-unit', app]
+
+        if attach_storage:
+            if isinstance(attach_storage, str):
+                args.extend(['--attach-storage', attach_storage])
+            else:
+                args.extend(['--attach-storage', ','.join(attach_storage)])
+        if num_units != 1:
+            args.extend(['--num-units', str(num_units)])
```
dimaqq, line 148: Would it make sense to validate the range, like disallowing negative value and zero?
Or does the user get a reasonable error from Juju cli in this case?

jubilant/_juju.py
```diff
@@ -119,6 +119,41 @@ def add_model(
...
+        if to:
+            if isinstance(to, str):
+                args.extend(['--to', to])
+            else:
+                args.extend(['--to', ','.join(to)])
+
+        self.cli(*args)
```
dimaqq, line 155: I'm a bit surprised that `juju add-unit` doesn't return anything, not even the new unit number, but I guess it is what it is.

test/unit/test_add_unit.py
benhoyt: >= 1 calls with the exact same arguments, yes. It seems to work well!

jubilant/_juju.py
benhoyt: Yep, the Juju CLI gives a nice message -- in general, I'm relying on the CLI for validation of most of these things.

```
$ juju add-unit database --num-units -3
ERROR --num-units must be a positive integer
```

Issue #31, opened 2025-02-19
Add integration tests


benhoyt:
We should have at least some basic integration tests (using Juju via Concierge), for example, testing `deploy` and `wait`.

PR #47, opened 2025-03-03
feat: add remove_unit method

### LINKED ISSUE #12: Add remove-unit command
Signature something like:

```python
def remove_unit(self, *units: str, destroy_storage: bool = False, force: bool = False, num_units: int = 0:
```

Note that the first element of `units` must actually be an application name if `num_units` is specified (K8s models only).

benhoyt:
This one's a little funky because of machines/K8s differences, but it matches the Juju CLI.

Fixes #12

### DIFF
--- a/docs/.custom_wordlist.txt
+++ b/docs/.custom_wordlist.txt
@@ -1,7 +1,6 @@
 # Leave a blank line at the end of this file to support concatenation
 args
 config
-num
 Pythonic
 stdout
 stderr
--- a/docs/.custom_wordlist.txt
+++ b/docs/.custom_wordlist.txt
@@ -1,6 +1,7 @@
 # Leave a blank line at the end of this file to support concatenation
 args
 config
+num
 Pythonic
 stdout
 stderr
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -290,6 +290,51 @@ class Juju:
         if model == self.model:
             self.model = None
 
+    def remove_unit(
+        self,
+        unit: str | Iterable[str],
+        *,
+        destroy_storage: bool = False,
+        force: bool = False,
+        num_units: int = 0,
+    ) -> None:
+        """Remove application units from the model.
+
+        Examples::
+
+            # Kubernetes model:
+            juju.remove_unit('wordpress', num_units=2)
+
+            # Machine model:
+            juju.remove_unit('wordpress/1')
+            juju.remove_unit(['wordpress/2', 'wordpress/3'])
+
+        Args:
+            unit: On machine models, this is the name of the unit or units to remove.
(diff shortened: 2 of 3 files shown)


docs/.custom_wordlist.txt
```diff
@@ -1,6 +1,7 @@
 # Leave a blank line at the end of this file to support concatenation
 args
 config
+num
```
benhoyt, line 4: @dwilding Is there a better thing to do here? I think it's complaining about me using `*num_units*` in the `remove_unit` docstring to mention the parameter name. However, using `*asterisks*` to italicize parameter names is how it's done in the Python docs, so I think we should follow suit. [For example, os.path docs.](https://github.com/python/cpython/blob/3a7f17c7e2f72a836a019c316818c446a0c71d75/Doc/library/os.path.rst#L59)

Is there a way to tell the spellchecker, this is a parameter name, don't spellcheck it?

dwilding: Yep, see https://github.com/canonical/jubilant/pull/48. If you merge my PR first, you'll be able to remove `num` from the word list.

benhoyt: Great, thanks! Fixed.

jubilant/_juju.py
```diff
@@ -290,6 +290,51 @@ def destroy_model(
...
             self.model = None
 
+    def remove_unit(
+        self,
+        unit: str | Iterable[str],
+        *,
+        destroy_storage: bool = False,
+        force: bool = False,
+        num_units: int = 0,
+    ) -> None:
```
james-garner-canonical, line 300: This is probably the right way to do it, since we're aiming to follow the cli. The api is a little weird, but that's just the cli -- the implementation on this end looks clean enough. On the other hand we did go with `run` only handling a single action. Thoughts on making `remove_unit` only remove a single unit? So `unit: str` and no `num_units` argument?

test/unit/test_remove_unit.py
```diff
@@ -0,0 +1,47 @@
+import pytest
```
james-garner-canonical, line 1: Unit tests look good. Will there be any integration tests for this?

james-garner-canonical (approved):

test/unit/test_remove_unit.py
benhoyt: Yes, definitely - they'll be coming with #31.

jubilant/_juju.py
benhoyt: I think `run` is different because the shape of the return value was changing so significantly. Here there's no return value and we don't have that problem. In addition, I think removing multiple units may actually be used in integration tests.

PR #49, opened 2025-03-06
feat: add test helpers and run integration tests in CI

### LINKED ISSUE #20: Add fixtures and integrate with pytest
These would live either in a _fixtures.py imported into `__init__.py`, or perhaps better, a separate `jubilant.fixtures` module.

We at least need a fixture to create a model with a random model name and yield the `jubilant.Juju` instance. Not sure what other fixtures we should start with. Look at Pietro's [conftest.py](https://github.com/canonical/tempo-coordinator-k8s-operator/blob/a487211b1afe6f2370af974089cc1c26ba2b9922/tests/integration/conftest.py) for inspiration.

Do we also need pytest command-line options, [similar to Pietro's](https://github.com/canonical/tempo-coordinator-k8s-operator/blob/a487211b1afe6f2370af974089cc1c26ba2b9922/tests/integration/conftest.py#L43)? IMO let's keep it more minimal for v1.

benhoyt:
This PR adds a `jubilant.temp_model` context manager, intended mainly for use in a pytest test fixture (like the one shown in `test/integration/conftest.py`. The idea is that charm authors would write their own "juju" or similar fixture with the scope they need (it's too hard to try to guess this, and awkward to provide `juju_session`, `juju_module` fixtures for the various scopes).

Also:

- Make `add_model` not switch, to avoid affecting the global state of CLI users
- Remove the `switch` method; similar to above, avoid ways to affect global state of CLI users, and just use `Juju(model=...)` instead
- Add `destroy_storage` arg to `destroy_model` method
- Override `Status.__eq__` to avoid comparing the always-updated controller timestamp

I'm going to add docs in a separate PR with a docs landing page / tutorial.

Fixes #20


.github/workflows/ci.yaml
```diff
@@ -41,12 +45,35 @@ jobs:
...
+        uses: astral-sh/setup-uv@v5
+
+      - name: Install concierge
+        run: sudo snap install --classic concierge
+
+      - name: Prepare Juju
+        run: sudo concierge prepare --verbose --juju-channel=3/stable --charmcraft-channel=3.x/stable -p microk8s
+
+      - name: Run integration tests
+        run: make integration
```
dimaqq, line 79: Off-topic: I've been using shorthand like this:

```suggestion
    steps:
      - uses: actions/checkout@v4
      - uses: astral-sh/setup-uv@v5
      - run: sudo snap install --classic concierge
      - run: sudo concierge prepare --verbose --juju-channel=3/stable --charmcraft-channel=3.x/stable -p microk8s
      - run: make integration
```

Which is both concise and expressive:

Screenshot 2025-03-10 at 13 10 12

☝🏻  https://github.com/dimaqq/otlp-json/actions/runs/13646276620/job/38145683050

jubilant/fixtures.py
```diff
@@ -0,0 +1,22 @@
...
+@pytest.fixture  # TODO: what should the scope be?
+def juju() -> Generator[Juju, None, None]:
+    """Fixture that creates a model and destroys it when done."""
+    juju = Juju()
+    model = 'jubilant-' + secrets.token_hex(6)  # 6 bytes (12 hex digits) should be plenty
+    juju.add_model(model)
+    try:
+        yield juju
+    finally:
+        juju.destroy_model(model, force=True)  # TODO: anything wrong with using --force here?
```
dimaqq, line 22: Based on my experience with ephemeral juju models that charm integration tests leave behind sometimes:

- force is possibly better than no force
- force is not enough

If storage was created in the model, force doesn't seem to destroy it and then the model can't be removed.

benhoyt: I've added `--destroy-storage`.

.github/workflows/ci.yaml
benhoyt: Interesting idea. I've looked at it both ways locally, and initially I quite liked your terse approach, but then it kinda broke down for `run: sudo concierge prepare --verbose --juju-channel=3/stable --charmcraft-channel=3.x/stable -p microk8s`. I think the nice names are okay, and I don't mind the increased verbosity here, so I'm going to leave them.

Makefile
```diff
@@ -26,7 +26,7 @@ format:
 # Run integration tests (slow, require real Juju)
 .PHONY: integration
 integration:
-	uv run pytest test/integration
+	uv run pytest test/integration -vv --log-level=INFO --log-format="%(asctime)s %(levelname)s %(message)s"
```
benhoyt, line 29: The default pytest log format doesn't include timestamps (weird!).

jubilant/_juju.py
```diff
@@ -221,6 +227,16 @@ def config(
 
         self.cli(*args)
 
+    def debug_log(self, *, limit: int = 0) -> str:
```
benhoyt, line 230: There are a lot more parameters to `debug-log`, but these will do for a start.

test/integration/conftest.py
```diff
@@ -0,0 +1,29 @@
...
+@pytest.fixture
+def juju(request: pytest.FixtureRequest) -> Generator[jubilant.Juju, None, None]:
+    """Pytest fixture that wraps :meth:`jubilant.with_model`.
+
+    This adds command line parameter ``--keep-models`` (see help for details).
+    """
+    keep_models = cast(bool, request.config.getoption('--keep-models'))
+    with jubilant.temp_model(keep=keep_models) as juju:
+        yield juju
+        if request.session.testsfailed:
```
benhoyt, line 27: This is asking whether any tests in the session failed, which isn't quite right, but it seems there's no easy way from the pytest "request" to ask "did this test fail?" Strange.

pyproject.toml
```diff
@@ -4,11 +4,15 @@ description = "Juju CLI wrapper for charm integration testing"
 readme = "README.md"
 requires-python = ">=3.12"
 dynamic = ["version"]
-
 dependencies = [
     "PyYAML==6.*",
 ]
 
+[project.optional-dependencies]
+pytest = [
```
tonyandrewmeyer, line 12: What's the purpose of this optional dependency?

jubilant/_test_helpers.py
```diff
@@ -0,0 +1,27 @@
...
+        keep: If true, keep the created model around when the context manager exits.
+    """
+    juju = Juju()
+    model = 'jubilant-' + secrets.token_hex(4)  # 4 bytes (8 hex digits) should be plenty
+    juju.add_model(model)
+    try:
+        yield juju
+    finally:
+        if not keep:
+            juju.destroy_model(model, destroy_storage=True, force=True)
```
tonyandrewmeyer, line 27: Does this need to have force? Even if the model is in a bad state it should still be able to be non-force destroyed in normal circumstances, right? What's the Juju'er stance on this?

test/unit/test_test_helpers.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+
+    with jubilant.temp_model() as juju:
+        assert juju.model == 'jubilant-abcd1234'
+        juju.deploy('app1')
+
+    assert juju.model is None
+    assert len(run.calls) == 3
+    assert run.calls[0].args[1] == 'add-model'
+    assert run.calls[1].args[1] == 'deploy'
```
tonyandrewmeyer, line 35: It makes the arrange/act/assert less clean, but it feels like the test ought to assert that the add and deploy were done when you first enter the context and the destroy is done after exiting.

tonyandrewmeyer (approved):
Looks good! (Quite a bit of this is the "also", but I guess that's no so important before an initial release).

I wonder about force, but maybe you've already worked through that?

Also unsure what the optional extra gives - is that a hangover from when you would provide fixtures, or is it doing something?

Note that I did not try this out - I assume we're still leaving that until a final stage.

jubilant/_test_helpers.py
benhoyt: Yeah, I'm unsure about this. But my thinking was that because this is used in test cleanup, we should do whatever we can to ensure it cleans up after itself, so `--force` seems appropriate here.

test/unit/test_test_helpers.py
benhoyt: Ah yes, good call! Fixed.

pyproject.toml
benhoyt: Oops, yes, a hangover from when I had `jubilant/fixtures.py` and used pytest there. No longer needed -- removed.

benhoyt:
Going to merge this now -- can always adjust before 1.0 release if there are problems.

PR #53, opened 2025-03-12
feat: add basic integration tests for most jubilant.Juju methods

### LINKED ISSUE #31: Add integration tests

benhoyt:
This adds some basic integration tests against a real Juju to test the main flows: deploy, adding and removing units, config, integrate, run (an action).

To make integration testing faster and more stable, this PR also introduces two test charms used just by these integration tests. These are packed from source before running the integration tests.

This PR also fixes an issue with `Juju.run` where it didn't work with the (confined) snap due to `/tmp` not being accessible. `tempfile.NamedTemporaryFile` uses `/tmp`, so we have to try to be smart and detect whether the Juju CLI is the snap. I'm not sure this is the best way, but it's lightweight and seems semi-reasonable.

Fixes #31


test/integration/test_basic.py
```diff
@@ -1,7 +1,62 @@
...
+    juju.config('wordpress-k8s', {'blog_hostname': 'example.com'})
+    config = juju.config('wordpress-k8s')
+    assert config['blog_hostname'] == 'example.com'
+
+
+def test_integrate_and_run(juju: jubilant.Juju):
+    juju.deploy('wordpress-k8s')
+    juju.deploy('mysql-k8s')
+    juju.integrate('wordpress-k8s', 'mysql-k8s')
+    status = juju.wait(jubilant.all_active, timeout=20 * 60)
```
benhoyt, line 51: I'm not happy with this long timeout. mysql-k8s seems like a very heavy charm. Any suggestions for two significantly lighter-weight charms I can integrate?

Makefile
```diff
@@ -1,7 +1,7 @@
 
 .PHONY: help
 help:
-	@echo "Usage: make [target]\n\nTargets:"
+	@echo "Usage: make [target] [ARGS='additional args']\n\nTargets:"
```
benhoyt, line 4: Pulling in these changes from https://github.com/canonical/jubilant/pull/54 (reviewable there)

test/integration/test_basic.py
james-garner-canonical: I don't have any suggestions for real charms, but it makes me wonder if we want two batches of tests here. One with very lightweight example charms to see at a glance that everything *should* be working, and one using heavier real world charms to see if things then fail anyway due to complications.

james-garner-canonical: I assume that Juju itself is supposed to take care of performing the integration only when the two charms have deployed correctly, but I wonder if a wait after the deploys (or after each deploy while debugging) could make it easier to see at a glance what went wrong when a wait fails.

IronCore864:
It seems the GitHub Actions run failed. I tried it locally, for me, deploying `wordpress-k8s` fails: 


```bash
controller-0: 03:14:36 INFO juju.worker.caasapplicationprovisioner.runner non-fatal error "wordpress-k8s": more than 20 retries ensuring trust: updating application "wordpress-k8s" to desired trust false: getting service account role "wordpress-k8s": k8s: roles.rbac.authorization.k8s.io "wordpress-k8s" not found
controller-0: 03:14:36 ERROR juju.worker.caasapplicationprovisioner.runner exited "wordpress-k8s": more than 20 retries ensuring trust: updating application "wordpress-k8s" to desired trust false: getting service account role "wordpress-k8s": k8s: roles.rbac.authorization.k8s.io "wordpress-k8s" not found
```

I can reproduce this even with a manual `juju deploy`, not sure if this is related to the environment. According to the [doc](https://charmhub.io/wordpress-k8s/docs/tutorial) there seems to be no other prerequisites, no need to manually create SAs and stuff.

test/integration/test_basic.py
benhoyt: I've decided to use some small test charms written specifically for these integration tests.

jubilant/_juju.py
```diff
@@ -548,6 +552,19 @@ def wait(
...
+    @functools.cached_property
+    def _juju_is_snap(self) -> bool:
+        which = shutil.which(self.cli_binary)
+        return which is not None and '/snap/' in which
+
+    @functools.cached_property
+    def _temp_dir(self) -> str:
+        if self._juju_is_snap:
+            # If Juju is running as a snap, we can't use /tmp, so put temp files here instead.
+            return os.path.expanduser('~/snap/juju')
```
dimaqq, line 564: This just occurred to me: do we ever want to run jubilant with sudo?

dimaqq: P.S. should this add `current` or `common`?

```
> ls ~/snap/juju/
29965/  30245/  common/  current@
```

Makefile
```diff
@@ -34,6 +34,12 @@ lint:
 	uv run ruff check
 	uv run ruff format --diff
 
+# Pack charms used by integration tests (requires charmcraft)
+.PHONY: pack
+pack:
+	cd test/integration/charms/testdb && charmcraft pack
+	cd test/integration/charms/testapp && charmcraft pack
+
```
james-garner-canonical, line 42: I wonder if, in future, we'd want to maintain some simple test charms, perhaps hosted (unlisted?) on charmhub, for use in our libraries, or even for charmers to use to test relations with? Just ideas, I think this is the right approach for now.

jubilant/_juju.py
```diff
@@ -548,6 +552,19 @@ def wait(
             exc.add_note(str(status))
         raise exc
 
+    @functools.cached_property
+    def _juju_is_snap(self) -> bool:
+        which = shutil.which(self.cli_binary)
+        return which is not None and '/snap/' in which
+
+    @functools.cached_property
+    def _temp_dir(self) -> str:
```
james-garner-canonical, line 561: I wonder if there's any value in having this return a temp directory for jubilant specifically? e.g. `f'{gettempdir()}/jubilant'` or `'~/snap/juju/common/jubilant'` -- though I guess you then have to clean that directory up at the appropriate time ... maybe in `Juju.__del__`?

pyproject.toml
```diff
@@ -89,7 +91,7 @@ builtins-ignorelist = ["id", "min", "map", "range", "type", "input", "format"]
 ]
 
 [tool.pyright]
-include = ["jubilant/*.py", "test/unit/*.py", "test/integration/*.py"]
+include = ["jubilant/*.py", "test/**/*.py"]
```
james-garner-canonical, line 94: Does `include = ["jubilant", "test"]` work?

PS: probably the wrong PR to bring this up, but how about `tests` instead of `test` for the directory? As used in pytest docs [here](https://docs.pytest.org/en/stable/explanation/goodpractices.html#tests-outside-application-code).

test/integration/charms/testapp/src/charm.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python3
+"""Test app ("db" consumer) charm for Jubilant integration tests"""
```
james-garner-canonical, line 2:
```suggestion
"""Test app ("db" consumer) charm for Jubilant integration tests."""
```
Nit: doesn't ruff normally [complain about this](https://docs.astral.sh/ruff/rules/ends-in-punctuation/)?

test/integration/test_basic.py
```diff
@@ -1,7 +1,80 @@
...
 import jubilant
 
+CHARMS_PATH = pathlib.Path(__file__).parent / 'charms'
+
 
 def test_deploy(juju: jubilant.Juju):
     juju.deploy('snappass-test')
+    status = juju.wait(jubilant.all_active)
+
+    address = status.apps['snappass-test'].units['snappass-test/0'].address
```
james-garner-canonical, line 14: I'd be tempted to factor out the charm/app name into a variable. What do you think? A bad idea in tests?

test/integration/charms/testdb/src/charm.py
```diff
@@ -0,0 +1,22 @@
...
+"""Test "db" provider charm for Jubilant integration tests."""
+
+import ops
+
+
+class Charm(ops.CharmBase):
+    def __init__(self, framework: ops.Framework):
+        super().__init__(framework)
+        self.framework.observe(self.on['db'].relation_created, self._on_db_relation_created)
+        self.framework.observe(self.on.do_thing_action, self._do_thing)
```
james-garner-canonical, line 11:
```suggestion
        self.framework.observe(self.on.['do-thing'].action, self._do_thing)
```

james-garner-canonical (approved):
Looks good, no substantial comments really, but I've noted down some minor thoughts.

jubilant/_juju.py
dwilding, line 564: I think `/snap/juju/common` is better here (which is [`SNAP_USER_COMMON`](https://snapcraft.io/docs/environment-variables#heading--snap-user-common) inside the snap).

IronCore864 (approved):

jubilant/_juju.py
benhoyt: Per David's comment further down, I'm going to use `common` (and ensure it's created).

Makefile
benhoyt: Yeah, we could definitely consider it. Nice thing about these very specific ones though, is they can be tailored for the specific tests.

jubilant/_juju.py
benhoyt: Per comments elsewhere, I'm going to use `~/snap/juju/common` and ensure it's created.

I don't think a jubilant sub-dir is necessary for this, as NamedTemporaryFile will ensure it creates a new temp file. Can reconsider later if there's too many or it becomes an issue.

pyproject.toml
benhoyt: Ah yes, `include = ["jubilant", "test"]` does work -- thanks.

Re `tests` dir, I've opened https://github.com/canonical/jubilant/issues/56 to track.

test/integration/charms/testapp/src/charm.py
benhoyt: Probably. I think it's because I have all documentation linting turned off under `test/`, which I think is right. Snippet of `pyproject.toml`:

```
[tool.ruff.lint.per-file-ignores]
"test/*" = [
    # All documentation linting.
    "D",
]
```

test/integration/test_basic.py
benhoyt: Done.

test/integration/charms/testdb/src/charm.py
benhoyt: Done

test/integration/charms/testapp/src/charm.py
james-garner-canonical: Ah yes, of course, that explains it, thanks.

jubilant/_juju.py
benhoyt: Done, thanks.

PR #55, opened 2025-03-13
feat: workflow and metadata for publishing to PyPI

### LINKED ISSUE #21: Add PyPI metadata and publish
This issue represents all the work to get a beta version on PyPI and publish it.

benhoyt:
This PR adds a workflow for (trusted) publishing to PyPI, as well as `pyproject.toml` metadata.

I've tested the `make publish` command locally with the [1.0.0b1](https://pypi.org/project/jubilant/1.0.0b1/) pre-release.

Fixes #21


.github/workflows/ci.yaml
```diff
@@ -62,6 +62,10 @@ jobs:
   integration:
     runs-on: ubuntu-latest
 
+    concurrency:
```
benhoyt, line 65: Drive-by change to cancel previous integration test run for this PR if a dev pushes a bunch of commits in quick succession.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Build dist packages
+        run: uv build
+
+      - name: Attest build provenance
+        uses: actions/attest-build-provenance@v2.2.3
+        with:
+          subject-path: 'dist/*'
+
+      - name: Publish
+        uses: pypa/gh-action-pypi-publish@release/v1
```
tonyandrewmeyer, line 34: It's not needed right now since there aren't (as far as I know) any secrets, and maybe there never will be. But it's probably worth using a separate environment here, which protects against mistakes being made in the future. Particularly given the recent actions attack that hit a lot of people, it seems worth doing the small amount of extra work in advance.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,34 @@
+name: Publish
+
+on:
+  push:
+    tags:
+      - '0.*'
+      - '1.*'
+
+jobs:
```
tonyandrewmeyer, line 9: Do you want a `needs` that has some of the jobs from `ci.yaml`? We have it in ops, and I guess it adds extra reliability, but it also does seem unnecessary because (assuming no-one with the required permissions overrides) those ought to have been ok when merging to main and shouldn't break afterwards.

I haven't thought about this in depth, but I think my preference is what you have here and potentially doing that in ops as well.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,34 @@
+name: Publish
+
+on:
+  push:
+    tags:
+      - '0.*'
+      - '1.*'
```
tonyandrewmeyer, line 7: Where do you sit on the "v or no v" question? GitHub's releases page says:

> It’s common practice to prefix your version names with the letter v. Some good tag names might be v1.0.0 or v2.3.4.

CPython, requests, Juju, numpy, Pydantic, use "v". uv, Django, Jinja, FastAPI, Go do not.

I'm not certain on this, but what I've heard is the original motivation of the "v" prefix was for tooling that was creating the artefacts, and that this is all out of date now.

I like the "no v" look better, but the prominence of the "use v" note in GitHub makes me wonder if that would be better.

Makefile
```diff
@@ -40,6 +40,13 @@ pack:
 	cd tests/integration/charms/testdb && charmcraft pack
 	cd tests/integration/charms/testapp && charmcraft pack
 
+# Publish to PyPI
+.PHONY:
+publish:
+	rm -rf dist
+	uv build
+	uv publish
+
```
tonyandrewmeyer, line 49: Having this during the initial development phase seems reasonable, and I can definitely see an argument that allowing publishing to *test* PyPI with both trusted publishing and a token is reasonable.

However, the recommendation is that you don't allow token based publishing once you're all in on trusted publishing, so I don't think a `make publish` command is a good addition at this stage, unless it was specifically for publishing to *test* PyPI.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+
+      - name: Build dist packages
+        run: uv build
+
+      - name: Attest build provenance
+        uses: actions/attest-build-provenance@v2.2.3
+        with:
+          subject-path: 'dist/*'
```
tonyandrewmeyer, line 31: Feel free to drop this if you'd prefer. There are still advantages over the attestation that's built into the PyPI publish action, but they will be going away as the PyPI system gets improvements, so I think it would be reasonable to just wait for those.

I feel like it's more worth it for ops, which is used in production, to have both for now, than Jubilant, where it seems unlikely that you'd verify your testing dependencies, although you probably ought to.

tonyandrewmeyer (approved):
Looks good. Some questions that you might have already thought through (but mentioning in case) but not ones where I feel any changes are required.

I do feel like it would be worth adding an environment, and I'm sceptical about `make publish`, but happy to go with your calls on those.

.github/workflows/publish.yaml
benhoyt: Good call -- I've added a `publish-pypi` environment in GitHub.

.github/workflows/publish.yaml
benhoyt: Yeah, I thought about that too, and I think because we run all these on every PR (and merge into main), it's not necessary to repeat that here.

.github/workflows/publish.yaml
benhoyt: Good call. I actually like the prefix (Go modules uses it as well) in part because it's easier to filter on, and so easy to see "this is a version tag". Changed to that.

PR #62, opened 2025-03-23
chore: use 24.04 and the uv plug-in in integration test charms

tonyandrewmeyer:
Two changes to the charms packed and used by the integration tests:
 * Use 24.04 as the base instead of 22.04 - this gets a more modern Python, is possibly a little faster, and seems a good choice being the latest LTS
 * Use the `uv` plug-in for `charmcraft` rather than the `charm` one - `uv`, `poetry`, or `python` are the faster and more modern choices

The main motivation is decreasing the time it takes to run the integration tests. In [my fork](https://github.com/tonyandrewmeyer/jubilant/actions/workflows/ci.yaml) my timings were:
* `charm` (current main): 5m34s, 5m43s, 5m46s
* `python`: 5m37s, 5m40s, 5m42s
* `uv`: 2m10s, 2m11s, 2m14s

The times seem quite consistent, with a win of about 3.5m, roughly 25% faster integration tests overall.

If you're open to this change, the CI should probably also refresh the lock files so that it doesn't fall behind on `ops` versions - I'm thinking that `make pack` could run `uv lock --no-config --refresh` (or similar), which should also be fast. I can add that or some other re-locking to the PR.


benhoyt (approved):
This is great, thanks!

I'm not totally opposed to re-locking, but I also don't think it's necessary (testing ops versions isn't really the point of the integration tests). So I'd be happy for this to stay manual too, or even specify `ops==2.19.4` in pyproject.toml. I'll leave it up to you.

benhoyt:
I've put a (small!) item on next cycle's roadmap to do this for the `charmcraft init` templates too.

tonyandrewmeyer:
> I'm not totally opposed to re-locking, but I also don't think it's necessary (testing ops versions isn't really the point of the integration tests). So I'd be happy for this to stay manual too, or even specify `ops==2.19.4` in pyproject.toml. I'll leave it up to you.

Ok, let's leave it as-is then.

PR #65, opened 2025-03-25
docs: make example fixtures module-scoped, explain Python 3.12 requirement

benhoyt:
This tweaks the docs in two ways (per feedback from @jnsgruk):

* Make the example fixtures in the README and docs *module*-scoped, to match what people are used to with pytest-operator (its fixtures are module-scoped).
* Prominently note the Python 3.12 requirement, as well as how to manage it if your charms are using on older version of Python (older Ubuntu base). I realise this is duplicated, but I think it's important enough to do in both README and docs tutorial.

### DIFF
--- a/docs/tutorial/getting-started.md
+++ b/docs/tutorial/getting-started.md
@@ -62,7 +62,7 @@ Model "test" is empty.
 
 We recommend using [pytest](https://docs.pytest.org/en/stable/) for writing tests. You can define a [pytest fixture](https://docs.pytest.org/en/stable/explanation/fixtures.html) to create a temporary Juju model for each test. The [](jubilant.temp_model) context manager creates a randomly-named model on entry, and destroys the model on exit.
 
-Here is a module-scope fixture called `juju`, which you would normally define in [`conftest.py`](https://docs.pytest.org/en/stable/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files):
+Here is a module-scoped fixture called `juju`, which you would normally define in [`conftest.py`](https://docs.pytest.org/en/stable/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files):
 
 ```python
 @pytest.fixture(scope='module')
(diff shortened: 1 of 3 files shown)


benhoyt:
@dwilding Spelling checker failed on "pytest’s" -- it's odd that `pytest's` gets flagged when `pytest` is in our custom wordlist. I wonder ... is the spelling checker causing more problems than it's solving?

benhoyt:
My shower thought last night was "I bet we could write a better/faster spell checker as a Hackathon project". :-)

docs/tutorial/getting-started.md
```diff
@@ -78,10 +80,10 @@ def test_deploy(juju: jubilant.Juju):
     assert status.apps['snappass-test'].scale == 1
 ```
 
-You may want to make your fixture [module-scoped](https://docs.pytest.org/en/stable/how-to/fixtures.html#scope-sharing-fixtures-across-classes-modules-packages-or-session), to allow all the tests in one file to share the same temporary model:
+You may want to adjust the [scope](https://docs.pytest.org/en/stable/how-to/fixtures.html#fixture-scopes) of your `juju` fixture. For example, if you want to create a new model for every test function, use `scope='function'`, or just omit the scope (function scope is pytest's default):
```
dwilding, line 83: I'd probably simplify this slightly:

```suggestion
You may want to adjust the [scope](https://docs.pytest.org/en/stable/how-to/fixtures.html#fixture-scopes) of your `juju` fixture. For example, to create a new model for every test function (pytest's default behavior), omit the scope:
```

dwilding:
> My shower thought last night was "I bet we could write a better/faster spell checker as a Hackathon project". :-)

I believe we do have a new spellchecker ([vale](https://vale.sh/docs/checks/spelling)) coming at some point.

Adding `pytest's` to the custom wordlist seems to work, but I agree it's annoying. For now, I don't have a problem if you want to remove the spellcheck from CI. I advised James to do that with pathops.

docs/tutorial/getting-started.md
benhoyt: Good call, done

benhoyt:
> Adding pytest's to the custom wordlist seems to work, but I agree it's annoying. For now, I don't have a problem if you want to remove the spellcheck from CI. I advised James to do that with pathops.

Good idea. I'll remove it for now, and we can add it back later once we have better tooling.

dwilding (approved):

dimaqq:
belated LGTM

PR #66, opened 2025-03-25
feat!: add exec method

### LINKED ISSUE #7: Add exec command
Signature should be something like:

```python
def exec(self, *, apps: list[str] | None = None, units: list[str] | None = None,
         machines: list[str] | None = None, all: bool = False):
```

TODO:

- Need to figure out what to do with the command output. And parallel/wait/execution-group.
- Do we need to support `--operator`, or is that only for legacy podspec charms?

For reference, discourse-k8s tests use exec pretty heavily.

benhoyt:
This adds `Juju.exec`.

This PR also renames `ActionResult` to `Task` (breaking change), as that's the Juju name for that entity, and it's used for both actions and run/exec. It also adds a `Task.raise_on_failure` helper method, and converts `success` to a calculated property.

Similar to `Juju.run`, it doesn't currently support multiple targets (multiple units, machines, or an app), as the return value would be a different shape (`dict[str, Task]`, and no raise `TaskError`). We can consider adding `exec_multi` or similar later.

Fixes #7.


jubilant/_juju.py
```diff
@@ -347,6 +347,72 @@ def destroy_model(
...
+        result = None
+        if machine is not None:
+            if str(machine) not in results:
+                raise ValueError(f'machine not found: {machine}')
+            result = results[str(machine)]
+        elif unit:
+            if unit not in results:
+                raise ValueError(f'unit not found: {unit}')
+            result = results[unit]
+        assert result is not None
```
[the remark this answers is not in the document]
james-garner-canonical: Alternatively, I wonder here if it might be nicer to explicitly handle the no results case separately if we get no stdout for the error cases?
```py
results = ...
if not results:
    if machine is not None:
        raise ValueError(f'machine not found: {machine}')
    raise ValueError(unit not found: {unit}')
if machine is not None:
    result = results[str(machine)]
else:
    assert unit
    result = results[unit]
```

jubilant/_juju.py
```diff
@@ -347,6 +347,72 @@ def destroy_model(
...
+        args.append('--')
+        args.extend(command)
+
+        try:
+            stdout = self.cli(*args)
+        except CLIError as exc:
+            # The "juju exec" CLI command itself fails if the exec'd command fails.
+            if 'task failed' not in exc.stderr:
+                raise
+            stdout = exc.stdout
```
james-garner-canonical, line 398: Does our error handling get easier if we do something like this?
```py
try:
    stdout, stderr = self._cli(*args)
except CLIError as exc:
    if 'task failed' not in exc.stderr:
        raise
    stdout = exc.stdout
    stderr = exc.stderr
```
And then later ...
```py
results = ...  # parse stdout
if not results:
    raise ValueError(stderr)  # maybe with some light parsing

File: tests/integration/test_basic.py [hunk 3]
```diff
@@ -50,15 +53,34 @@ def test_config_and_run(juju: jubilant.Juju):
...
+    assert task.success
+    assert task.stdout == 'bar baz\n'
+
+    with pytest.raises(jubilant.TaskError) as excinfo:
+        juju.exec('sleep x', unit=charm + '/0')
+    task = excinfo.value.task
+    assert not task.success
+    assert task.stdout == ''
+    assert 'invalid time' in task.stderr
+
```
james-garner-canonical, line 83: Similar to #67, do you think it would be good to exercise the other error paths here?
- machine/unit not found shouldn't `CLIError` but should result in a `ValueError` when handling results
- one or more things that raise a `CLIError`, like a bad value for the machine id argument (e.g. machine='foo'), or using the machine argument on a k8s charm with a good value (e.g. machine=0)

tests/unit/test_exec.py
```diff
@@ -0,0 +1,141 @@
...
+    run.handle(
+        ['juju', 'exec', '--format', 'json', '--unit', 'ubuntu/0', '--', 'sleep x'],
+        returncode=2,
+        stdout='OUT',
+        stderr='ERR',
+    )
+    juju = jubilant.Juju()
+
+    with pytest.raises(jubilant.CLIError) as excinfo:
+        juju.exec('sleep x', unit='ubuntu/0')
```
james-garner-canonical, line 115: It's probably silly of me, because the actual `juju` call is being mocked out, but it felt odd to see the same bad command example of 'sleep x' being used for both the `TaskError` and `CLIError` case.

I made a similar comment on #67, but a comment or assertion on the part of the mocked value that makes the difference might be good -- `assert 'task failed' (not) in stderr`. Maybe the returncode should be the same in both tests to make it more clear that it doesn't matter.

tests/unit/test_exec.py
```diff
@@ -0,0 +1,141 @@
...
+    juju = jubilant.Juju()
+
+    with pytest.raises(ValueError):
+        juju.exec('echo', unit='u/0')
+
+
+def test_no_target_arg():
+    juju = jubilant.Juju()
+    with pytest.raises(TypeError):
+        juju.exec('echo')  # type: ignore
```
james-garner-canonical, line 141: Worth adding a test for calling with both the `machine` and `unit` args provided?

jubilant/_juju.py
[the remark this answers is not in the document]
james-garner-canonical: Ultimately I wonder if this really means that we need a way (internally) to get both stdout and stderr in the case where the Juju cli succeeds (0 exit code), but doesn't really succeed (no results in stdout, reason given in stderr).

I guess the most Juju-cli-following thing to do here would be to return `None` in the "empty stdout but 0 exit code" case, but I think raising an error is the right call.

jubilant/_task.py
```diff
@@ -40,30 +37,39 @@ class ActionResult:
...
+            id=d['id'],
             status=status,
             results=results,
             return_code=return_code,
             stdout=stdout,
             stderr=stderr,
             message=d.get('message') or '',
             log=d.get('log') or [],
-            task_id=d['id'],
         )
```
james-garner-canonical, line 55: nit: handle `id` and `status` args the same way now that `status` isn't needed for `success`? e.g. move `status=d['status']` to line 48.

james-garner-canonical (commented):
Nice, `TaskResult/Error` seems like a good idea. A couple of substantial comments around error handling -- small issue of covering the case where both `machine` and `unit` args are provided, but more broadly it seems like error handling might be simplified here by having access to Juju's stderr in the non-`CLIError` case.

jubilant/_task.py
benhoyt: Done, thanks.

tests/unit/test_exec.py
benhoyt: Done.

tests/unit/test_exec.py
benhoyt: Fair -- done (the second paragraph with a comment, similar to previous PR).

tests/integration/test_basic.py
benhoyt: Good call -- I've added a few more. Though for integration tests I'm not aiming for 100% coverage. I'm also not testing against lxd/machines in integration tests. May add that later, but K8s will do for now.

jubilant/_juju.py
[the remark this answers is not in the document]
benhoyt: I've simplified using an if/else like you suggest in your first comment.

I decided against having an internal `_cli` that returns stdout and stderr, as I don't want to be in the business of parsing and relying on the "shape" of Juju's stderr. I've done that out of necessity for "task failed", but I want to do it as little as possible. Instead, I'm just logging a warning when there's unexpected stderr, so hopefully that is enough when diagnosing these issues.

jubilant/_juju.py
benhoyt: See above.

jubilant/_juju.py
```diff
@@ -347,6 +347,72 @@ def destroy_model(
...
+            ValueError: if the machine or unit doesn't exist.
+            TaskError: if the command failed.
+        """
+        args = ['exec', '--format', 'json']
+        if machine is not None:
+            args.extend(['--machine', str(machine)])
+        elif unit:
+            args.extend(['--unit', unit])
+        else:
+            raise TypeError('must specify "machine" or "unit"')
```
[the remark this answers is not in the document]
benhoyt: Juju actually allows both machine and unit -- it's kind weird, but it works, and it returns both results. However, I'm not going to allow that here (as there's only one result). I've simplified somewhat here and done the TypeError checking at the top.

benhoyt:
@james-garner-canonical Ready for re-review. I've decided to handling the unexpected stderr a different way -- just log a warning and include the stderr. Let me know what you think of this compromise.

james-garner-canonical (approved):
Hopefully logging is sufficient if people run into this in their integration tests -- I like this a lot better than not surfacing the information at all.

Are we expecting this to only happen in cases where we raise an exception anyway? If not, maybe it could be a `logger.error` instead of a `logger.warning`.

If we are thinking it might happen all the time though, then `logger.error` is probably a bit noisy -- but maybe a sign that we should be figuring out when it's an error ourselves (e.g. no stdout in this case).

benhoyt:
Yeah, I think you're right -- I've changed it to `logger.error` as we only expect this if we raise an error anyway.

benhoyt:
Hmm, no, that's not correct at all. I'll think about this some more tomorrow:

```python
>>> j.run('testdb/0', 'do-thing')
cli: unexpected stderr running juju run --format json testdb/0 do-thing:
Running operation 31 with 1 task
  - task 32 on unit-testdb-0

Waiting for task 32...

Task(id='32', status='completed', results={'config': {'testoption': ''}, 'thingy': 'foo'}, return_code=0, stdout='', stderr='', message='', log=[])
```

james-garner-canonical (approved):
Look good to me!

jubilant/_juju.py
```diff
@@ -178,11 +183,7 @@ def cli(self, *args: str, include_model: bool = True) -> str:
             )
         except subprocess.CalledProcessError as e:
             raise CLIError(e.returncode, e.cmd, e.stdout, e.stderr) from None
-        if process.stderr:
-            logger.error(
-                'cli: unexpected stderr running juju %s:\n%s', shlex.join(args), process.stderr
-            )
-        return process.stdout
+        return (process.stdout, process.stderr)
```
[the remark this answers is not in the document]
benhoyt: Yeah, I can handle either. I somewhat prefer the explicit parens as IMO it reads more clearly that we're returning a tuple in many cases.

PR #73, opened 2025-04-01
fix: wait_timeout 3m -> 10m to match python-libjuju wait_for_idle's

benhoyt:
Changing the default timeout to match python-libjuju's `wait_for_idle` default, just to make porting to Jubilant easier.

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -77,7 +77,7 @@ class Juju:
         self,
         *,
         model: str | None = None,
-        wait_timeout: float = 3 * 60.0,
+        wait_timeout: float = 10 * 60.0,
         cli_binary: str | os.PathLike[str] | None = None,
     ):
         self.model = model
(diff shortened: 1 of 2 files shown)


tonyandrewmeyer (approved):
I think I argued against such a long time originally. 10 minutes still feels like a long time to me, particularly if it's "I made a dumb mistake and it's going to time out". I guess Jubilant's behaviour of showing what's happening (or not) during the wait helps with that.

I do still feel that a lower default is a better choice, because people will easily increase the time when their specific case needs more time, but people are unlikely to decrease the time if the extra time isn't required. It also gives a "doing things in Juju takes a long time" vibe, but maybe that's actually (unfortunately) true.

My preference would be that when charms port their tests, they take some time (which they'll get back in shorter waits :wink:) to consider what the right magic number for waiting is in each case, rather than just always using the default provided.

However, I can see an argument that we just want to make migrating as easy as possible, so being consistent with py-libjuju does that.

TL:DR: I'm -0 on this, so approving with a mini rant :laughing:.

benhoyt:
Actually, now that I think about it, you can just put `jubilant.Juju(wait_timeout=10*60)` in your conftest.py fixture when porting ... so yeah, maybe it is better to just keep the class-level default at 3m.

PR #74, opened 2025-04-01
fix!: rename deploy param resource -> resources

benhoyt:
This is based on feedback from Jon that the singular is just weird here as it's always a dict, never a single item. I think this is fair, and we don't have to be quite so pedantic about the 1:1 match to the CLI arg.


james-garner-canonical (approved):
Makes sense, looks good. Approving before integration tests finish as they're unchanged (maybe an integration test that uses `resources` would be good?).

benhoyt:
Yep, fair call. I've opened https://github.com/canonical/jubilant/issues/75 to track.

PR #78, opened 2025-04-03
feat: add remove_relation method

benhoyt:
Add the `Juju.remove_relation` method for removing a relation (the opposite of `integrate`). Type signature is similar to `integrate`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 
 # We're using Make as a command runner, so always make (avoids need for .PHONY)
-MAKEFLAGS += "--always-make"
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -1,6 +1,6 @@
 
 # We're using Make as a command runner, so always make (avoids need for .PHONY)
-MAKEFLAGS += "--always-make"
+MAKEFLAGS += --always-make
```
benhoyt, line 3: The quotes around here made this not actually work, oops.

IronCore864 (approved):

PR #79, opened 2025-04-03
feat: condense Juju.wait status logging by using gron-style diff

### LINKED ISSUE #63: Condense wait status logging, perhaps by using gron-style diff?
Currently, when deploying a large model and then waiting with `Juju.wait`, the "status changed" logging is huge, tens of hundreds of thousands of lines long. We need to reduce the volume, perhaps with a gron-style diff.

benhoyt:
The status logging from a long `Juju.wait` sequence, especially for a larger deployment, ends up being huge. Hopefully this will shrink it down significantly. The first status is shown in full, but then each time the status changes, we only show a diff of changed lines (inspired by gron). For example:

```
INFO - wait: status changed:
- .apps['database'].app_status.current = 'active'
+ .apps['database'].app_status.current = 'waiting'
- .apps['database'].relations['db'][0].scope = 'global'
- .apps['database'].relations['db'][1].related_app = 'dummy'
- .apps['database'].relations['db'][1].interface = 'xyz'
- .apps['database'].relations['db'][1].scope = 'foobar'
+ .apps['database'].relations['db'][0].scope = 'testy'"""
```

[Here's a gist](https://gist.github.com/benhoyt/e0067b641461b45158ab83ee9077f568) that shows output from the Jubilant integration test run of `test_charm_basics`.

Fixes #63


tests/unit/test_pretty.py
```diff
@@ -150,3 +150,28 @@ def test_large():
...
+    new_json['applications']['database']['relations']['db'][0]['scope'] = 'testy'
+    del new_json['applications']['database']['relations']['db'][1]
+    old_status = jubilant.Status._from_dict(old_json)
+    new_status = jubilant.Status._from_dict(new_json)
+
+    diff = jubilant._juju._status_diff(old_status, new_status)
+
+    assert (
+        diff
+        == """\
```
dimaqq, line 169: nitpick: I'm allergic to line continuation characters, so I'd personally do this instead:

```
assert foo == """
text
text
""".strip()
```

dimaqq: Some folks like to indent such test vectors and dedent on the fly.

Others move the expected value to a module-level constant to make the test look nicer.

I don't particularly care, just shooting ideas 🙇🏻

jubilant/_juju.py
```diff
@@ -695,3 +695,12 @@ def _format_config(k: str, v: ConfigValue) -> str:
     if isinstance(v, bool):
         v = 'true' if v else 'false'
     return f'{k}={v}'
+
+
+def _status_diff(old: Status | None, new: Status) -> str:
+    if old is None:
+        return '\n'.join('+ ' + x for x in _pretty.gron(new))
```
dimaqq, line 702:
```suggestion
        return '\n'.join(f'+ {x}' for x in _pretty.gron(new) if not x.startswith('.controller.timestamp'))
```

Or maybe just set `old_lines` to an empty list and let the differ do the formatting.

jubilant/_juju.py
```diff
@@ -654,7 +654,7 @@ def wait(
             prev_status = status
             status = self.status()
             if status != prev_status:
-                logger.info('wait: status changed:\n%s', status)
+                logger.info('wait: status changed:\n%s', _status_diff(prev_status, status))
```
dimaqq, line 657: Do we always exit hard if `juju status` fails?

I've only seen that when a model is removed, but I wonder about the potential of other corner cases.

dimaqq (approved):

dimaqq:
I wonder if it's cleaner to filter out the `xxx.juju_status.since` lines.

tests/unit/test_pretty.py
benhoyt: That's not bad, though I prefer the direct approach with the `\` (James suggested it originally) rather than dedent or strip, so I'm going to leave this as is. It's also consistent with the other such strings above.

jubilant/_juju.py
benhoyt: Good idea - done.

jubilant/_juju.py
benhoyt: Well, if `juju status` fails we'll raise a `CLIError`. I think that's okay, because like you say, it should only happen if the model's torn down (which will normally only happen after the test finishes). If this is a problem we can always revisit.

benhoyt:
> I wonder if it's cleaner to filter out the {model,app,juju}_status.since lines.

Good idea -- done.

jnsgruk:
Love this, excellent work 💪

PR #80, opened 2025-04-04
fix: allow the library to work on Python 3.8+

benhoyt:
This includes:

- mostly typing tweaks
- avoid use kw_only=True in dataclasses
- avoid use of Exception.add_note in Juju.wait

We've decided we lose almost nothing by having the minimum set back to Python 3.8, and it doesn't add any serious ugliness to the code. It's going to cause charmers who start using Jubilant less annoyance, so just go ahead and do it.


pyproject.toml
```diff
@@ -31,14 +31,10 @@ dev = [
     "ops==2.19.4",  # for static type checking test charms
     "pyright==1.1.398",
     "pytest==8.3.4",
-    "pytest-cov==6.0.0",
+    "pytest-cov==5.0.0",
     "requests>=2.32.3",
     "ruff==0.9.6",
 ]
-docs = [
```
benhoyt, line 38: This wasn't actually being used -- `make docs` uses its own venv handling in `docs/Makefile`.

jubilant/_all_any.py
```diff
@@ -116,7 +116,7 @@ def any_waiting(status: Status, apps: Iterable[str] | None = None) -> bool:
 
 
 def _all_statuses_are(expected: str, status: Status, apps: Iterable[str] | None) -> bool:
-    if isinstance(apps, str | bytes):
+    if isinstance(apps, (str, bytes)):
```
james-garner-canonical, line 119: I didn't even realise that worked at runtime nowadays, neat

jubilant/_juju.py
```diff
@@ -39,7 +39,7 @@ class SecretURI(str):
     """A string subclass that represents a secret URI ("secret:...")."""
 
 
-type ConfigValue = bool | int | float | str | SecretURI
+ConfigValue = Union[bool, int, float, str, SecretURI]
```
james-garner-canonical, line 42: I typically write this as `ConfigValue: TypeAlias = 'bool | int | float | str | SecretURI'` here (importing `TypeAlias` as a `TYPE_CHECKING` only import from `typing_extensions` since it's only available after 3.10) ... but I'm not sure if that's even worthwhile anymore since `TypeAlias` is now officially deprecated in favour of `type` statements.

jubilant/_yaml.py
```diff
@@ -13,7 +13,7 @@ class SupportsRead(Protocol[_T_co]):
     def read(self, length: int = ..., /) -> _T_co: ...
 
 
-type _ReadStream = str | bytes | SupportsRead[str] | SupportsRead[bytes]
+_ReadStream = Union[str, bytes, SupportsRead[str], SupportsRead[bytes]]
```
james-garner-canonical, line 16: See previous comment

james-garner-canonical (approved):
Looks good to me, I didn't notice any blockers

benhoyt:
Thanks @james-garner-canonical. I've released this is Jubilant v0.3.0 if you want to use it in pathops.

PR #81, opened 2025-04-05
feat: add trust command

### LINKED ISSUE #15: Add trust command
Signature something like:

```python
def trust(self, app: str, *, remove: bool = False, scope: str | None = None):
    """Sets the trust status of a deployed application."""
```

benhoyt:
This PR adds support for `juju trust`.

Fixes #15.


james-garner-canonical (approved):
Looks good to me. I think handling the `scope` argument like this makes sense in terms of following the CLI and extensibility with Juju updates (which the CLI flag is clearly designed to facilitate).

PR #82, opened 2025-04-05
feat: add offer command

### LINKED ISSUE #9: Add offer command
Signature something like:

```python
def offer(self, app: str, endpoints: list[str], *, name: str | None = None):
```

Note that `app` may be "model.app" like `db2:db`. So `juju offer model.db2:db,log hosted-db2` is equivalent to `juju.offer('mdl.db2', ['db', 'log'], name='hosted-db2')`.

benhoyt:
This PR adds support for `juju offer`.

```python
def offer(self, app: str, *endpoint: str, name: str | None = None) -> None: ...

juju.offer('mysql', 'db', name='offr_name')
```

Fixes #9

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -456,6 +456,27 @@ class Juju:
                 args.extend(['--via', ','.join(via)])
         self.cli(*args)
 
+    def offer(self, app: str, endpoint: str | Iterable[str], *, name: str | None = None) -> None:
+        """Offer application endpoints for use in other models.
+
+        Args:
+            app: Application name to offer endpoints for.
+            endpoint: Endpoint or endpoints to offer, for example ``'db'`` or ``['db', 'log']``.
+            name: Name of the offer. By default, the offer is named after the application.
+        """
+        if ':' in app:
+            raise TypeError('must provide endpoint in "endpoint" argument not in "app"')
+
+        args = ['offer']
+        if isinstance(endpoint, str):
+            args.append(f'{app}:{endpoint}')
+        else:
+            args.append(f'{app}:{",".join(endpoint)}')
+        if name is not None:
+            args.append(name)
+
+        self.cli(*args)
+
     def remove_relation(self, app1: str, app2: str, *, force: bool = False) -> None:
         """Remove an existing relation between two applications (opposite of :meth:`integrate`).
 
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -463,22 +463,26 @@ class Juju:
                 args.extend(['--via', ','.join(via)])
         self.cli(*args)
 
-    def offer(self, app: str, endpoint: str | Iterable[str], *, name: str | None = None) -> None:
+    def offer(self, app: str, *endpoint: str, name: str | None = None) -> None:
         """Offer application endpoints for use in other models.
+
+        Examples::
+
+            juju.offer('mysql', 'db')
+            juju.offer('mymodel.mysql', 'db', 'log', name='altname')
 
         Args:
             app: Application name to offer endpoints for.
-            endpoint: Endpoint or endpoints to offer, for example ``'db'`` or ``['db', 'log']``.
+            endpoint: Endpoint or endpoints to offer.
             name: Name of the offer. By default, the offer is named after the application.
         """
         if ':' in app:
             raise TypeError('must provide endpoint in "endpoint" argument not in "app"')
+        if not endpoint:
+            raise TypeError('must provide at least one endpoint')
 
-        args = ['offer']
-        if isinstance(endpoint, str):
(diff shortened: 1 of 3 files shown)


tests/unit/test_wait.py
```diff
@@ -109,3 +109,13 @@ def test_timeout_override(run: mocks.Run, time: mocks.Time):
     assert len(run.calls) == 5
     assert time.monotonic() == 5
     assert 'mdl' in str(excinfo.value)
+
+
+def test_timeout_zero(time: mocks.Time):
```
benhoyt, line 114: This gets the coverage of `_juju.py` back up to 100%.

james-garner-canonical (approved):
In general  I'm in favour of being a bit more Pythonic than the cli where possible.

Consistency would be nice though. I can see why it makes sense for `integrate` to keep things in the string form, given that there are two positional arguments with optional prefixes and suffixes. I wonder if the consistency consideration should be stronger in this case, since `offer` and `integrate` have a certain amount of symmetry. 

Here are a couple of ideas, but ultimately it seems like it's all just tradeoffs, and the one you've made currently might be the best one.

1. Make `endpoint: str | Iterable[str]` optional, and allow `:` in `app` if `endpoint` is `None`, similar to how an endpoint of `"foo,bar"` ends up working the same way as `["foo", "bar"]`. 

2. Make it `*endpoint: str`, so that it's called like `offer('mysql', 'db', 'log')` instead of `offer('mysql', ['db', 'log'])`, which is flatter and a little closer to the CLI. This also simplifies the typing and eliminates the need to do `isinstance` for `endpoint`. The major downside I see is that it makes adding further positional args harder in future, and makes it possible to call `offer` with no endpoint (though this could be fixed with `endpoint: str, *more_endpoints: str`).

tests/unit/test_wait.py
dimaqq: Lovely!

The test could be better though...
For one, the test essentially relies on `[time.monotonic(), time.monotonic()]` to produce distinct values.
That's not guaranteed in Python, even if typically true.

jubilant/_juju.py
```diff
@@ -456,6 +456,27 @@ def integrate(self, app1: str, app2: str, *, via: str | Iterable[str] | None = N
...
+    def offer(self, app: str, endpoint: str | Iterable[str], *, name: str | None = None) -> None:
+        """Offer application endpoints for use in other models.
+
+        Args:
+            app: Application name to offer endpoints for.
+            endpoint: Endpoint or endpoints to offer, for example ``'db'`` or ``['db', 'log']``.
+            name: Name of the offer. By default, the offer is named after the application.
+        """
+        if ':' in app:
+            raise TypeError('must provide endpoint in "endpoint" argument not in "app"')
```
dimaqq, line 468: Is the departure from CLI shorthand intentional?
I had just done `juju offer tempo:tracing` yesterday.

benhoyt: Yeah, but I'm unsure about it -- see the PR description. I think I might switch it back, especially given your feedback. F-strings are so easy to use these days if you do have them as separate variables: `f'{app}:{endpoint}'` (though it seems much more likely it'll be hard-coded).

tests/unit/test_wait.py
benhoyt: No, I don't think it does. For one thing, we're using my little time mock, which only increments when you do a `time.sleep`. So both `time.monotonic()`s are guaranteed to return 0. Then the `while` condition is `time.monotonic() - start < timeout`, which is `while 0 - 0 < 0`, so the while loop never executes and it falls down to the `status is None` case.

jubilant/_juju.py
tonyandrewmeyer: My thoughts are pretty identical to @james-garner-canonical 's. A "app:e1,e2,e3" string feels very raw and not in sync with the Jubilant goals. It's easy to combine strings, but it's also easy to split them, so I don't find that compelling either way. `juju.offer('tempo', 'tracing')` seems a very reasonable interpretation to me, not a departure.

I like the `*args` suggestion @james-garner-canonical made, particularly given the use in other places.

It is inconsistent with integrate, but it's different anyway since you can only have a single endpoint per app in that case. It's also less common to need to specify an endpoint there, whereas you *have* to have one here.

tonyandrewmeyer (approved):
I like this as-is, although I think the *args suggestion would be nicer still. I'm not totally opposed to the "make it a raw string" choice, but I definitely think this is better.

jubilant/_juju.py
benhoyt: Cool, I've gone with `*endpoint`, so `juju.offer('tempo', 'tracing')`.

Issue #14, opened 2025-01-22
Add ssh command


benhoyt:
Signature something like:

```python
def ssh(self, target: str, *command: str, user: str | None = None, ssh_options: list[str] | None = None,
        container: str | None = None, no_host_key_checks: bool = False, remote) -> ???
    """Executes a command using SSH on a Juju machine or container."""
```

TODO:

- Need to figure out output
- Do we need `--remote`, or is that legacy podspec charms only?

PR #83, opened 2025-04-07
feat: add "stdin" parameter to Juju.cli

benhoyt:
This adds the ability to send standard input to `Juju.cli`, for things like sending Postgres passwords to `scp` and `ssh`.

### DIFF
--- a/tests/integration/test_basic.py
+++ b/tests/integration/test_basic.py
@@ -32,12 +32,12 @@ def test_add_and_remove_unit(juju: jubilant.Juju):
     juju.wait(lambda status: jubilant.all_active(status) and len(status.apps[charm].units) == 1)
 
 
-# Tests config get, config set, run, and exec
+# Tests config get, config set, trust, run, exec, and cli with input
 def test_charm_basics(juju: jubilant.Juju):
     charm = 'testdb'
     juju.deploy(charm_path(charm))
 
-    # unit should come up as "unknown"
+    # Unit should come up as "unknown"
     juju.wait(
         lambda status: status.apps[charm].units[charm + '/0'].workload_status.current == 'unknown'
     )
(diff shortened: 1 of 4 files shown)


tests/integration/test_basic.py
```diff
@@ -121,6 +121,10 @@ def test_charm_basics(juju: jubilant.Juju):
     with pytest.raises(jubilant.CLIError):
         juju.exec('echo foo', machine=0)  # unable to target machines with a k8s controller
 
+    # Test cli with input
+    stdout = juju.cli('ssh', '--container', 'charm', charm + '/0', 'cat', stdin='foo')
```
james-garner-canonical, line 125: Just for the record, I see how `juju.ssh` wouldn't add tons of value over this, but I reckon it could still be worth adding. Even just being able to specify `container` as a Python keyword argument would be nice.

tests/unit/mocks.py
```diff
@@ -45,7 +47,7 @@ def __call__(
         assert args_tuple in self._commands, f'unhandled command {args}'
 
         returncode, stdout, stderr = self._commands[args_tuple]
-        self.calls.append(Call(args_tuple, returncode, stdout, stderr))
+        self.calls.append(Call(args_tuple, returncode, input, stdout, stderr))
```
james-garner-canonical, line 50: Speaking of keyword arguments, how about:
```suggestion
        self.calls.append(Call(args_tuple, returncode, stdin=input, stdout=stdout, stderr=stderr))
```

james-garner-canonical (approved):

tests/integration/test_basic.py
```diff
@@ -32,12 +32,12 @@ def test_add_and_remove_unit(juju: jubilant.Juju):
     juju.wait(lambda status: jubilant.all_active(status) and len(status.apps[charm].units) == 1)
 
 
-# Tests config get, config set, run, and exec
+# Tests config get, config set, trust, run, exec, and cli with input
```
benhoyt, line 35: I realise this is getting a bit out of hand. I've opened https://github.com/canonical/jubilant/issues/84 to track: split integration tests into multiple files with a module-scoped fixture.

tests/integration/test_basic.py
benhoyt: Agreed, that's tracked as #14.

tests/unit/mocks.py
benhoyt: Fair -- added.

PR #86, opened 2025-04-07
feat: add remove_application method

dwilding:
This PR adds `Juju.remove_application`.

Fixes #11.

PR #87, opened 2025-04-07
docs: minor rewording in docstring of remove_unit method

dwilding:
This PR makes some minor changes to the docstring of `Juju.remove_unit`, for consistency with #86.

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -505,8 +505,8 @@ class Juju:
                 On Kubernetes models, this is actually the application name (a single string),
                 as individual units are not named; you must use *num_units* to remove more than
                 one unit on a Kubernetes model.
-            destroy_storage: If True, also destroy storage attached to the unit(s).
-            force: Force removal even if unit is in an error state.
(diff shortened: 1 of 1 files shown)


jubilant/_juju.py
```diff
@@ -505,8 +505,8 @@ def remove_unit(
                 On Kubernetes models, this is actually the application name (a single string),
                 as individual units are not named; you must use *num_units* to remove more than
                 one unit on a Kubernetes model.
-            destroy_storage: If True, also destroy storage attached to the unit(s).
-            force: Force removal even if unit is in an error state.
+            destroy_storage: If True, also destroy storage attached to units.
```
benhoyt, line 508: Thanks for this. The removal of "the" feels a bit odd -- was that intentional?

dwilding: It was, but only because I wanted to keep the implication that there might be one or many units. I'd actually be OK with "the units", and we assume that people understand it applies to the one-unit and many-units case. Generally I think it's fine to drop "(s)" in favour of "s" and people will get the point.

benhoyt (approved):

PR #88, opened 2025-04-07
fix: don't log "juju status" calls in Juju.wait()

benhoyt:
This removes the logging of every `cli: juju status ...` call while polling status in `Juju.wait` sequences.

Also removes certain "glitches" where it'd show a status change when only `.since` timestamp lines changed, and would result in blank lines like this:

```
INFO wait: status changed:

INFO ... more logs ...
```

Fix #86

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -685,7 +685,9 @@ class Juju:
             status = Status._from_dict(result)
 
             if status != prev_status:
-                logger.info('wait: status changed:\n%s', _status_diff(prev_status, status))
+                diff = _status_diff(prev_status, status)
+                if diff:
+                    logger.info('wait: status changed:\n%s', diff)
 
             if error is not None and error(status):
                 raise WaitError(f'error function {error.__qualname__} returned false\n{status}')
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -174,12 +174,13 @@ class Juju:
         return stdout
 
     def _cli(
-        self, *args: str, include_model: bool = True, stdin: str | None = None
+        self, *args: str, include_model: bool = True, stdin: str | None = None, log: bool = True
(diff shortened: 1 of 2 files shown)


tests/unit/test_wait.py
```diff
@@ -26,9 +26,8 @@ def test_logging(run: mocks.Run, time: mocks.Time, caplog: pytest.LogCaptureFixt
 
     juju.wait(lambda _: True)
 
-    logs = [r for r in caplog.records if r.msg.startswith('wait:')]
-    assert len(logs) == 1  # only logs on first call or when status changes
-    message = logs[0].getMessage()
+    assert len(caplog.records) == 1  # only logs on first call or when status changes
```
dimaqq, line 29: I'm very happy to see this nit gone!

jubilant/_juju.py
```diff
@@ -678,9 +679,15 @@ def wait(
 
         while time.monotonic() - start < timeout:
             prev_status = status
-            status = self.status()
+
+            stdout, _ = self._cli('status', '--format', 'json', log=False)
```
dimaqq, line 683: At first I thought this was a potential code smell: is self.status() is not longer used, why not remove it too?

I guess the thinking here is that public user-facing Juju.status() is detached from its use in this wait function.

I guess it's fine. I hope the use here (cli/json/from_dict) doesn't diverge from that in `status()` over time.

In any case, this approach is better than propagating `log=False` to "arbitrary" functions like `.status()`. Python-libjuju suffers from a case of that.

👍🏻

dimaqq (approved):

jubilant/_juju.py
benhoyt: Yeah, the idea is that `Juju.status` is exposed separately. I could create an `_status` helper that both call, but it didn't feel worth it to save two lines.

PR #89, opened 2025-04-07
fix!: change remove_unit to use *args

dwilding:
This PR changes the signature of `Juju.remove_unit` to use `*args` instead of distinguishing between a string or an iterable. This matches the approach for `Juju.exec`.

I've added an extra unit test to guard against an off-by-one error.

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -508,8 +508,7 @@ class Juju:
             force: Force removal even if a unit is in an error state.
             num_units: Number of units to remove (Kubernetes models only).
         """
-        args = ['remove-unit', '--no-prompt']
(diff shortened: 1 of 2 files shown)


benhoyt (approved):

benhoyt:
Thanks for this. I've also opened https://github.com/canonical/jubilant/issues/90 to track reviewing the entire API before the 1.0 release.

benhoyt:
Would be good to get @tonyandrewmeyer's take on this too.

tonyandrewmeyer (approved):
I like it - removing the isinstance check is nice, and this feel closer to the CLI.

I wish the CLI was designed differently, but given what we have, I think this is a good improvement.

PR #92, opened 2025-04-10
feat: add ssh command

### LINKED ISSUE #14: Add ssh command

benhoyt:
This PR adds support for "juju ssh":

    def ssh(
        self,
        target: str | int,
        command: str,
        *args: str,
        container: str | None = None,
        host_key_checks: bool = True,
        ssh_options: Iterable[str] = (),
        user: str | None = None,
    ) -> str:

    >>> juju.ssh('snappass-test/0', 'echo', 'foo')
    'foo\n'

The "user" is a separate argument, not stuffed into the target as on the CLI.

One decision point was the host_key_checks argument. I lean towards what I have here, but we could also do `no_host_key_checks: bool = False`, which is more 1:1 with the CLI. It just felt weird to have an inverted "no_foo" parameter when we can just invert the default in Python.

Fixes #14

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -644,22 +644,19 @@ class Juju:
 
     def ssh(
         self,
+        target: str | int,
         *command: str,
-        machine: int | None = None,
-        unit: str | None = None,
         container: str | None = None,
         host_key_checks: bool = True,
-        ssh_options: list[str] | None = None,
+        ssh_options: Iterable[str] = (),
         user: str | None = None,
     ) -> str:
         """Executes a command using SSH on a machine or container and returns its standard output.
 
-        You must specify either *machine* or *unit*, but not both.
-
         Args:
+            target: Where to run the command; this is a unit name such as ``mysql/0`` or a machine
+                ID such as ``0``.
             command: Command to run, along with its arguments.
-            machine: ID of machine to run the command on.
-            unit: Name of unit to run the command on, for example ``mysql/0`` or ``mysql/leader``.
             container: Name of container for Kubernetes charms. Defaults to the charm container.
             host_key_checks: Set to False to disable host key checking (insecure).
             ssh_options: OpenSSH client options, for example ``['-i', '/path/to/private.key']``.
@@ -667,25 +664,17 @@ class Juju:
         """
         if not command:
             raise TypeError('must provide a command')
-        if (machine is not None and unit is not None) or (machine is None and unit is None):
-            raise TypeError('must specify "machine" or "unit", but not both')
 
         args = ['ssh']
         if container is not None:
             args.extend(['--container', container])
         if not host_key_checks:
             args.append('--no-host-key-checks')
-
-        target = user + '@' if user is not None else ''
-        if machine is not None:
-            target += str(machine)
+        if user is not None:
+            args.append(f'{user}@{target}')
         else:
-            assert unit is not None
-            target += unit
-        args.append(target)
-
-        if ssh_options:
-            args.extend(ssh_options)
+            args.append(str(target))
+        args.extend(ssh_options)
         args.extend(command)
 
         return self.cli(*args)
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -274,7 +274,7 @@ class Juju:
         Args:
             charm: Name of charm or bundle to deploy, or path to a local file (must start with
                 ``/`` or ``.``).
-            app: Optional application name within the model; defaults to the charm name.
+            app: Optional application name within the model. Defaults to the charm name.
             attach_storage: Existing storage(s) to attach to the deployed unit, for example,
                 ``foo/0`` or ``mydisk/1``. Not available for Kubernetes models.
             base: The base on which to deploy, for example, ``ubuntu@22.04``.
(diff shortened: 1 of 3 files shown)


jubilant/_juju.py
```diff
@@ -642,6 +642,54 @@ def run(
...
+    def ssh(
+        self,
+        *command: str,
+        machine: int | None = None,
+        unit: str | None = None,
+        container: str | None = None,
+        host_key_checks: bool = True,
+        ssh_options: list[str] | None = None,
+        user: str | None = None,
+    ) -> str:
```
james-garner-canonical, line 654: I'm also wondering if it might be more pythonic to allow environment variables for the command to be passed as a dictionary instead of in the command, though I don't think this is necessary for the current PR.

If it was imeplemented then this command:
```bash
juju ssh my-charm/0 VAR1=foo VAR2=bar command arg
```
could become:
```python
juju.ssh('my-charm/0', 'command', 'arg', env={'VAR1': 'foo', 'VAR2': 'bar'})
```
instead of:
```python
juju.ssh('my-charm/0', 'VAR1=foo', 'VAR2=bar', 'command', 'arg')
```
As long as the latter form continued to work, this would be a backwards compatible change, so an `env` argument could always be added later if it seems useful.

james-garner-canonical (commented):
Looks good. I have a few suggestions. In particular, I think that the `machine`/`unit` args would probably be simpler all round as a `target` positional arg, and that the typing for `ssh_options` would be both simpler and more permissive as `Sequence[str]` rather than `list[str] | None`.

jubilant/_juju.py
benhoyt: Thanks. Yeah, I think I'm going to skip this for now. There's a straight-forward workaround. We can always add it later if people are using lots of env vars.

jubilant/_juju.py
```diff
@@ -642,6 +642,54 @@ def run(
...
                 os.remove(params_file.name)
 
+    def ssh(
+        self,
+        *command: str,
+        machine: int | None = None,
+        unit: str | None = None,
+        container: str | None = None,
+        host_key_checks: bool = True,
+        ssh_options: list[str] | None = None,
```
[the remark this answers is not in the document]
benhoyt: Good call. I've made this change, except used `Iterable[str]` instead of `Sequence[str]` to match the other similar uses.

james-garner-canonical (approved):
Looks good to me =)

jubilant/_juju.py
```diff
@@ -642,6 +642,43 @@ def run(
...
+            target: Where to run the command; this is a unit name such as ``mysql/0`` or a machine
+                ID such as ``0``.
+            command: Command to run, along with its arguments.
+            container: Name of container for Kubernetes charms. Defaults to the charm container.
+            host_key_checks: Set to False to disable host key checking (insecure).
+            ssh_options: OpenSSH client options, for example ``['-i', '/path/to/private.key']``.
+            user: User account to make connection with. Defaults to ``ubuntu`` account.
+        """
+        if not command:
+            raise TypeError('must provide a command')
```
james-garner-canonical, line 666: I guess we could do `command: str, *args: str` to have the type checker surface this and remove the need for the manual check? The only downside is that the name of the first argument is a misnomer if it's an environment variable assignment statement, but maybe not a big deal.

jubilant/_juju.py
```diff
@@ -642,6 +642,43 @@ def run(
...
+        args = ['ssh']
+        if container is not None:
+            args.extend(['--container', container])
+        if not host_key_checks:
+            args.append('--no-host-key-checks')
+        if user is not None:
+            args.append(f'{user}@{target}')
+        else:
+            args.append(str(target))
+        args.extend(ssh_options)
```
james-garner-canonical, line 677: I just realised that because `str` is also an `Iterable[str]`, this could silently have weird behaviour. Maybe we want:
```suggestion
        if isinstance(ssh_options, str):
            args.append(ssh_options)
        else:
            args.extend(ssh_options)
```

jubilant/_juju.py
```diff
@@ -642,6 +642,43 @@ def run(
...
+        user: str | None = None,
+    ) -> str:
+        """Executes a command using SSH on a machine or container and returns its standard output.
+
+        Args:
+            target: Where to run the command; this is a unit name such as ``mysql/0`` or a machine
+                ID such as ``0``.
+            command: Command to run, along with its arguments.
+            container: Name of container for Kubernetes charms. Defaults to the charm container.
+            host_key_checks: Set to False to disable host key checking (insecure).
```
james-garner-canonical, line 661:
```suggestion
            host_key_checks: Set to ``False`` to disable host key checking (insecure).
```

james-garner-canonical (approved):

jubilant/_juju.py
benhoyt: Ah, good point. However, I'd prefer to raise TypeError as this is `ssh_options` plural. We can always loosen that restriction later if we want.

jubilant/_juju.py
```diff
@@ -671,6 +671,10 @@ def ssh(
             ssh_options: OpenSSH client options, for example ``['-i', '/path/to/private.key']``.
             user: User account to make connection with. Defaults to ``ubuntu`` account.
         """
+        # Need this check because str is also an iterable of str.
+        if isinstance(ssh_options, str):
+            raise TypeError('ssh_options must be an iterable of str, not str')
+
```
james-garner-canonical, line 677: This is reasonable, but perhaps surprising given that the command and it's arguments can be specified as either separate arguments or all in one string.

jubilant/_juju.py
benhoyt: Ah yes, that's a good idea. I've done that, and similar for `juju.exec`. I considered it also for `juju.offer` with endpoints and more_endpoints like you suggested earlier, but that feels different to me -- command and args are different things (kind of), whereas splitting endpoints into two feels weird, so I'll leave `offer` alone.

jubilant/_juju.py
benhoyt: I'm going to skip this one, as I think it just adds visual noise. There's no ambiguity, and I haven't done this elsewhere in the docs.

jubilant/_juju.py
benhoyt: Yeah, but that's different (I think), because command and args are executed via a shell, whereas this is a list of options included in the command line.

jubilant/_juju.py
james-garner-canonical: Makes sense, I should have checked what had been done elsewhere.

PR #93, opened 2025-04-10
feat: add refresh command

### LINKED ISSUE #10: Add refresh command
Signature something like (somewhat similar to deploy):

```python
def refresh(
    self,
    app: str,
    *,
    channel: str | None = None,
    config: dict[str, Any] | None = None,  # TODO: is Any correct here?
    path: str | None = None,
    resources: dict[str, str] | None = None,
    revision: int | None = None,
    trust: bool = False,
):
```

benhoyt:
This PR adds support for `juju refresh`. The arguments are very similar to `Juju.deploy`.

Nothing too controversial here. The one thing that differs from the CLI is that I've stuffed all the "force_*" arguments under one, so `force=True` becomes `--force --force-base --force-units`. It seems to me that if you want to force and bypass checks, you want to force, and it saves a proliferation of arguments.

    def refresh(
        self,
        app: str,
        *,
        base: str | None = None,
        channel: str | None = None,
        config: Mapping[str, ConfigValue] | None = None,
        force: bool = False,
        path: str | None = None,
        resources: Mapping[str, str] | None = None,
        revision: int | None = None,
        storage: Mapping[str, str] | None = None,
        trust: bool = False,
    ):

    >>> juju.refresh('snappass-test')
    >>>

Fixes #10

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -488,6 +488,61 @@ class Juju:
 
         self.cli(*args)
 
+    def refresh(
+        self,
+        app: str,
+        *,
+        base: str | None = None,
+        channel: str | None = None,
+        config: Mapping[str, ConfigValue] | None = None,
+        force: bool = False,
+        path: str | None = None,
+        resources: Mapping[str, str] | None = None,
+        revision: int | None = None,
+        storage: Mapping[str, str] | None = None,
+        trust: bool = False,
+    ):
+        """Refresh (upgrade) an application's charm.
+
+        Args:
+            app: Name of application to refresh.
+            base: Select a different base than is currently running.
+            channel: Channel to use when deploying from Charmhub, for example, ``latest/edge``.
+            config: Application configuration as key-value pairs.
+            force: If true, bypass checks such as supported bases.
+            path: Refresh to a charm located at this path.
+            resources: Specify named resources to use for deployment, for example:
+                ``{'bin': '/path/to/some/binary'}``.
+            revision: Charmhub revision number to deploy.
+            storage: Constraints for named storage(s), for example, ``{'data': 'tmpfs,1G'}``.
+            trust: If true, allows charm to run hooks that require access to cloud credentials.
+        """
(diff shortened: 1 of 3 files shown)


IronCore864 (approved):
Approving with two minor comments:

1. Shall we add a "(Required)" for the app parameter in the docstring to make it more obvious? Glancing over the docstring, I felt overwhelmed until I noticed the asterisk after the app parameter.

2. I find the args handling slightly difficult to read, especially we have both `args.extend` and `args.append`, then `_format_config` which handles bool to str, and using `f'{k}={v}` directly. Would a helper func help? Something like `_add_arg`, so that we can simply:

```python
    _add_arg(args, '--channel', channel)
    _add_arg(args, '--config', config)
    _add_arg(args, '--force', force, is_flag=True)
```

benhoyt:
Thanks. Replying to your points:

> 1. Shall we add a "(Required)" for the app parameter in the docstring to make it more obvious? Glancing over the docstring, I felt overwhelmed until I noticed the asterisk after the app parameter.

I don't think so, as the `*` is Python syntax, but if we do, we should do this on all the methods. I've opened https://github.com/canonical/jubilant/issues/95 to track.

> 2. I find the args handling slightly difficult to read, especially we have both args.extend and args.append, then _format_config which handles bool to str, and using f'{k}={v} directly. Would a helper func help? Something like _add_arg, so that we can simply:
>
>     _add_arg(args, '--channel', channel)
>     _add_arg(args, '--config', config)
>     _add_arg(args, '--force', force, is_flag=True)

I don't think this is better, as a simple `if` statement and `list.append` or `list.extend` are very basic Python functionality, whereas with the helper a code reader would have to go look up how it worked. We could do it for "simple" string args. But the special cases like `config`  still need special handling. In your snippet above, for example, `_add_arg` would have to have special handling for `--config` using `_format_config()`. The `--force` arg is also incorrect above, as in the PR it actually adds all the `--force --force-base --force-units` flags. So I'm going to leave that as is.

PR #101, opened 2025-04-16
fix: Do not crash on Juju 2.9 status format

ghislainbourgeois:
This PR prevents Jubilant from crashing when fed a status from Juju 2.9. We have a suite of integration tests that need to work with that version of Juju as it is still supported, and this prevents using Jubilant for those tests.

I am aware that the rest of the API is not compatible with that version of Juju, but working directly with `juju.cli()` is feasible in our case. The only blocker is Jubilant crashing when parsing the `relations` in the status output.

### DIFF
--- a/jubilant/statustypes.py
+++ b/jubilant/statustypes.py
@@ -85,6 +85,8 @@ class AppStatusRelation:
 
     @classmethod
     def _from_dict(cls, d: dict[str, Any]) -> AppStatusRelation:
+        if not hasattr(d, 'get'):
+            return cls()
         return cls(
             related_app=d.get('related-application') or '',
             interface=d.get('interface') or '',
--- a/tests/unit/fake_statuses.py
+++ b/tests/unit/fake_statuses.py
@@ -218,3 +218,70 @@ DATABASE_WEBAPP_JSON = """
     }
 }
 """
+
+JUJU29_JSON = """
+{
+    "model": {
+        "name": "jubilant-0b6f94f3",
+        "type": "caas",
+        "controller": "microk8s-localhost",
+        "cloud": "microk8s",
+        "region": "localhost",
+        "version": "2.9.51",
+        "model-status": {
+            "current": "available",
+            "since": "16 Apr 2025 20:40:45Z"
+        },
+        "sla": "unsupported"
+    },
+    "machines": {},
+    "applications": {
+        "tls-certificates-requirer": {
+            "charm": "local:jammy/tls-certificates-requirer-0",
+            "series": "jammy",
+            "os": "ubuntu",
+            "charm-origin": "local",
+            "charm-name": "tls-certificates-requirer",
+            "charm-rev": 0,
+            "scale": 1,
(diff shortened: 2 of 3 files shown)


ghislainbourgeois:
> This is an interesting idea, and kudos for trying Jubilant with Juju 2.9! On reflection, however, I think it's a bit of a slippery slope -- it would be easy to sneak in more and more little fixes to make things work with 2.9 ("oh, this other field needs fixing as well" or "oh, what about `deploy`, all we have to do is tweak this argument name") and before you know it the code becomes a mess. So I'd rather not start.
> 
> What I suggest is that you patch `AppStatusRelation._from_dict` instead (it's a private method, but it's there to stay). I've got this working locally. There are various ways to do that, but this works for me locally:
> 
> ```python
> _old_from_dict = jubilant.statustypes.AppStatusRelation._from_dict
> 
> def _new_from_dict(v: dict[str, Any] | str):
>     if isinstance(v, str):
>         return jubilant.statustypes.AppStatusRelation(related_app=v)
>     return _old_from_dict(v)
> 
> jubilant.statustypes.AppStatusRelation._from_dict = _new_from_dict
> 
> status = jubilant.Status._from_dict(json.loads(JUJU29_JSON))
> print(status)
> ```
> 
> So I would just put something like that in your `conftest.py`, either at the top level or in a session-scoped, "autouse" fixture.
> 
> Note also how I pass the `relation_app` argument, which I think is correct here (instead of dropping the string).

I already am patching this in my WIP test case, and I hesitated before sending this PR out, because as you say there are a lot of other things that do not work even with that (`integrate()`, `run()`, most builtin functions to pass to `wait()`).

The reason why I sent this one is that this one makes Jubilant crash, and for the rest it works great if you limit yourself to `wait()` with carefully constructed lambdas and `cli()`.

I completely understand if you do not want to merge this and I am fine with patching for our use case (in TLS and Telco, I only have 1 test file that needs to work with Juju 2.9). I know that some other teams also need some kind of support for 2.9 for some charms, so that might be useful for them, but those tests would probably not be migrated to Jubilant.

In any case, thanks for this project, I really like the API and that's why I'm still fighting with Juju 2.9 to use it!

benhoyt:
> In any case, thanks for this project, I really like the API and that's why I'm still fighting with Juju 2.9 to use it!

Excellent.

I'll leave this PR open for a week or two in case anyone else has comments/ideas.

@ghislainbourgeois Can you please send a link to your test-migration PR, for our interest?

ghislainbourgeois:
> > In any case, thanks for this project, I really like the API and that's why I'm still fighting with Juju 2.9 to use it!
> 
> Excellent.
> 
> I'll leave this PR open for a week or two in case anyone else has comments/ideas.
> 
> @ghislainbourgeois Can you please send a link to your test-migration PR, for our interest?

Of course: https://github.com/canonical/tls-certificates-interface/pull/337

benhoyt:
Closing per comments above.

PR #102, opened 2025-04-17
chore: split integration tests into multiple files

### LINKED ISSUE #84: Split integration tests into multiple files with module-scoped fixture
`test_charm_basics` is becoming a bit silly. We should split the tests into multiple files and make `juju` a module-scoped fixture instead of a function-scoped one, with the first test doing the deploy and subsequent tests operating on that model.

benhoyt:
This splits the integration tests into multiple files, grouped by "category". I've changed the `juju` fixture to module-scoped so there's only one model created per test file. Test files with more than one test have a top-level `setup` fixture that deploys the charm under test (I kinda hate "autouse" but it doesn't seem too terrible used locally here).

For the most part the tests are order independent, but at least one isn't (`test_remove_application`). I think that's an okay trade-off.

Fixes #84.


james-garner-canonical (approved):
Looks great.

I think the `autouse` fixtures are a decent way to say 'always run this first' in a given module -- it's a shame `pytest` doesn't make it easier to define ordering or dependencies like that out of the box.

The order dependence of `test_remove_application` seems ok to me since we're running the tests serially anyway, and it's the last test in a small module.

PR #108, opened 2025-04-22
ci: test deploy with resources (and base)

### LINKED ISSUE #75: Test more features in integration tests
Ideally we should have tests of most of the options in integration test, for example `deploy()` with `resources=...` specified.

See also #71.

benhoyt:
Just adds an integration test for `juju deploy` with `--resource`.

Fixes #75. Most high-level features are covered by integration tests now. Not all arguments, but I think this is enough for now.


benhoyt:
Simple integration test -- merging without further review.

PR #109, opened 2025-04-22
feat: add scp command

### LINKED ISSUE #94: Design scp command
We should at least design a `Juju.scp` method signature, to ensure what we've come up with for `Juju.ssh` will work with a future `scp` method. I don't think adding `scp` is necessary right now, but we should at least plan ahead for a design that gels, before shipping 1.0.

benhoyt:
This PR adds `Juju.scp`, as follows:

```python
    def scp(
        self,
        source: str,
        destination: str,
        *,
        container: str | None = None,
        host_key_checks: bool = True,
        scp_options: Iterable[str] = (),
    ) -> None:
```

It's roughly consistent with the existing `ssh` command, though that separates `user` into a separate argument (whereas here `source` and `destination` may have different users, so I don't think it's worth doing that).

Fixes #94

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -703,6 +703,36 @@ class Juju:
             if params_file is not None:
                 os.remove(params_file.name)
 
+    def scp(
+        self,
+        source: str,
+        destination: str,
+        *,
+        container: str | None = None,
+        host_key_checks: bool = True,
+        scp_options: Iterable[str] = (),
+    ) -> None:
+        """Securely transfer files within a model.
+
+        Args:
+            source: Source of file, in format ``[[<user>@]<target>:]<path>``.
+            destination: Destination for file, in format ``[<user>@]<target>:[<path>]``.
+            container: Name of container for Kubernetes charms. Defaults to the charm container.
+            host_key_checks: Set to False to disable host key checking (insecure).
+            scp_options: ``scp`` client options, for example ``['-r', '-C']``.
+        """
+        args = ['scp']
+        if container is not None:
+            args.extend(['--container', container])
+        if not host_key_checks:
+            args.append('--no-host-key-checks')
+        args.append('--')
+        args.extend(scp_options)
+        args.append(source)
+        args.append(destination)
+
+        self.cli(*args)
+
     def ssh(
         self,
         target: str | int,
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -4,6 +4,7 @@ import functools
 import json
 import logging
 import os
+import pathlib
 import shlex
 import shutil
 import subprocess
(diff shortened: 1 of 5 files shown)


jubilant/_juju.py
```diff
@@ -703,6 +703,36 @@ def run(
...
+        *,
+        container: str | None = None,
+        host_key_checks: bool = True,
+        scp_options: Iterable[str] = (),
+    ) -> None:
+        """Securely transfer files within a model.
+
+        Args:
+            source: Source of file, in format ``[[<user>@]<target>:]<path>``.
+            destination: Destination for file, in format ``[<user>@]<target>:[<path>]``.
```
james-garner-canonical, line 719: Maybe a small typo here?
```suggestion
            destination: Destination for file, in format ``[<user>@]<target>[:<path>]``.
```

jubilant/_juju.py
```diff
@@ -703,6 +703,36 @@ def run(
             if params_file is not None:
                 os.remove(params_file.name)
 
+    def scp(
+        self,
+        source: str,
```
james-garner-canonical, line 708: Idea -- since `source` without any special characters `@`, `:` is just the path bit, it would be simple to support `pathlib.Path` here, which could be nice as some users (me) pretty much always use `pathlib.Path` for local file paths ... though currently users would just need to write `source=str(path)`

```suggestion
        source: str | pathlib.Path,
```

`args.append(source)` on line 731 would also need to become `args.append(str(source))`

benhoyt: Good call. And same for `destination`, right? (See the integration tests, where destination is a local path.)

benhoyt: Or maybe `os.PathLike[str]` like `deploy`'s *charm* argument?

james-garner-canonical: > same for destination, right?

Oh if `destination` has the same semantics then yes.

I thought from the help text that it didn't -- for `source` 'foo' would be treated as a local file named foo, while for `destination` 'foo' would be treated as the ssh target -- so `destination=Path('foo')` being treated as the ssh target would be surprising if we did `args.append(str(destination))` -- though maybe there's a way to tell Juju it's a local path in that case?

james-garner-canonical: > Or maybe os.PathLike[str] like deploy's charm argument?

That would work too, though I believe that the right way to get a string representation of a filesystem path from an [os.PathLike object](https://docs.python.org/3/library/os.html#os.PathLike) is via its `__fspath__` method, while `pathlib.Path` is both `os.PathLike` and guarantees that `str` is the string representation of the filesystem path

james-garner-canonical (approved):
This does seem to be the cleanest way to handle `source` and `destination` matching the cli behaviour.

Not bad with f-strings anyway, since you can `f'{user}@{target}:{path}'`.

jubilant/_juju.py
benhoyt: Good call, fixed.

jubilant/_juju.py
benhoyt: Cool, I've updated this (and other similar places) to `pathlib.Path`. I've also added it to `refresh()`'s *path* argument.

james-garner-canonical (approved):

PR #110, opened 2025-04-22
fix: log wait status diffs under jubilant.wait logger name

### LINKED ISSUE #104: Log verbosity options
I am currently migrating `kafka-bundle` tests to jubilant and everything is smooth, great job jubilant team!

There's one issue about the verbosity of jubilant logs, especially in CI. it would be much better If we had an option to control the log verbosity level. We could do `pytest --log-disable jubilant` but that would disable logging for jubilant altogether, which is also not desirable

benhoyt:
This changes the `Juju.wait` "status diff logs" to log under a separate `jubilant.wait` logger, rather than the top-level `jubilant` logger, so that it's easier for users to control them or turn them off separately.

Fixes #104.

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -17,6 +17,7 @@ from ._task import Task
 from .statustypes import Status
 
 logger = logging.getLogger('jubilant')
+logger_wait = logging.getLogger('jubilant.wait')
 
 
 class CLIError(subprocess.CalledProcessError):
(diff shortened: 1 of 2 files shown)


dimaqq (approved):
I'd personally call it `wait_logger` because that reads better, but in the end it's not a public name, so it's fine.

P.S.
I'd expect interested charmers to do smth like `logging.getLogger("jubilant.wait").setLevel("DEBUG" [or logging.WARNING)])`

PR #112, opened 2025-04-22
feat: add add_secret method

### LINKED ISSUE #4: Add add-secret command

benhoyt:
This PR adds support for the `juju add-secret` command to add a new named ("user") secret. It sets the secret content using a temporary file (like `run` with parameters) to avoid putting sensitive data on the command line.

Fixes #4

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -127,6 +127,36 @@ class Juju:
         self.cli(*args, include_model=False)
         self.model = model
 
+    def add_secret(
+        self,
+        name: str,
+        content: Mapping[str, str],
+        *,
+        info: str | None = None,
+    ) -> SecretURI:
+        """Add a new named secret and returns its secret URI.
+
+        Args:
+            name: Name for the secret.
+            content: Key-value pairs that represent the secret content, for example
+                ``{'password': 'hunter2'}``.
+            info: Optional description for the secret.
+        """
+        args = ['add-secret', name]
+        if info is not None:
+            args.extend(['--info', info])
+
+        with tempfile.NamedTemporaryFile('w+', delete=False, dir=self._temp_dir) as file:
+            _yaml.safe_dump(content, file)
+            args.extend(['--file', file.name])
+
+        try:
+            output = self.cli(*args)
+        finally:
+            os.remove(file.name)
+
+        return SecretURI(output.strip())
+
     def add_unit(
         self,
         app: str,
--- a/tests/integration/test_config.py
+++ b/tests/integration/test_config.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import json
+
 import pytest
 
 import jubilant
@@ -36,3 +38,18 @@ def test_trust(juju: jubilant.Juju):
     juju.trust('testdb', remove=True, scope='cluster')
     app_config = juju.config('testdb', app_config=True)
     assert app_config['trust'] is False
+
+
+def test_add_secret(juju: jubilant.Juju):
+    uri = juju.add_secret(
+        'sec1', {'username': 'usr', 'password': 'hunter2'}, info='A description.'
+    )
+
+    id = str(uri[len('secret:') :] if uri.startswith('secret:') else uri)
+    output = juju.cli('show-secret', 'sec1', '--reveal', '--format', 'json')
+    result = json.loads(output)
+    secret = result[id]
+
+    assert secret['name'] == 'sec1'
+    assert secret['description'] == 'A description.'
+    assert secret['content']['Data'] == {'username': 'usr', 'password': 'hunter2'}
--- a/tests/integration/test_config.py
+++ b/tests/integration/test_config.py
@@ -44,12 +44,11 @@ def test_add_secret(juju: jubilant.Juju):
     uri = juju.add_secret(
         'sec1', {'username': 'usr', 'password': 'hunter2'}, info='A description.'
     )
+    assert uri.startswith('secret:')
 
-    id = str(uri[len('secret:') :] if uri.startswith('secret:') else uri)
     output = juju.cli('show-secret', 'sec1', '--reveal', '--format', 'json')
     result = json.loads(output)
-    secret = result[id]
-
+    secret = result[uri[len('secret:'):]]
     assert secret['name'] == 'sec1'
     assert secret['description'] == 'A description.'
     assert secret['content']['Data'] == {'username': 'usr', 'password': 'hunter2'}
--- a/tests/unit/test_add_secret.py
+++ b/tests/unit/test_add_secret.py
@@ -1,0 +1,44 @@
+from __future__ import annotations
+
+import os.path
+import subprocess
+from typing import Any
+
+import pytest
+import yaml
+
+import jubilant
+
+
+def test_normal(monkeypatch: pytest.MonkeyPatch):
+    path = ''
+
+    def mock_run(args: list[str], **_: Any) -> subprocess.CompletedProcess[str]:
+        nonlocal path
+        assert args[:-1] == [
+            'juju',
+            'add-secret',
+            'sec1',
+            '--info',
+            'A description.',
+            '--file',
+        ]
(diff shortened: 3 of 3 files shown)


tests/integration/test_secrets.py
james-garner-canonical, line 1: Should `test_add_secret` be here?

tests/integration/test_config.py
```diff
@@ -36,3 +38,17 @@ def test_trust(juju: jubilant.Juju):
...
+
+def test_add_secret(juju: jubilant.Juju):
+    uri = juju.add_secret(
+        'sec1', {'username': 'usr', 'password': 'hunter2'}, info='A description.'
+    )
+    assert uri.startswith('secret:')
+
+    output = juju.cli('show-secret', 'sec1', '--reveal', '--format', 'json')
+    result = json.loads(output)
+    secret = result[uri[len('secret:') :]]
```
james-garner-canonical, line 51: Would it be worth dropping the `'secret:'` prefix? I was kind of surprised to learn that `SecretURI` was a `str` subclass -- but one idea if we want to keep the result as a string with the `'secret:'` prefix would be to add `SecretURI.uri` which drops the prefix.

tests/unit/test_add_secret.py
```diff
@@ -0,0 +1,44 @@
...
+        nonlocal path
+        assert args[:-1] == [
+            'juju',
+            'add-secret',
+            'sec1',
+            '--info',
+            'A description.',
+            '--file',
+        ]
+        (path,) = args[-1:]  # Ensure there's no extra args
```
james-garner-canonical, line 26:
```suggestion
        args_list, path = args
        assert args_list = [
            'juju',
            'add-secret',
            'sec1',
            '--info',
            'A description.',
            '--file',
        ]
```

tests/unit/test_add_secret.py
```diff
@@ -0,0 +1,44 @@
...
+    monkeypatch.setattr('subprocess.run', mock_run)
+    juju = jubilant.Juju()
+
+    secret_uri = juju.add_secret(
+        'sec1', {'username': 'usr', 'password': 'hunter2'}, info='A description.'
+    )
+
+    assert isinstance(secret_uri, jubilant.SecretURI)
+    assert secret_uri == 'secret:0123456789abcdefghji'  # noqa: S105
+    assert path
```
james-garner-canonical, line 43: Maybe clearer with an initial value of `path = None` and `assert path is not None`

jubilant/_juju.py
```diff
@@ -127,6 +127,36 @@ def add_model(
         self.cli(*args, include_model=False)
         self.model = model
 
+    def add_secret(
+        self,
+        name: str,
+        content: Mapping[str, str],
```
dimaqq, line 133: Wasn't there some way to create secrets with binary data?

benhoyt: Yeah, you can do that with the `#base64` syntax (I'm not sure if that works inside a YAML file or not). However, we don't support that in Ops yet, so I don't think any charms are using it. I was going to leave it out for now.

tests/unit/test_add_secret.py
benhoyt: Good idea, though I think you meant `*args_list, path = args`. I've made the same change to similar code in `test_run.py`.

tests/unit/test_add_secret.py
benhoyt: Good call, done (here and in `test_run.py`).

tests/integration/test_secrets.py
benhoyt: Oops, yes - done.

tests/integration/test_config.py
benhoyt: I want to keep the `secret:` prefix (whatever we get from Juju). However, I've added a `unique_identifier` property -- good idea. This is what we call this in Ops (`Secret.unique_identifier`) to distinguish it from "id" and "uri" (which is the whole thing with the `secret:` part).

jubilant/_juju.py
dimaqq: Perhaps the library users can supply binary secrets as encoded strings with `"#base64"` prefix even as it is now 🤔 

There's barely a couple of uses of python-libjuju `model.add_secret(...)` in the integration tests in our charm collection, and those are hardcoded strings.

And a single case of shelling out to set secrets, with `data: dict[str, str]`.

I guess you're right, it's very unlikely that someone needs binary secret data in integration tests.

jubilant/_juju.py
```diff
@@ -39,6 +39,20 @@ class WaitError(Exception):
...
+
+        This is the secret's globally-unique identifier (currently a 20-character Xid,
+        for example "9m4e2mr0ui3e8a215n4g").
+        """
+        if '/' in self:
+            return self.rsplit('/', maxsplit=1)[-1]
+        elif self.startswith('secret:'):
+            return self[len('secret:') :]
+        else:
+            return str(self)
```
dimaqq, line 54: This is enough to create secrets and pass them in config, but doesn't cover all possible secretey URIs, like cross-model secret references.

I'm pointing this out because it's a public class.

jubilant/_juju.py
```diff
@@ -127,6 +141,36 @@ def add_model(
...
+            args.extend(['--info', info])
+
+        with tempfile.NamedTemporaryFile('w+', delete=False, dir=self._temp_dir) as file:
+            _yaml.safe_dump(content, file)
+            args.extend(['--file', file.name])
+
+        try:
+            output = self.cli(*args)
+        finally:
+            os.remove(file.name)
```
dimaqq, line 170: Could be more concise:

```suggestion
        with tempfile.NamedTemporaryFile('w+', dir=self._temp_dir) as file:
            _yaml.safe_dump(content, file)
            args.extend(['--file', file.name])
            output = self.cli(*args)
```

dimaqq (approved):
FYI the tests that `add-secret` often do other things with secrets: https://github.com/canonical/object-storage-integrators/blob/f0a94cb9055d7a71293a52398bd7f74d9dfabe99/azure_storage/tests/integration/helpers.py#L128-L159

jubilant/_juju.py
benhoyt: I think this is enough. It's more or less copied from the Ops version [here](https://github.com/canonical/operator/blob/23f482a02f821d2a357ce43a56ba5e80c8d19f2b/ops/model.py#L1387). It handles cross-model secret references, as they're in format `secret://MODEL-UUID/UNIQUE-IDENTIFIER`, so that's handled by the `'/' in self` case.

In any case, I guess it's probably better to not make this public for now. This shouldn't really be used by tests (only by `add_secret` getting the thing out of the dict). I'll make it private.

benhoyt: Actually, on second thoughts, I do think it should be kept public, as it's needed for the integration test. This is kind of due to the fact we haven't implemented `show-secret` yet, but still, it's needed because of how Juju returns the data in a dict keyed by this unique_identifier.

jubilant/_juju.py
benhoyt: Good call, done. You need to `file.flush()` to ensure the content is written out before using it.

benhoyt:
I've also opened https://github.com/canonical/jubilant/issues/117 to track adding additional secret commands.

jubilant/_juju.py
james-garner-canonical: Worth adding comments to explain the cases? e.g. '/' case is to handle 'secret://MODEL-UUID/UNIQUE-IDENTIFIER'

james-garner-canonical (approved):

jubilant/_juju.py
benhoyt: Oops, I missed this comment before merging. Will update it as a drive-by in the PR I'm working on: https://github.com/canonical/jubilant/pull/116

Issue #90, opened 2025-04-07
Review API for consistency before 1.0 release


benhoyt:
To avoid inconsistencies in the API methods (for example, like https://github.com/canonical/jubilant/pull/89), review the API for consistency before the 1.0 release. Perhaps pair with another engineer on this to get two pairs of eyes on it.

Potential discussion points:

- [x] Is the signature of `juju.wait` as simple and consistent as it could be?
- [x] Behaviour of `juju.wait` and agent "idle" - see [Dima's comment here](https://github.com/canonical/jubilant/issues/91#issuecomment-2825845927). Are our defaults of delay=1s, timeout=60s, successes=3 sensible?
- [x] Is the behaviour of "reset via None" in `config` and `model_config` what we want?

benhoyt:
We went over the API as a team, and the two changes (minor but breaking) we've decided to make are:

- https://github.com/canonical/jubilant/pull/125
- https://github.com/canonical/jubilant/pull/126

We'll also add `all_units_idle`, #91.

PR #115, opened 2025-04-22
feat: add model-config command

### LINKED ISSUE #106: model-config wrapper?
is there a model-config wrapper? should there be one?

benhoyt:
This adds support for `juju model-config`: `Juju.model_config`, very similar to the existing `Juju.config` method. I've also added an example of an ops_test-like `fast_forward` context manager.

I don't want to add `fast_forward`, as I don't like it as a practice. I also don't like the name, and it seems too specific. I would consider adding a more general context manager like `jubilant.temp_model_config`, but I think it's better at least for now to leave it to the charms.

Fixes #106.

### DIFF
--- a/docs/how-to/migrate-from-pytest-operator.md
+++ b/docs/how-to/migrate-from-pytest-operator.md
@@ -325,7 +325,7 @@ juju.cli(
 
 ### A `fast_forward` context manager
 
-Pytest-operator has a `fast_forward` context manager which temporarily speeds up `update-status` hooks to fire every 10 seconds (instead of Juju's default of every 5 minutes). Jubilant doesn't provide this context manager, as we don't recommend it for new tests, but if you need it for migrating existing tests, you can define it as:
+Pytest-operator has a `fast_forward` context manager which temporarily speeds up `update-status` hooks to fire every 10 seconds (instead of Juju's default of every 5 minutes). Jubilant doesn't provide this context manager, as we don't recommend it for new tests. If you need it for migrating existing tests, you can define it as:
 
 ```python
 @contextlib.contextmanager
--- a/docs/how-to/migrate-from-pytest-operator.md
+++ b/docs/how-to/migrate-from-pytest-operator.md
@@ -323,6 +323,22 @@ juju.cli(
 )
 ```
 
+### A `fast_forward` context manager
+
+Pytest-operator has a `fast_forward` context manager which temporarily speeds up `update-status` hooks to fire every 10 seconds (instead of Juju's default of every 5 minutes). Jubilant doesn't provide this context manager, as we don't recommend it for new tests, but if you need it for migrating existing tests, you can define it as:
+
+```python
+@contextlib.contextmanager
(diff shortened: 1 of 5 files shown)


benhoyt:
@james-garner-canonical You're on a roll, but I think this will be all from me for a while. :-)

jubilant/_juju.py
```diff
@@ -470,6 +470,41 @@ def integrate(self, app1: str, app2: str, *, via: str | Iterable[str] | None = N
...
+        self, values: Mapping[str, ConfigValue | None] | None = None
+    ) -> Mapping[str, ConfigValue] | None:
+        """Get or set the configuration of the model.
+
+        If called with no arguments, get the model config and return it.
+        If called with the *values* argument, set the model config values and return ``None``.
+
+        Args:
+            values: Mapping of model config names to values to set, for example
+                ``{'update-status-hook-interval': '10s'}``. Reset values that are ``None``.
```
dwilding, line 489: It looks like the formatting went wrong here. "Reset values that are None" should belong with the `reset` arg, right?

docs/how-to/migrate-from-pytest-operator.md
```diff
@@ -323,6 +323,22 @@ juju.cli(
 )
 ```
 
+### A `fast_forward` context manager
+
+Pytest-operator has a `fast_forward` context manager which temporarily speeds up `update-status` hooks to fire every 10 seconds (instead of Juju's default of every 5 minutes). Jubilant doesn't provide this context manager, as we don't recommend it for new tests, but if you need it for migrating existing tests, you can define it as:
```
dwilding, line 328: Minor suggestion to avoid a long sentence:

```suggestion
Pytest-operator has a `fast_forward` context manager which temporarily speeds up `update-status` hooks to fire every 10 seconds (instead of Juju's default of every 5 minutes). Jubilant doesn't provide this context manager, as we don't recommend it for new tests. If you need it for migrating existing tests, you can define it as:
```

dwilding (approved):

jubilant/_juju.py
james-garner-canonical: I think this is correct, there is no `reset` arg, just a `reset` list constructed in the function body below.

It does differ from the `--reset` option in the cli, which I guess was discussed when implementing `config`.

I kind of wonder if an explicit `reset` argument would be good, but then I guess you do still have to handle `None` in values anyway, so I'm not sure.

james-garner-canonical (approved):
Looks good to me. I guess we can possibly talk about a `reset` argument in the api review next week.

jubilant/_juju.py
benhoyt: I've made some tweaks to the docstring here, similar to what David suggested. I've also added an item to https://github.com/canonical/jubilant/issues/90 about "reset" for our API review next weekl.

docs/how-to/migrate-from-pytest-operator.md
benhoyt: Done, thx

PR #120, opened 2025-04-23
fix!: return data rather than raising an exception on status-error

### LINKED ISSUE #118: Wrong usage of `is_xyz` status checks could lead to subtle issues
I have noticed that this check:

```
juju.wait(lambda status: status.apps["my-app"].is_active)
```

Specially, immediately after deploying an app, could lead to random (race-condition) errors. Adding some safeguard like this:

```
juju.wait(lambda status: "my-app" in status.apps and status.apps["my-app"].is_active)
```

Would resolve the issue. While it's pretty obvious in hindsight, I think maybe the docs should explicitly include some provisions about this so that charmers would avoid this pitfall.

benhoyt:
Juju can return the funky `{"status-error": ...}` object as a machine status when a machine is still being set up. This causes the current Jubilant code to raise `StatusError` and fail the test right away. This is not what we want -- we just want `juju.wait()` to wait till the machine has been set up and wait for the status they're looking for, without raising an exception.

We set the status to "failed", which is a [valid Juju status string](https://github.com/juju/juju/blob/5634dd4869129e1bf87e14afb586a28e97078729/core/status/status.go#L109), but unlikely to ever be tested in integration tests. In particular, we don't to use "error", which might need to be distinguished in integration tests, and if people are using `Juju.wait` with `error=jubilant.any_error`, will raise errors -- which we don't want for transient failures like this where Juju is (say) setting up a machine.

Here's an [example traceback](https://github.com/canonical/kafka-bundle/actions/runs/14587903350/job/40916796622#step:7:2367) that shows this:

```
...
   File "/home/runner/work/kafka-bundle/kafka-bundle/tests/integration/e2e/test_backup.py", line 173, in test_new_cluster_migration
    juju.wait(lambda status: status.apps["new-zk"].is_active, timeout=3600, delay=10)
  File "/home/runner/work/kafka-bundle/kafka-bundle/.tox/integration-e2e-backup/lib/python3.10/site-packages/jubilant/_juju.py", line 830, in wait
    status = Status._from_dict(result)
  File "/home/runner/work/kafka-bundle/kafka-bundle/.tox/integration-e2e-backup/lib/python3.10/site-packages/jubilant/statustypes.py", line 720, in _from_dict
    machines={k: MachineStatus._from_dict(v) for k, v in d['machines'].items()},
  File "/home/runner/work/kafka-bundle/kafka-bundle/.tox/integration-e2e-backup/lib/python3.10/site-packages/jubilant/statustypes.py", line 720, in <dictcomp>
    machines={k: MachineStatus._from_dict(v) for k, v in d['machines'].items()},
  File "/home/runner/work/kafka-bundle/kafka-bundle/.tox/integration-e2e-backup/lib/python3.10/site-packages/jubilant/statustypes.py", line 570, in _from_dict
    StatusInfo._from_dict(d['juju-status']) if 'juju-status' in d else StatusInfo()
  File "/home/runner/work/kafka-bundle/kafka-bundle/.tox/integration-e2e-backup/lib/python3.10/site-packages/jubilant/statustypes.py", line 69, in _from_dict
    raise StatusError(d['status-error'])
jubilant.statustypes.StatusError: cannot get status: machine not found
```

I believe the issue Iman pointed out regarding checking `"my-app" in status.apps` first is a red herring. This PR will solve the actual issue of the `StatusError` being raised sometimes, presumably before Juju has created the machine, and the `wait` will work as expected.

Fixes #118.

### DIFF
--- a/jubilant/statustypes.py
+++ b/jubilant/statustypes.py
@@ -25,7 +25,6 @@ __all__ = [
     'RemoteAppStatus',
     'RemoteEndpoint',
     'Status',
-    'StatusError',
     'StatusInfo',
     'StorageAttachments',
     'StorageInfo',
@@ -35,10 +34,6 @@ __all__ = [
     'VolumeAttachments',
     'VolumeInfo',
 ]
-
-
-class StatusError(Exception):
-    """Raised when ``juju status`` returns a status-error for certain types."""
 
 
 @dataclasses.dataclass(frozen=True)
@@ -66,7 +61,7 @@ class StatusInfo:
     @classmethod
     def _from_dict(cls, d: dict[str, Any]) -> StatusInfo:
         if 'status-error' in d:
-            raise StatusError(d['status-error'])
+            return cls(current='error', message=d['status-error'])
         return cls(
             current=d.get('current') or '',
             message=d.get('message') or '',
@@ -108,7 +103,10 @@ class UnitStatus:
     @classmethod
     def _from_dict(cls, d: dict[str, Any]) -> UnitStatus:
         if 'status-error' in d:
-            raise StatusError(d['status-error'])
+            return cls(
+                workload_status=StatusInfo(current='error', message=d['status-error']),
+                juju_status=StatusInfo(current='error', message=d['status-error']),
+            )
         return cls(
             workload_status=(
                 StatusInfo._from_dict(d['workload-status'])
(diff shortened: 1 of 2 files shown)


tonyandrewmeyer (approved):
Looks good - as per discussion earlier, agreed that a real Juju status but not the special error one is the best choice.

PR #124, opened 2025-04-28
feat: deploy a temp model into a specified controller

MichaelThamm:
To test a CMR in an integration test, it would be great to surface the controller option from the `add_model` method to the `temp_model` method so I can choose whether to deploy the model to LXD or microk8s.

I could manage this manually with juju.add_model, but then I lose the benefit of the automatic teardown.

### DIFF
--- a/jubilant/_pretty.py
+++ b/jubilant/_pretty.py
@@ -73,7 +73,7 @@ def dump(value: object, indent: str = '') -> str:
         return repr(value)
 
 
-def gron(value: object, prefix: str = '') -> Generator[str, None, None]:
+def gron(value: object, prefix: str = '') -> Generator[str]:
     """Yield gron-style lines of all fields within value, recursively.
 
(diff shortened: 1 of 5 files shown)


MichaelThamm:
I would appreciate your review @benhoyt

jubilant/_pretty.py
```diff
@@ -73,7 +73,7 @@ def dump(value: object, indent: str = '') -> str:
         return repr(value)
 
 
-def gron(value: object, prefix: str = '') -> Generator[str, None, None]:
+def gron(value: object, prefix: str = '') -> Generator[str]:
```
benhoyt, line 76: Drive-by: I've discovered the `, None, None` are optional here, which is nice, so I've simplified those.

benhoyt:
This look reasonable to you, @james-garner-canonical?

james-garner-canonical (approved):
Looks good to me!

PR #125, opened 2025-04-28
fix!: make "reset" an explicit arg for config/model_config

benhoyt:
This is per our API discussion, to make this a bit clearer/explicit, and more like the Juju CLI.

I needed a third `@overlord` to make the reset-only option work.

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -242,29 +242,44 @@ class Juju:
     def config(self, app: str, *, app_config: bool = False) -> Mapping[str, ConfigValue]: ...
 
     @overload
-    def config(self, app: str, values: Mapping[str, ConfigValue | None]) -> None: ...
+    def config(
+        self,
+        app: str,
+        values: Mapping[str, ConfigValue],
+        *,
+        reset: Iterable[str] = (),
+    ) -> None: ...
+
+    @overload
+    def config(self, app: str, *, reset: Iterable[str]) -> None: ...
 
     def config(
         self,
         app: str,
-        values: Mapping[str, ConfigValue | None] | None = None,
+        values: Mapping[str, ConfigValue] | None = None,
         *,
         app_config: bool = False,
+        reset: Iterable[str] = (),
     ) -> Mapping[str, ConfigValue] | None:
         """Get or set the configuration of a deployed application.
 
         If called with only the *app* argument, get the config and return it.
 
-        If called with the *values* argument, set the config values and return None.
-        For values in *values* that are None, reset them to their defaults.
+        If called with the *values* or *reset* arguments, set the config values and return None,
+        and reset any keys in *reset* to their defaults.
 
         Args:
             app: Application name to get or set config for.
-            values: Mapping of config names to values to set. Reset values that are None.
+            values: Mapping of config names to values to set.
             app_config: When getting config, set this to True to get the
                 (poorly-named) "application-config" values instead of charm config.
+            reset: List of keys to reset to their defaults.
         """
-        if values is None:
+        # Need this check because str is also an iterable of str.
+        if isinstance(reset, str):
+            raise TypeError('reset must be an iterable of str, not str')
+
+        if values is None and not reset:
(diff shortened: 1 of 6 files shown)


jubilant/_juju.py
```diff
@@ -242,29 +242,44 @@ def _cli(
...
+    def config(
+        self,
+        app: str,
+        values: Mapping[str, ConfigValue],
+        *,
+        reset: Iterable[str] = (),
+    ) -> None: ...
+
+    @overload
+    def config(self, app: str, *, reset: Iterable[str]) -> None: ...
```
james-garner-canonical, line 254: With the overloads getting more complicated, would you consider adding some type checking tests? That is, code that would live in the `tests` folder, be type checked in our static analysis (but not do anything at runtime). For example:
```py
def typecheck_juju_config(
    juju: jubilant.Juju,
    app: str,
    values: Mapping[str, jubilant.ConfigValue],
    app_config: bool,
    reset: Iterable[str]
):
    # get config
    config = juju.config(app)
    typing.assert_type(config, Mapping[str, ConfigValue])
    config = juju.config(app, app_config=app_config)
    typing.assert_type(config, Mapping[str, ConfigValue])
    # set config
    none = juju.config(app, values)
    typing.assert_type(none, None)
    none = juju.config(app, reset=reset)
    typing.assert_type(none, None)
    none = juju.config(app, values, reset=reset)
    typing.assert_type(none, None)
    # errors
    juju.config(app, values, app_config=app_config)  # type: ignore
```
We could drop the return type checking and the error cases to keep it simpler, but at least we want to make sure that all the intended ways to call it aren't errors.

james-garner-canonical (approved):
Looks good.

I wonder if we should handle `None` values explicitly since this is a breaking change.

Additionally, maybe a `str` for `reset` could be ok?

Finally, with the overloads getting more complicated, I wonder if it's worth adding some type checking tests to ensure that users don't get type checking errors for the intended ways of calling the methods (and perhaps also if they do get errors for the ways that would be errors at runtime).

jubilant/_juju.py
```diff
@@ -526,34 +537,40 @@ def integrate(self, app1: str, app2: str, *, via: str | Iterable[str] | None = N
...
 
-        reset: list[str] = []
         args = ['model-config']
-        for k, v in values.items():
-            if v is None:
-                reset.append(k)
-            else:
-                args.append(_format_config(k, v))
+        if values:
+            args.extend(_format_config(k, v) for k, v in values.items())
```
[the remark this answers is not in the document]
benhoyt: Yeah, that's not a bad idea. I agree that we shouldn't normally do this, but I've added an `if v is None` check in `_format_config` in this specific case.

jubilant/_juju.py
```diff
@@ -526,34 +537,40 @@ def integrate(self, app1: str, app2: str, *, via: str | Iterable[str] | None = N
...
         Args:
             values: Mapping of model config names to values to set, for example
-                ``{'update-status-hook-interval': '10s'}``. Reset values that are None.
+                ``{'update-status-hook-interval': '10s'}``.
+            reset: List of keys to reset to their defaults.
         """
-        if values is None:
+        # Need this check because str is also an iterable of str.
+        if isinstance(reset, str):
+            raise TypeError('reset must be an iterable of str, not str')
```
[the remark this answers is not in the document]
benhoyt: Done (except the other way around)

jubilant/_juju.py
benhoyt: I don't think that's necessary, as the unit tests are already Pyright-checked and have all the possible cases tested, so they'll already flag type checker errors if we've got it wrong. For the error cases, I think we should lean heavily on the static type checking.

james-garner-canonical: > For the error cases, I think we should lean heavily on the static type checking.

I agree with this. Type checking tests help with this. For example, the current implementation code and unit tests would not tell you if a case that was intended to raise a static type checking error actually did result in such an error. Not a big problem or anything though.

PR #126, opened 2025-04-28
fix!: in "offer", move endpoint to keyword-only arg instead of *args

benhoyt:
Per our API discussions. This is a bit more explicit and simpler, and provides better type checking by default.

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -559,25 +559,22 @@ class Juju:
 
         self.cli(*args)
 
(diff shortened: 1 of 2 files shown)


james-garner-canonical (approved):
Looks good, I think this is nicer. Note that this is a case where we explicitly support `str | Iterable[str]` to make the single argument case more ergonomic (with the side effect that strings like 'foo,bar' just work too) -- maybe worth following this approach for the `reset` arg of `config`?

benhoyt:
Yes, that's a fair point about `reset=` in `config` -- I'll make that change there.

PR #127, opened 2025-04-29
feat: add all_agents_idle helper

### LINKED ISSUE #91: Do we need an all_units_idle helper?
Do we need `jubilant.all_units_idle`? See [here](https://github.com/canonical/discourse-k8s-operator/pull/326/files#diff-c87c5173e88bb0f60b00107f4586304f4869a976a4f998425efe3481c0db9f56R228) in the discourse-k8s conftest.py. Or are there better ways of testing the same thing?

We should bikeshed on the name: `all_units_idle` is more correct, but `all_idle` is more discoverable. `all_agents_idle` or `all_idle_agents` might be worth considering too.

benhoyt:
I've called this `all_agents_idle` rather than `all_idle`, because it's not like the others -- it's looking at Juju unit agent status, not workload application and unit status.

The term "Agent" is also what you see in the default "juju status" output:

```
Unit              Workload  Agent  Address       Ports  Message
snappass-test/0*  active    idle   10.1.164.165         snappass started
```

Fixes #91.


jubilant/_all_any.py
```diff
@@ -50,6 +50,38 @@ def all_error(status: Status, *apps: str) -> bool:
...
+        # Use a lambda to wait for unit agents only from specified apps (blog, mysql).
+        juju.wait(lambda status: jubilant.all_idle_agents(status, 'blog', 'mysql'))
+
+    Args:
+        status: The status object being tested.
+        apps: An optional list of application names. If provided, only the unit agents of these
+            applications are tested, and each must be present in ``status.apps``.
+    """
+    if not apps:
+        apps = tuple(status.apps.keys())
```
tonyandrewmeyer, line 73: Is there a reason to not just have `apps = status.apps` here? Or `apps = tuple(status.apps)` if you want a copy for some reason?

jubilant/_all_any.py
```diff
@@ -50,6 +50,38 @@ def all_error(status: Status, *apps: str) -> bool:
...
+        # Use the callable directly to wait for unit agents from all apps to be idle.
+        juju.wait(jubilant.all_idle_agents)
+
+        # Use a lambda to wait for unit agents only from specified apps (blog, mysql).
+        juju.wait(lambda status: jubilant.all_idle_agents(status, 'blog', 'mysql'))
+
+    Args:
+        status: The status object being tested.
+        apps: An optional list of application names. If provided, only the unit agents of these
+            applications are tested, and each must be present in ``status.apps``.
```
tonyandrewmeyer, line 70: The wording here made me think that if they weren't present some sort of error would be raised, rather than the result being false. What about about switching it around to something like ", and if not present in status.apps will test as non-idle"?

tonyandrewmeyer (approved):
Re: the name - I'm definitely +1 on including "agents" since it's that way in `juju status` and to avoid confusion.

I somewhat feel like people would type "idle" for the autocomplete rather than "all_i" so "all_agents_idle" would be ok, and it would distinguish it more from the status ones (all-object-status vs. all-status rather than all-status-object vs. all-status). But I'm convinced enough by the "all_i" possibility that I think it's fine (basically, a +0 on the name).

james-garner-canonical (approved):
Looks good to me. +1 on Tony's comments. Name idea: `all_idle_unit_agents`.

jubilant/_all_any.py
benhoyt: There's no reason to use `.keys()` -- thanks, I've removed that. The `tuple(...)` is to satisfy the type checker. But I've actually refactored this to use a helper function like the other `all_*` methods, and it's better and more consistent that way.

jubilant/_all_any.py
benhoyt: Reworded. Not sure it's perfect, so feel free to comment further (will save for future PRs though).

benhoyt:
Ha, I hadn't actually tested autocomplete, but `all_i` finds `all_agents_idle` in PyCharm, and I presume it would in VS Code too. In that case I much prefer `all_agents_idle` -- updated!

jubilant/_all_any.py
```diff
@@ -150,9 +150,31 @@ def any_waiting(status: Status, *apps: str) -> bool:
...
+        # Use the callable directly to wait for unit agents from all apps to be idle.
+        juju.wait(jubilant.all_agents_idle)
+
+        # Use a lambda to wait for unit agents only from specified apps (blog, mysql).
+        juju.wait(lambda status: jubilant.all_agents_idle(status, 'blog', 'mysql'))
+
+    Args:
+        status: The status object being tested.
+        apps: An optional list of application names. If provided, only the unit agents of these
+            (and their units) are tested, and if not present in ``status.apps``, returns False.
```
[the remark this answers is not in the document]
benhoyt: I feel this is saying too much (the last sentence is implied). But I've taken some of the suggestion, and actually removed the first sentence of the arg docstring from all of them.

PR #132, opened 2025-04-30
fix: fix format of --bind argument for deploy

paulomach:
Fix the format of the `--bind` argument to deploy from comma-separated to space-separated key=value pairs. See [Juju source](https://github.com/juju/juju/blob/214723de24e82f5d14fd255de0885092e0a4dd7c/cmd/juju/application/binding.go#L17-L24).

### DIFF
--- a/jubilant/__init__.py
+++ b/jubilant/__init__.py
@@ -43,4 +43,4 @@ __all__ = [
     'temp_model',
 ]
 
-__version__ = '1.0.0'
+__version__ = '1.0.1'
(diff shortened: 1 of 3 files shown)


benhoyt:
Ah, thanks! I had misread the Juju code earlier, but it looks like the format for multiple bindings is actually a space-separated list of key=value pairs (Juju doesn't complain about multiple `--bind` options, but only uses the last one, which is a bit disconcerting). See the [Juju source](https://github.com/juju/juju/blob/214723de24e82f5d14fd255de0885092e0a4dd7c/cmd/juju/application/binding.go#L17-L24).

I've made this update and will release a new version soon.

Issue #117, opened 2025-04-23
Consider adding additional secret-related commands


benhoyt:
For example, the [object-storage-integrators integration tests](https://github.com/canonical/object-storage-integrators/blob/f0a94cb9055d7a71293a52398bd7f74d9dfabe99/azure_storage/tests/integration/helpers.py#L128-L159) use `show-secret`, `grant-secret`, and `update-secret`, in addition to `add-secret` (which is being added in #112).

tonyandrewmeyer:
Done in #135.

PR #135, opened 2025-04-30
feat: secret management

kayra1:
This PR adds the ability to read as well as add secrets to the model. It mirrors the API of the `config` function.

### DIFF
--- a/jubilant/__init__.py
+++ b/jubilant/__init__.py
@@ -43,4 +43,4 @@ __all__ = [
     'temp_model',
 ]
 
-__version__ = '1.0.0'
+__version__ = '1.0.1'
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import datetime
 import functools
 import json
 import logging
@@ -11,6 +12,7 @@ import subprocess
 import tempfile
 import time
 from collections.abc import Callable, Iterable, Mapping
+from dataclasses import dataclass
 from typing import Any, Literal, Union, overload
 
 from . import _pretty, _yaml
@@ -55,6 +57,31 @@ class SecretURI(str):
             return self[len('secret:') :]
         else:
             return str(self)
+
+
+@dataclass
+class RedactedSecret:
+    uri: SecretURI
+    name: str | None
+    owner: str
+    description: str
+    rotation: str | Literal['never']
+    revision: int
+    created: datetime.datetime
+    updated: datetime.datetime
+
+
+@dataclass
+class RevealedSecret:
+    uri: SecretURI
+    revision: int
+    checksum: str
+    owner: str
+    description: str
+    name: str | None
+    created: datetime.datetime
+    updated: datetime.datetime
+    content: dict[str, str]
 
 
 ConfigValue = Union[bool, int, float, str, SecretURI]
@@ -143,33 +170,6 @@ class Juju:
 
         self.cli(*args, include_model=False)
         self.model = model
-
-    def add_secret(
-        self,
-        name: str,
-        content: Mapping[str, str],
-        *,
-        info: str | None = None,
-    ) -> SecretURI:
-        """Add a new named secret and returns its secret URI.
-
-        Args:
-            name: Name for the secret.
-            content: Key-value pairs that represent the secret content, for example
-                ``{'password': 'hunter2'}``.
-            info: Optional description for the secret.
-        """
-        args = ['add-secret', name]
-        if info is not None:
-            args.extend(['--info', info])
-
-        with tempfile.NamedTemporaryFile('w+', dir=self._temp_dir) as file:
-            _yaml.safe_dump(content, file)
-            file.flush()
-            args.extend(['--file', file.name])
-            output = self.cli(*args)
-
-        return SecretURI(output.strip())
 
     def add_unit(
         self,
@@ -836,6 +836,124 @@ class Juju:
         args.append(str(destination))
 
         self.cli(*args)
+
+    @overload
+    def secret(
+        self, *, label: Literal[None] = None, uri: Literal[None] = None
+    ) -> Iterable[RedactedSecret]: ...
+
+    @overload
+    def secret(
+        self, *, label: str | None = None, revision: int | None = None
+    ) -> RevealedSecret: ...
+
+    @overload
+    def secret(
+        self, *, uri: str | SecretURI | None = None, revision: int | None = None
+    ) -> RevealedSecret: ...
+
+    @overload
+    def secret(
+        self,
+        label: str | None,
+        values: Mapping[str, str],
+        *,
+        update: bool = False,
+        revision: int | None = None,
+        info: str | None = None,
+    ) -> SecretURI: ...
+
+    def secret(
+        self,
+        label: str | None = None,
+        values: Mapping[str, str] = {},
+        info: str | None = None,
+        update: bool = False,
+        revision: int | None = None,
+        uri: str | SecretURI | None = None,
+    ) -> Iterable[RedactedSecret] | RevealedSecret | SecretURI:
+        """Get or set any secret in the model.
+
+        If called with only the label argument, get the revealed secret and return it.
+
+        If called with only the label or uri argument, get the revealed secret and return it.
+
+        If called with the *values* and other optional arguments, create the secret.
+        If update is set to true, attempt to update an existing secret instead.
+
+        Examples:
+            juju.secret()
+            juju.secret(label="mysecret")
+            juju.secret(label="secret", {"password": "hunter2"})
+            juju.secret(label="secret", {"password": "hunter2"}, update=True)
+
+        Args:
+            label: Secret label to get or set. This is prioritized over the uri if both are set.
+            uri: Secret URI to get or set.
+            values: Mapping of secret keys to values to set.
+            info: Description of the secret to set.
+            update: When adding a secret, set this to True to update the label/uri instead.
+            revision: Revision number of the secret to get.
+        """
+        if label is None and uri is None:
+            stdout = self.cli('secrets', '--format', 'json')
+            output = json.loads(stdout)
+            return [
+                RedactedSecret(
+                    revision=int(obj['revision']),
+                    uri=SecretURI('secret:' + uri_from_juju),
+                    name=str(obj['label']) if 'label' in obj else None,
+                    owner=str(obj['owner']) if 'owner' in obj else '<model>',
+                    description=str(obj['description']) if 'description' in obj else '',
+                    rotation=str(obj['rotation']) if 'rotation' in obj else 'never',
+                    created=datetime.datetime.fromisoformat(str(obj['created']))
+                    if 'created' in obj
+                    else datetime.datetime.now(),
+                    updated=datetime.datetime.fromisoformat(str(obj['updated']))
+                    if 'updated' in obj
+                    else datetime.datetime.now(),
+                )
+                for uri_from_juju, obj in output.items()
+            ]
+
+        if not values:
+            if label is not None and uri is not None:
+                raise TypeError('cannot specify both label and uri')
+            args = ['get-secret', '--format', 'json']
+            if label is not None:
+                args.extend([label])
+            elif uri is not None:
+                args.extend([str(uri)])
+            if revision is not None:
+                args.extend(['--revision', str(revision)])
+            stdout = self.cli(*args)
+            output = json.loads(stdout)
+            uri_from_juju, obj = next(iter(output.items()))
+            return RevealedSecret(
+                uri=SecretURI('secret:' + uri_from_juju),
+                revision=int(obj['revision']) if 'revision' in obj else 0,
+                checksum=str(obj['checksum']) if 'checksum' in obj else '',
+                owner=str(obj['owner']) if 'owner' in obj else '<model>',
+                description=str(obj['description']) if 'description' in obj else '',
+                name=str(obj['label']) if 'label' in obj else None,
+                created=datetime.datetime.fromisoformat(str(obj['created'])),
+                updated=datetime.datetime.fromisoformat(str(obj['updated'])),
+                content=obj['content'],
+            )
+
+        args = ['add-secret']
+        if update:
+            args = ['update-secret']
+        id = label if label is not None else uri
+        args.extend([str(id)])
+        if info is not None:
+            args.extend(['--info', info])
+        with tempfile.NamedTemporaryFile('w+', dir=self._temp_dir) as file:
+            _yaml.safe_dump(values, file)
+            file.flush()
+            args.extend(['--file', file.name])
+            stdout = self.cli(*args)
+        return SecretURI(stdout.strip())
 
     def ssh(
         self,
--- a/tests/integration/test_secrets.py
+++ b/tests/integration/test_secrets.py
@@ -1,12 +1,11 @@
+import datetime
 import json
 
 import jubilant
 
 
 def test_add_secret(juju: jubilant.Juju):
-    uri = juju.add_secret(
-        'sec1', {'username': 'usr', 'password': 'hunter2'}, info='A description.'
-    )
+    uri = juju.secret('sec1', {'username': 'usr', 'password': 'hunter2'}, info='A description.')
 
     output = juju.cli('show-secret', 'sec1', '--reveal', '--format', 'json')
     result = json.loads(output)
@@ -14,3 +13,61 @@ def test_add_secret(juju: jubilant.Juju):
     assert secret['name'] == 'sec1'
     assert secret['description'] == 'A description.'
     assert secret['content']['Data'] == {'username': 'usr', 'password': 'hunter2'}
+
+
+def test_update_secret(juju: jubilant.Juju):
+    uri = juju.secret('sec1', {'username': 'usr', 'password': 'hunter2'}, info='A description.')
+    uri2 = juju.secret(
+        'sec1', {'username': 'usr', 'password': 'hunter3'}, info='A new description.'
+    )
+
+    assert uri == uri2
+    output = juju.cli('show-secret', 'sec1', '--reveal', '--format', 'json')
+    result = json.loads(output)
+    secret = result[uri.unique_identifier]
+    assert secret['name'] == 'sec1'
+    assert secret['description'] == 'A new description.'
+    assert secret['content']['Data'] == {'username': 'usr', 'password': 'hunter3'}
+
+
+def test_get_all_secrets(juju: jubilant.Juju):
+    uris: list[str] = []
+    args = [
+        {'label': 'sec1', 'username': 'usr', 'password': 'hunter2', 'info': 'A description'},
+        {'label': 'sec2', 'username': 'usr2', 'password': 'hunter3', 'info': 'A new description'},
+    ]
+    for arg in args:
+        uri = juju.secret(
+            arg['label'],
+            {'username': arg['username'], 'password': arg['password']},
+            info=arg['info'],
+        )
+        uris.append(uri)
+    secrets = juju.secret()
+
+    assert isinstance(secrets, list)
+    assert len(secrets) > 0
+    for secret in secrets:
+        assert secret.uri in uris
+        assert secret.revision == 1
+        assert secret.name == '<model>'
+        assert secret.description in [arg['info'] for arg in args]
+        assert secret.created.year == datetime.datetime.now().year
+        assert secret.created == secret.updated
+
+
+def test_get_secret(juju: jubilant.Juju):
+    uri = juju.secret('sec1', {'username': 'usr', 'password': 'hunter2'}, info='A description.')
+
+    secret = juju.secret(label='sec1')
+    secret2 = juju.secret(uri=uri)
+
+    assert secret == secret2
+    assert secret.uri == uri
+    assert secret.revision == 1
+    assert secret.owner == '<model>'
+    assert secret.name == 'sec1'
+    assert secret.description == 'A description'
+    assert secret.created.year == datetime.datetime.now().year
+    assert secret.created == secret.updated
+    assert secret.content == {'username': 'usr', 'password': 'hunter2'}
--- a/tests/unit/test_secret.py
+++ b/tests/unit/test_secret.py
@@ -34,7 +34,7 @@ def test_normal(monkeypatch: pytest.MonkeyPatch):
     monkeypatch.setattr('subprocess.run', mock_run)
     juju = jubilant.Juju()
 
-    secret_uri = juju.add_secret(
+    secret_uri = juju.secret(
         'sec1', {'username': 'usr', 'password': 'hunter2'}, info='A description.'
     )
 


benhoyt (commented):
Thanks for the PR. We won't be adding this in its current state, because it's a breaking change, and we've just release 1.0 with a guarantee of a stable API -- this breaks `add_secret`.

We are interested in this functionality if it's useful in integration tests (see https://github.com/canonical/jubilant/issues/117).

What's your own rationale: do you use this in your integration tests, and if so, can you please point me to where?

In any case, we'd want the API to mirror the Juju API, which has separate commands for secret handling (`add-secret`, `update-secret`, `secrets`, and so on). The `config` command is overloaded in the way it is because that's how the Juju CLI works too.

So I'm not opposed to an addition *like* this, but let's discuss the motivation/necessity of it first, as well as sketch out the proposed API.

kayra1:
hello, yes it's pretty common for us to mess with secrets in the TLS charms. [This](https://github.com/canonical/notary-k8s-operator/pull/89) is the second charm in the team that I'm moving to jubilant.

I see the motivation behind mirroring the cli. I can update the PR to mirror the secrets functions too. Would it be possible to merge it in that case? 

Besides that, what do you think about the dataclass return types? Is it good or is there a better recommendation?

benhoyt:
> I see the motivation behind mirroring the cli. I can update the PR to mirror the secrets functions too. Would it be possible to merge it in that case?

I'm sorry to put on the brakes, but I'd like to move a bit more slowly and carefully here, especially now that Jubilant is 1.0. I am willing to add these commands, but let's make sure we agree on all aspects of the API first.

I believe the following mirrors the API more closely. I also think it's better if there's only one Secret type. What do you think of this:
 
```python
def secrets(
    self,
    *,
    owner: str | None = None,
) -> list[Secret]: ...

# would raise ValueError if secret not found (similar to exec/run)
def show_secret(
    self,
    name_or_uri: str,
    *,
    reveal: bool = False,
    revision: int | None = None,
    revisions: bool = False,
) -> Secret:

def update_secret(
    self,
    name_or_uri: str,
    content: Mapping[str, str],
    *,
    info: str | None = None,
) -> None:
```

Then in a new file `secrettypes.py` (similar to `statustypes.py`), we'd have this (but `Secret` would be exported in `__init__.py` too):

```python
@dataclasses.dataclass
class Secret:
	uri: SecretURI
	revision: int
	checksum: str
	expires: str | None
	rotation: str
	rotates: str | None
	owner: str | None
	description: str | None
	name: str | None
	label: str | None
	created: str
	updated: str
	error: str  # I don't know what this represents
	content: dict[str, str]
	revisions: list[SecretRevision]
	access: list[SecretAccess]

@dataclasses.dataclass
class SecretAccess:
	target: str
	scope: str
	role: str

@dataclass.dataclass
class SecretRevision:
	revision: int
	backend: str
	created: str
	updated: str
	expires: str | None
```

In addition, we should have new unit test files for each of these, `tests/unit/test_add_secret.py` and so on.

> Besides that, what do you think about the dataclass return types? Is it good or is there a better recommendation?

Yeah, I like dataclasses -- though I think we should only have one, with "content" only populated if you do `show_secret(reveal=True)`. But per above, let's add a new `jubilant.secrettypes` file with these in them, and import `Status` and `StatusURI` only to the top level (the dataclasses used in signatures directly).

jubilant/_juju.py
```diff
@@ -905,6 +1025,35 @@ def trust(
...
         self.cli(*args)
 
+    def update_secret(
+        self,
+        name_or_uri: str,
+        content: Mapping[str, str],
+        *,
+        info: str | None = None,
+        name: str | None = None,
+        auto_prune: bool = False,
```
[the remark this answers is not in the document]
kayra1: fixed.

for what it's worth enabling strict mode on pylance in this repo in VSCode returns a ton of errors for me, so I've disabled automatic static checks and just use the CLI like the integration tests do.

jubilant/_juju.py
```diff
@@ -689,6 +671,19 @@ def remove_relation(self, app1: str, app2: str, *, force: bool = False) -> None:
             args.append('--force')
         self.cli(*args)
 
+    def remove_secret(self, name_or_uri: SecretURI, *, revision: int | None = None) -> None:
```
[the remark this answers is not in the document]
kayra1: done

jubilant/_juju.py
```diff
@@ -689,6 +671,19 @@ def remove_relation(self, app1: str, app2: str, *, force: bool = False) -> None:
...
+        """Remove a secret from the model.
+
+        Args:
+            name_or_uri: The URI of the secret to remove.
+            revision: The revision of the secret to remove.
+        """
+        args = ['remove-secret', str(name_or_uri)]
```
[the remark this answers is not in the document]
kayra1: done

jubilant/_juju.py
```diff
@@ -689,6 +671,19 @@ def remove_relation(self, app1: str, app2: str, *, force: bool = False) -> None:
...
+        if revision is not None:
+            args.append('--revision')
+            args.append(str(revision))
```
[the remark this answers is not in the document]
kayra1: done

jubilant/_juju.py
```diff
@@ -837,6 +832,131 @@ def scp(
 
         self.cli(*args)
 
+    def secrets(self) -> Iterable[Secret[Hidden]]:
```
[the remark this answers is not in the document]
kayra1: done

jubilant/_juju.py
```diff
@@ -837,6 +832,131 @@ def scp(
...
+        """Get all secrets in the model.
+
+        Returns:
+            A list of :class:`RedactedSecret` objects, one for each secret in the model.
```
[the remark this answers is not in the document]
kayra1: done

jubilant/_juju.py
```diff
@@ -837,6 +832,131 @@ def scp(
...
+        output = json.loads(stdout)
+        return [
+            Secret(
+                uri=SecretURI('secret:' + uri_from_juju),
+                name=str(obj['name']) if 'name' in obj else None,
+                owner=str(obj['owner']) if 'owner' in obj else '<model>',
+                rotation=str(obj['rotation']) if 'rotation' in obj else 'never',
+                revision=int(obj['revision']),
+                description=str(obj['description']) if 'description' in obj else '',
+                created=datetime.datetime.fromisoformat(str(obj['created'].replace('Z', '+00:00')))
```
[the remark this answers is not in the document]
kayra1: I personally think it's fine as it is.

* The format is quite standard, so I doubt juju will ever change it
* It's an easy change to make in the future
* This package is for integration tests, so nothing in production would run into issues if this were to fail.

jubilant/_juju.py
```diff
@@ -837,6 +832,131 @@ def scp(
...
+                if 'updated' in obj
+                else datetime.datetime.now(),
+                checksum='',
+                expires=None,
+                rotates=None,
+                label=None,
+                error='',
+                content=None,
+                access=[],
+                revisions=[],
```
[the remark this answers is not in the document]
kayra1: These fields came from Ben's suggested class, but I personally couldn't find any way to get them from the CLI directly. I wasn't sure to keep it in code for future developers since it may be surfaced by juju later, or delete it since they are currently not relevant.

jubilant/_juju.py
```diff
@@ -837,6 +832,131 @@ def scp(
...
+        """Get all secrets in the model.
+
+        Returns:
+            A list of :class:`RedactedSecret` objects, one for each secret in the model.
+        """
+        stdout = self.cli('secrets', '--format', 'json')
+        output = json.loads(stdout)
+        return [
+            Secret(
+                uri=SecretURI('secret:' + uri_from_juju),
```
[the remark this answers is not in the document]
james-garner-canonical: Ah, that makes sense, I understand why you're prepending `'secret:'` -- and it looks like Juju consistently outputs the uri without any prefix with these methods. I had wondered if you were prepending `secret:` to ensure roundtrip safety, but as that's not a concern, I think what you have here is fine, feel free to resolve this thread.

jubilant/_juju.py
```diff
@@ -837,6 +832,131 @@ def scp(
...
+                name=str(obj['name']) if 'name' in obj else None,
+                owner=str(obj['owner']) if 'owner' in obj else '<model>',
+                rotation=str(obj['rotation']) if 'rotation' in obj else 'never',
```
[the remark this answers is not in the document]
kayra1: the str() calls were breaking for some reason, I've switched them out for get. I'll try to keep them this way even if we don't go with typeddicts.

jubilant/_juju.py
```diff
@@ -837,6 +832,131 @@ def scp(
...
+    def secrets(self) -> Iterable[Secret[Hidden]]:
+        """Get all secrets in the model.
+
+        Returns:
+            A list of :class:`RedactedSecret` objects, one for each secret in the model.
+        """
+        stdout = self.cli('secrets', '--format', 'json')
+        output = json.loads(stdout)
+        return [
+            Secret(
```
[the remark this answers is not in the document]
kayra1: introduced a typeddict.

jubilant/_juju.py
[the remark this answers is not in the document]
kayra1: If I recall correctly, those can be set for secrets created by a charm. I will validate and let you know.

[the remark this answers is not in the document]
tonyandrewmeyer: Yes, the CLI lets you inspect charm secrets, and charms can [set an expiry and rotation policy](https://ops.readthedocs.io/en/latest/reference/ops.html#ops.Application.add_secret).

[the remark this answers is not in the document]
james-garner-canonical: In that case, let's make sure we include `expires`, `rotates`, `rotation`, correctly populating the fields from the data returned by Juju, and ideally adding an integration test that covers them.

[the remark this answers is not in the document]
james-garner-canonical: I'd love to know how `error` and `access` can be set and what the data coming back from Juju looks like, and ideally doing the same.

jubilant/secrettypes.py
```diff
@@ -0,0 +1,106 @@
...
+    owner: str
+    description: str
+    name: str
+    label: str
+    created: str
+    updated: str
+    error: str
+    content: NotRequired[dict[str, dict[str, str]]]
+    revisions: list[SecretRevisionResponse]
+    access: list[SecretAccess]
```
james-garner-canonical, line 97: If we can figure out how to get `access` populated, presumably we'll want a `SecretAccessResponse` here instead.

jubilant/_juju.py
```diff
@@ -837,6 +831,135 @@ def scp(
...
+                updated=datetime.datetime.fromisoformat(
+                    str(obj.get('updated', '').replace('Z', '+00:00'))
+                ),
+                content=None,
+                checksum='',
+                expires=None,
+                rotates=None,
+                label=None,
+                error='',
+                access=[],
```
tonyandrewmeyer, line 862: I think this should probably be `None`. If we set it to `[]` when `secrets()` is called (as far as I know, there's no way to have the Juju CLI include access in `secrets` output) then that looks the same as if it's `[]` when `show_secret()` is called, but it's a potentially different meaning.

Basically: we need to distinguish between "don't know" and "no access granted". `None` seems ok for the former, although I can also see people doing `if access:` and the false-y nature of `None` causing issues. So maybe there should be some sort of sentinel meaning "don't know" instead?

One thought: maybe we put it in `_access` as `None`, and make `access` a property that calls `show_secret()` to get the value if it's `None`? Maybe that's too magic, and it would mean a subprocess call in a property, which I don't love as a pattern, but it would mean that it was always the correct value.

jubilant/_juju.py
```diff
@@ -837,6 +831,135 @@ def scp(
...
+                revisions=[],
```
tonyandrewmeyer, line 863: This has the same issue as `access`, although less concerning since `[]` is clearly an invalid value. However, I think `None` would make more sense here, or perhaps just copying the same solution as for `access`.

jubilant/_juju.py
```diff
@@ -837,6 +831,135 @@ def scp(
...
+            )
+            for uri_from_juju, obj in output.items()
+        ]
+
+    @overload
```
tonyandrewmeyer, line 868: The overloads are nice, thanks!

Could you add another one that has `reveal` and `revisions` both as True, that returns `NoReturn`? That would make it clear to type checkers that it's invalid.

jubilant/_juju.py
```diff
@@ -837,6 +831,135 @@ def scp(
...
+        if revisions:
+            args.append('--revisions')
+        if revision is not None:
+            args.extend(['--revision', str(revision)])
+        stdout = self.cli(*args)
+        output: dict[str, SecretResponse] = json.loads(stdout)
+        uri_from_juju, obj = next(iter(output.items()))
+        return Secret[Any](
+            uri=SecretURI('secret:' + uri_from_juju),
+            revision=obj.get('revision', 0),
```
tonyandrewmeyer, line 928: Same as above, I'm sure the output can never be missing this.

```suggestion
            revision=obj['revision'],
```

jubilant/_juju.py
```diff
@@ -837,6 +831,135 @@ def scp(
...
+            checksum=obj.get('checksum', ''),
```
tonyandrewmeyer, line 929: Same as above, I'm sure you cannot get this.

jubilant/_juju.py
```diff
@@ -837,6 +831,135 @@ def scp(
...
+            owner=obj.get('owner', '<model>'),
```
tonyandrewmeyer, line 930: Same as above, I'm sure this is always provided.
```suggestion
            owner=obj['owner'],
```

jubilant/_juju.py
```diff
@@ -837,6 +831,135 @@ def scp(
...
+            description=obj.get('description', ''),
+            name=obj.get('name', None),
```
tonyandrewmeyer, line 932:
```suggestion
            name=obj.get('name'),
```

jubilant/_juju.py
```diff
@@ -837,6 +831,135 @@ def scp(
...
+            created=datetime.datetime.fromisoformat(
+                str(obj.get('created', '').replace('Z', '+00:00'))
+            ),
+            updated=datetime.datetime.fromisoformat(
+                str(obj.get('updated', '').replace('Z', '+00:00'))
+            ),
```
tonyandrewmeyer, line 938: Same comment as above: no need for `str`, `""` is invalid, these are always present anyway.

jubilant/_juju.py
```diff
@@ -837,6 +831,135 @@ def scp(
...
+            revisions=[
+                SecretRevision(
+                    revision=revision.get('revision', ''),
+                    backend=revision.get('backend', ''),
+                    created=datetime.datetime.fromisoformat(
+                        str(revision.get('created', '').replace('Z', '+00:00'))
+                    ),
+                    updated=datetime.datetime.fromisoformat(
+                        str(revision.get('updated', '').replace('Z', '+00:00'))
+                    ),
```
tonyandrewmeyer, line 950: Same as above, this can be simplified because they'll always be present and always be a string.

jubilant/_juju.py
```diff
@@ -837,6 +831,135 @@ def scp(
...
+                )
+                for revision in obj['revisions']
+            ]
+            if 'revisions' in obj
+            else [],
+            expires=None,
```
tonyandrewmeyer, line 956: This should handle the expiry being set.

jubilant/_juju.py
```diff
@@ -837,6 +831,135 @@ def scp(
...
+            label=obj.get('label', None),
+            error='',
```
tonyandrewmeyer, line 958: Same as above, I think we should leave this out or find out when it can be set.

jubilant/_juju.py
```diff
@@ -837,6 +831,135 @@ def scp(
...
+            rotates=obj.get('rotates', None),
```
tonyandrewmeyer, line 959: Same as above, this should be a datetime if present.

jubilant/secrettypes.py
```diff
@@ -0,0 +1,106 @@
...
+            return str(self)
+
+
+@dataclasses.dataclass
+class SecretAccess:
+    """Represents access to a secret."""
+
+    target: str
+    scope: str
+    role: str
```
tonyandrewmeyer, line 38: I think there are only 2 possible values for role, so it would be better as an enum (or even Literal). However, Ben might be less keen on that, so feel free to leave as-is and we can confirm when he's back.

tonyandrewmeyer (commented):
Thanks for continuing with this! Even though Ben did recommend the API shape previously, he's very keen on closely monitoring the Jubilant API in its early stages, so I think we should hold off merging this until he's back (which is only a couple more weeks). In the meantime, we can get this so that it's very likely to get his approval quickly.

I've added a few comments particularly around the fields.

jubilant/_juju.py
```diff
@@ -837,6 +831,135 @@ def scp(
...
+                revision=obj.get('revision', 1),
+                description=obj.get('description', ''),
+                created=datetime.datetime.fromisoformat(
+                    str(obj.get('created', '').replace('Z', '+00:00'))
+                ),
+                updated=datetime.datetime.fromisoformat(
+                    str(obj.get('updated', '').replace('Z', '+00:00'))
+                ),
+                content=None,
+                checksum='',
```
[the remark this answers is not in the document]
james-garner-canonical: `checksum` shows up in the CLI response for me when `--reveal` is used.

If we continued with the `Secret[Hidden]` / `Secret[Revealed]` approach to typing, we could link the typing of `checksum` to it with a `TypeVar`, but this complicates the implementation, so I suggest we either stick with `str` (with a default of an empty string), or go with `str | None`.

jubilant/_juju.py
james-garner-canonical: Surprisingly, you actually get a better developer experience from leaving out the invalid `overload` rather than adding it with `NoReturn`. With `NoReturn`, subsequent code is marked as unreachable, but the type checker doesn't otherwise complain. When you leave out the invalid `overload`, the type checker will complain at the call site. So I'd recommend against adding a `NoReturn` `overload`.

jubilant/_juju.py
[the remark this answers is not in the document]
tonyandrewmeyer: Ah, sorry, I'm wrong, you're right: `checksum` is there if you get the content. I completely missed that. So ignore my comment, it should definitely be there, I think `str | None` is most correct.

(I don't think the charm can get the checksum. It's not in `secret-info-get`, which only has the metadata, and it's not in `secret-get`, which has the content).

jubilant/secrettypes.py
```diff
@@ -0,0 +1,106 @@
...
+
+@dataclasses.dataclass
+class Secret(Generic[T]):
+    """Represents a secret."""
+
+    uri: SecretURI
+    revision: int
+    checksum: str
+    expires: str | None
+    rotation: str | Literal['never']
```
tonyandrewmeyer, line 65: Do we need the `Literal['never']` here? Wouldn't that be covered by `str`? If we're going to have the possible values ("daily", "weekly", etc) then it seems like we should have them all.

jubilant/secrettypes.py
```diff
@@ -72,3 +74,33 @@ class Secret(Generic[T]):
     content: T
     revisions: list[SecretRevision]
     access: list[SecretAccess]
+
+
+class SecretResponse(TypedDict, total=False):
```
james-garner-canonical, line 79: These `TypedDict`s should be private, as they're just to internally express our assumptions about the data that Juju returns.

jubilant/secrettypes.py
```diff
@@ -0,0 +1,106 @@
...
+            # Handle 'secret://MODEL-UUID/UNIQUE-IDENTIFIER'
+            return self.rsplit('/', maxsplit=1)[-1]
+        elif self.startswith('secret:'):
+            # Handle common case of 'secret:UNIQUE-IDENTIFIER'
+            return self[len('secret:') :]
+        else:
+            return str(self)
+
+
+@dataclasses.dataclass
```
james-garner-canonical, line 32: These should be `frozen=True`.

jubilant/secrettypes.py
```diff
@@ -0,0 +1,74 @@
...
+    revision: int
+    backend: str
+    created: datetime.datetime
+    updated: datetime.datetime
+
+
+NoneType = type(None)
+Hidden = NoneType
+Revealed = Dict[str, str]
+T = TypeVar('T', Revealed, Hidden)
```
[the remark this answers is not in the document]
james-garner-canonical: There's a little bit of design work that needs to be settled around the `Secret` type(s). Feel free to hold off on this suggestion until we work that out.

jubilant/secrettypes.py
james-garner-canonical: By the way, if we clean up the construction of the `Secret` instances well enough (e.g. moving the logic to a `Secret._from_dict` method + @tonyandrewmeyer's various corrections), then we would probably drop the `TypedDict`s altogether, seeing as the various `statustype` types just use `_from_dict(self, d: dict[str, Any])`. Feel free to move in that direction. Apologies for any contradictory advice -- I was hoping the `TypedDict`s would be helpful for expressing which fields are always there, etc., to help us clean up the construction logic.

jubilant/_juju.py
james-garner-canonical: I think we still need some more design around these fields -- the fields whose presence or absence is determined by which Juju command / flags are used. I suggest we keep the implementation simple for now and just make these fields `| None`.

Currently, the `content` field is handled differently, with a `Generic` and a `TypeVar`. This is nice, but doesn't scale well to all the fields that in principle could use the same treatment. We should think about whether `content` is special enough to be worth treating differently. I suspect we may want to go with `| None` here too, and drop the `TypeVar`.

It would be nice though if `secrets` and `get_secrets` (via its overloads) could clearly communicate to the types to the user. It would be unfortunate if users had to write `assert content is not None` after every `show_secret(..., reveal=True)`, or `assert revisions is not None` after every `show_secret(..., revisions=True)`, or `assert access is not None` after every `show_secret` call.

If we start with a single, non-`Generic` `Secret` class with `| None` on a lot of fields, which could implement a `_from_dict` method that takes care of most (all?) of the constructor logic from `secrets` and `get_secret`, then we could narrow the types for our users with thin subclasses of `Secret` that only update the type annotations -- but lets try to get the rest of the PR ironed out before trying to add these subclasses.

jubilant/_juju.py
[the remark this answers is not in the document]
james-garner-canonical: As per comments elsewhere, I think we should go with `None` when the `checksum` isn't present.

jubilant/secrettypes.py
kayra1: yes, I'm still working on this

jubilant/_juju.py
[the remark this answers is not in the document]
kayra1: I have no idea for error, but here's the output with access:
```
{
  "d0vdtqnmp25c762e7jug": {
    "revision": 2,
    "checksum": "0a3322b0d9c2b3ac8f7bae0da21c8828d69ada7d64cb26c0f5c55d2706971c91",
    "owner": "<model>",
    "name": "admin-account",
    "created": "2025-06-03T11:39:23Z",
    "updated": "2025-06-03T11:39:35Z",
    "content": {
      "Data": {
        "password": "hunter2",
        "username": "josh"
      }
    },
    "access": [
      {
        "target": "application-example-charm",
        "scope": "model-7fc4913e-e356-4783-8e80-f2ac9f19759e",
        "role": "view"
      }
    ]
  }
}
```

Here's one with expire and rotation:
```
{
  "d1eiag83594sp69hjut0": {
    "revision": 1,
    "checksum": "7f163d49d8cc63abc0ce15ce201362a9cf2f95ab38b3a59bdbce60c79ca67ed7",
    "expires": "2025-06-27T10:47:30Z",
    "rotation": "quarterly",
    "rotates": "2025-09-26T10:47:30Z",
    "owner": "example-charm",
    "description": "some description",
    "label": "example-secret-label",
    "created": "2025-06-26T10:47:30Z",
    "updated": "2025-06-26T10:47:30Z",
    "content": {
      "Data": {
        "password": "secret",
        "username": "admin-test"
      }
    }
  }
}
```

jubilant/_juju.py
```diff
@@ -837,6 +831,135 @@ def scp(
...
+            description=obj.get('description', ''),
+            name=obj.get('name', None),
+            created=datetime.datetime.fromisoformat(
+                str(obj.get('created', '').replace('Z', '+00:00'))
+            ),
+            updated=datetime.datetime.fromisoformat(
+                str(obj.get('updated', '').replace('Z', '+00:00'))
+            ),
+            content=obj.get('content', {}).get('Data', None),
+            access=[],
```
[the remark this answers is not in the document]
kayra1: what other roles are there for the enum?

jubilant/_juju.py
```diff
@@ -837,6 +831,135 @@ def scp(
...
+        """
+        stdout = self.cli('secrets', '--format', 'json')
+        output: dict[str, SecretResponse] = json.loads(stdout)
+        return [
+            Secret[Hidden](
+                uri=SecretURI('secret:' + uri_from_juju),
+                name=obj.get('name', None),
+                owner=obj.get('owner', '<model>'),
+                rotation=obj.get('rotation', 'never'),
+                revision=obj.get('revision', 1),
```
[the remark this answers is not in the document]
kayra1: updated

jubilant/_juju.py
```diff
@@ -837,6 +831,135 @@ def scp(
...
+                created=datetime.datetime.fromisoformat(
+                    str(obj.get('created', '').replace('Z', '+00:00'))
+                ),
+                updated=datetime.datetime.fromisoformat(
+                    str(obj.get('updated', '').replace('Z', '+00:00'))
+                ),
+                content=None,
+                checksum='',
+                expires=None,
+                rotates=None,
```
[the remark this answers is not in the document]
kayra1: done

jubilant/_juju.py
```diff
@@ -837,6 +831,135 @@ def scp(
...
+                label=None,
```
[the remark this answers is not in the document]
kayra1: some of these were placeholders until i could inspect the CLI's output more closely. updated them all

jubilant/_juju.py
kayra1: updated

jubilant/_juju.py
kayra1: done

jubilant/secrettypes.py
kayra1: done

jubilant/_juju.py
tonyandrewmeyer: > I agree with @james-garner-canonical, marking it as NoReturn indicates that the function returned successfully but is expected to return nothing, when instead this would probably be a reason to raise an `Error` of some sort.

No, that's not the meaning of `NoReturn`. A Python function can't return nothing - if there's no `return` statement (or a plain `return`) then that's an implicit `return None`. If you want to say that it returned successfully without any other value, that's a `-> None` return type. [`NoReturn` means that the function never returns](https://docs.python.org/3/library/typing.html#typing.NoReturn).

If both are provided, then what will happen is that the process will exit with a non-zero exit code, and Jubilant will raise a `CLIError`. That's exactly the sort of situation that `NoReturn` indicates.

@james-garner-canonical's point is that with something like this:

```python
@overload
def func(*, a: int, b: None = None) -> int: ...

@overload
def func(*, a: None = None, b: str) -> str: ...

@overload
def func(*, a: int, b: str) -> NoReturn: ...

def func(*, a: int | None = None, b: str | None = None) -> int | str:
    if a is not None and b is not None:
        raise TypeError("Cannot specify both 'a' and 'b'")
    return f"{a} {b}"

print(func(a=5, b="World"))
print(func(a=5))
print(func(b="Hello"))
```

In your IDE, you should expect to see that the last two lines are unreachable (VS Code will put them in a lighter colour, for example, with a hover that explains that Pylance sees them as unreachable).

However, running pyright will not show any problems. If I remove that overload, then pyright (and Pylance) will show that the call with both args doesn't have a matching overload.

So I'm fine leaving it out, but it's for that reason, not because it's the wrong annotation.

jubilant/secrettypes.py
```diff
@@ -0,0 +1,125 @@
...
+    backend: str
+    created: str
+    updated: str
+
+class _SecretAccessResponse(TypedDict):
+    """TypedDict for access response that arrives from the juju CLI."""
+
+    target: str
+    scope: str
+    role: Literal['read',]  # TBD
```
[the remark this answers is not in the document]
kayra1: I think enums are fine here, since we already have one defined it's not introducing too much noise to add 2 more, and it's just as simple to add a new field to an enum.

I've also added `grant_secret`, it's a much simpler command than the rest and relevant to the rest of the PR.

jubilant/_juju.py
james-garner-canonical: My 2c:

I'm also -1 on multiple `TypeVar`s.

I'm reasonably convinced that `content` is likely to be the most common case for checking an optional field, and I quite like a single `TypeVar` for that.

While it does seem like a shame to make users do extra work for `revisions` and `access`, I'm ok with making that tradeoff to keep the implementation cleaner.

jubilant/_juju.py
```diff
@@ -503,6 +495,16 @@ def exec(
         task.raise_on_failure()
         return task
 
+    def grant_secret(self, name_or_uri: str | SecretURI, applications: Iterable[str]) -> None:
```
tonyandrewmeyer, line 498: General comment about all the `name_or_uri` arguments: I don't love this name. It's accurate and descriptive, but `grant_secret(name_or_uri='something')` feels pretty awkward.

What if we made this positional-only everywhere that we have it, which would let us rename it to something else later in the future (and remove the positional-only restriction at that time)? I think people will typically be using it positionally anyway.

james-garner-canonical: I quite like the suggestion of making them positional only.

jubilant/_juju.py
```diff
@@ -503,6 +495,16 @@ def exec(
...
+        """Grant access to a secret to a list of applications.
+
+        Args:
+            name_or_uri: The name or URI of the secret to grant access to.
+            applications: Iterable of application names to grant access to.
```
tonyandrewmeyer, line 503: I think the pattern we're converging on in Jubilant is to have this as a `str | Iterable[str]` and then in the code have an `if isinstance(applications, str)`, like:

```python
if not isinstance(applications, str):
    applications = ','.join(applications)
```

jubilant/_juju.py
tonyandrewmeyer: Do you still have code that you haven't pushed, maybe? It's not updated as far as I can see here.

jubilant/_juju.py
kayra1: done

jubilant/_juju.py
kayra1: sorry, it looks like i missed some. i've updated it now.

jubilant/_juju.py
```diff
@@ -837,6 +857,142 @@ def scp(
...
+        """
+        stdout = self.cli('secrets', '--format', 'json')
+        output: dict[str, _SecretResponse] = json.loads(stdout)
+        return [
+            Secret[Hidden](
+                uri=SecretURI('secret:' + uri_from_juju),
+                name=obj.get('name'),
+                label=obj.get('label'),
+                owner=obj.get('owner'),
+                rotates=datetime.datetime.fromisoformat(obj['rotates'].replace('Z', '+00:00'))
```
benhoyt, line 874: I'd prefer this wrapped in a little `_datetime_from_iso` helper which took `str | None` and returned `datetime | None`. Then you could use it in all these places, and here with:

```suggestion
                rotates=_datetime_from_iso(obj.get('rotates')),
```

jubilant/_juju.py
```diff
@@ -905,6 +1061,40 @@ def trust(
...
+        auto_prune: bool = False,
+    ) -> None:
+        """Update the content of a secret.
+
+        Args:
+            name_or_uri: The name or the secret uri for the secret.
+            content: Key-value pairs that represent the secret content, for example
+                ``{'password': 'hunter2'}``.
+            info: Optional description for the secret.
+            name: Optional new name for the secret.
```
benhoyt, line 1081: The type annotations indicate these are optional, so I'd do:

```suggestion
            info: New description for the secret.
            name: New name for the secret.
```

jubilant/secrettypes.py
```diff
@@ -0,0 +1,143 @@
...
+
+import dataclasses
+import datetime
+import enum
+from typing import Dict, Generic, TypedDict, TypeVar
+
+from typing_extensions import Required
+
+
+class SecretRotateCadence(enum.Enum):
```
benhoyt, line 13: Could these type names all drop the `Secret` prefix, given they're in the `secrettypes` module? Also, this is called `SecretRotate` in ops, so I think we can drop the `Cadence` suffix too.
```suggestion
class Rotate(enum.Enum):
```

But obviously not for `Secret` itself, and `SecretURI` is imported to the top level, so  not that one either.

jubilant/secrettypes.py
```diff
@@ -0,0 +1,143 @@
...
+    MONTHLY = 'monthly'
+    QUARTERLY = 'quarterly'
+    YEARLY = 'yearly'
+
+
+class SecretAccessScope(enum.Enum):
+    """Secret access scopes."""
+
+    UNIT = 'unit'
+    APPLICATION = 'application'
```
benhoyt, line 29: Given our renames of "app" to "application" throughout Jubilant, I suggest:

```suggestion
    APP = 'application'
```

jubilant/secrettypes.py
```diff
@@ -0,0 +1,143 @@
...
+    expires: str | None
+    rotation: SecretRotateCadence | None
+    rotates: datetime.datetime | None
+    owner: str
+    description: str | None
+    name: str | None
+    label: str | None
+    created: datetime.datetime
+    updated: datetime.datetime
+    content: T
```
benhoyt, line 84: I'm not a fan of this generic stuff here, and the associated type complexity. Can we just make this `content: dict[str, str] | None`, and it's only set if you do `show_secret(reveal=True)`? I think all the overloads and generics here are overkill, and would much rather it be simpler.

jubilant/_juju.py
```diff
@@ -837,6 +857,142 @@ def scp(
...
+            name=obj.get('name'),
+            label=obj.get('label'),
+            owner=obj.get('owner'),
+            rotates=datetime.datetime.fromisoformat(obj['rotates'].replace('Z', '+00:00'))
+            if 'rotates' in obj
+            else None,
+            rotation=obj.get('rotation'),
+            expires=obj.get('expires'),
+            revision=obj.get('revision'),
+            description=obj.get('description', ''),
```
benhoyt, line 964: Why is this typed as `str | None`, but we use a default of `''` here?

jubilant/secrettypes.py
```diff
@@ -0,0 +1,143 @@
...
+    revisions: list[SecretRevision] | None
+    access: list[SecretAccess] | None
```
benhoyt, line 86: `list | None` types are awkward. Can we just make these `[]` empty list if not present?

jubilant/_juju.py
```diff
@@ -837,6 +857,142 @@ def scp(
...
+            created=datetime.datetime.fromisoformat(obj['created'].replace('Z', '+00:00')),
+            updated=datetime.datetime.fromisoformat(obj['updated'].replace('Z', '+00:00')),
+            content=obj.get('content', {}).get('Data', None),
+            checksum=obj.get('checksum', ''),
+            access=[
+                SecretAccess(
+                    role=SecretAccessRole(access.get('role')),
+                    scope=SecretAccessScope(access.get('scope')),
+                    target=access.get('target'),
```
benhoyt, line 973: Let's move this to a `SecretAccess._from_dict` helper (same for `SecretRevision._from_dict`), similar to how we parse the big `Status` object.

benhoyt (changes requested):
Thanks, this is getting close. I'd like to see the types simplified (and the TypeDicts removed), and a unit test file should be added for each new command.

jubilant/secrettypes.py
tonyandrewmeyer: I believe I suggested these previously. (None, or a "not retrieved" sentinel, or subclasses, I think).

I don't think we use `None` for `access`, because then how do you distinguish between "no-one has access" and "I got this from the command that doesn't provide access information"? I guess the answer is that you distinguish it by the command that was run, but I don't think that's a good API.

If `None` means "I don't know the access", then it seems more consistent to have that same meaning for `revisions`, but there isn't an issue there as "no revisions" is impossible.

I think the challenge here comes from `juju secrets` returning *almost* everything you get from a non-reveal `juju show-secret`. If `juju secrets` only gave limited information, I think `secrets()` would probably return a sequence of identifiers, and then you'd use `show_secret()` with each of them. Or if `juju secrets` gave all the data other than the content, then it'd obviously be straightforward for `secrets()` to return the same object as `show_secret()` without reveal.

jubilant/secrettypes.py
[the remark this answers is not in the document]
james-garner-canonical: Resolving this thread as it sounds like we'll go with the simplest implementation: drop the `Generic` and `overload`s and just type the `Secret.content` as `| None`.

jubilant/secrettypes.py
james-garner-canonical: I do quite like the user experience of knowing the type of `Secret.content` depending on whether you use `reveal=True` or not.

However, the simpler approach fits well with the overall `jubilant` philosophy and ux and will likely be easier to maintain in the long run, so I'm on board with the simpler approach.

Developers will hopefully be guided by their IDEs and the docs to write (e.g.) `assert secret.content is not None` after calling `show_secret(..., reveal=True)`.

jubilant/secrettypes.py
```diff
@@ -0,0 +1,143 @@
...
+@dataclasses.dataclass(frozen=True)
+class SecretAccess:
+    """Represents access to a secret."""
+
+    target: str
+    scope: SecretAccessScope
+    role: SecretAccessRole
+
+
+class _SecretResponse(TypedDict, total=False): # pyright: ignore[reportUnusedClass]
```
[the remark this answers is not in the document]
james-garner-canonical: These were at least helpful for me in getting my head around what the `Secret` constructors should look like, but with those in good shape I'm on board with dropping these TypedDicts.

jubilant/_juju.py
```diff
@@ -503,6 +495,20 @@ def exec(
         task.raise_on_failure()
         return task
 
+    def grant_secret(
+        self, name_or_uri: str | SecretURI, /, applications: str | Iterable[str]
```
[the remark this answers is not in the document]
kayra1: I do not have strong opinions on this. changed the `name_or_uri` to `identifier` and `applications` to `app`

jubilant/_juju.py
kayra1: We have docs that start with Optional on `run` and `deploy` too. Should I remove those too?

jubilant/_juju.py
```diff
@@ -503,6 +495,20 @@ def exec(
...
+    ) -> None:
+        """Grant access to a secret to a list of applications.
```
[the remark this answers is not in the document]
kayra1: done

jubilant/secrettypes.py
kayra1: done

jubilant/secrettypes.py
kayra1: done

tests/integration/test_secrets.py
```diff
@@ -14,3 +15,78 @@ def test_add_secret(juju: jubilant.Juju):
...
+        },
+        {
+            'name': 'sec2',
+            'username': 'usr2',
+            'password': 'hunter3',
+            'info': 'A new description.',
+            'revision': 2,
+        },
+    ]
+    secrets = juju.secrets()
```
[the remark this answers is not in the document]
kayra1: done

jubilant/secrettypes.py
kayra1: I will defer to your opinion if you think this is for the best, but I would personally be really disappointed if this feature was taken out.

If jubilant becomes exactly 35 lines "simpler", then the effort for type narrowing is pushed to the users who will have to `assert` every field every time they use it in every charm project. A large issue with pytest_operator is the fact that half of our tests are assert statements on fields we know for a fact are there, but they are typed optional and we have to keep pyright happy. 

Unlike for example `description`, the `content` field is not optional. The user enables or disables that field explicitly. I think this is as large of a distinction in the type as tagging the return value as `int` for a function that actually returns a `str`.

Not to mention jubilant already has overloads for `config` for exactly the same purpose as this, which is great to use.

jubilant/_juju.py
```diff
@@ -837,6 +842,79 @@ def scp(
...
+        identifier: str | SecretURI,
+        *,
+        reveal: bool = False,
+        revision: int | None = None,
+        revisions: bool = False,
+    ) -> Secret[Revealed] | Secret[Hidden]:
+        """Get the content of a secret.
+
+        Args:
+            identifier: Name or URI of the secret to reveal.
```
benhoyt, line 899: Because we're only "reveal"ing it if reveal=True, let's change this to:

```suggestion
            identifier: Name or URI of the secret to return.
```

jubilant/_juju.py
```diff
@@ -905,6 +983,39 @@ def trust(
...
+        content: Mapping[str, str],
+        *,
+        info: str | None = None,
+        name: str | None = None,
+        auto_prune: bool = False,
+    ) -> None:
+        """Update the content of a secret.
+
+        Args:
+            identifier: The name or the secret uri for the secret.
```
benhoyt, line 998:
```suggestion
            identifier: The name or URI of the secret to update.
```

jubilant/secrettypes.py
```diff
@@ -0,0 +1,146 @@
...
+
+from __future__ import annotations
+
+import dataclasses
+import datetime
+import enum
+from typing import Any, Dict, Generic, TypeVar
+
+
+class Rotate(enum.Enum):
```
benhoyt, line 11: @james-garner-canonical Do you think we should make these `(str, enum.Enum)` (equivalent of StrEnum but on versions less than Python 3.13) as plain enums are annoying to deal with with the `.value` thing?

benhoyt: Actually, I think we should drop these enum types and just make them `str`. We haven't used enum types for `statustypes` fields (or anywhere else in Jubilant), and it's more flexible and future-proof. I also don't think these will be used much at all.

jubilant/secrettypes.py
```diff
@@ -0,0 +1,146 @@
...
+    @classmethod
+    def _from_dict(cls, d: dict[str, Any]) -> Secret[T]:
+        return cls(
+            uri=SecretURI('secret:' + d.get('uri', '')),
+            name=d.get('name'),
+            label=d.get('label'),
+            owner=d.get('owner', ''),
+            rotates=datetime.datetime.fromisoformat(d['rotates'].replace('Z', '+00:00'))
+            if 'rotates' in d
+            else None,
```
benhoyt, line 95: Yeah, so let's add that `_datetime_from_iso` helper and use it here (and elsewhere) like so:

```suggestion
            rotates=_datetime_from_iso(d['rotates']) if 'rotates' in d else None,
```

jubilant/_juju.py
kayra1: updated the `run` and `deploy` docstring too.

jubilant/_juju.py
```diff
@@ -689,6 +682,18 @@ def remove_relation(self, app1: str, app2: str, *, force: bool = False) -> None:
             args.append('--force')
         self.cli(*args)
 
+    def remove_secret(self, identifier: str | SecretURI, *, revision: int | None = None) -> None:
+        """Remove a secret from the model.
+
+        Args:
+            identifier: The URI of the secret to remove.
```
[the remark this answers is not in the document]
kayra1: done

jubilant/_juju.py
kayra1: done

jubilant/__init__.py
```diff
@@ -14,15 +14,17 @@
     any_maintenance,
     any_waiting,
 )
-from ._juju import CLIError, ConfigValue, Juju, SecretURI, WaitError
+from ._juju import CLIError, ConfigValue, Juju, WaitError
 from ._task import Task, TaskError
 from ._test_helpers import temp_model
+from .secrettypes import Secret, SecretURI
```
benhoyt, line 20: I guess we should add `RevealedSecret` here too now (and to `__all__`), as `Juju.show_secret` can return that type.

benhoyt (commented):
Looking good, thanks for the updates. Just the unit tests now.

kayra1:
> Just the unit tests now.

I've added all the unit tests and a `File` mock for the file writing in add and update secret

jubilant/__init__.py
kayra1: done

jubilant/_juju.py
```diff
@@ -837,6 +842,82 @@ def scp(
...
+        revision: int | None = None,
+        revisions: bool = False,
+    ) -> Secret | RevealedSecret:
+        """Get the content of a secret.
+
+        Args:
+            identifier: Name or URI of the secret to return.
+            reveal: Whether to reveal the secret content.
+            revisions: Whether to include all revisions of the secret. Mutually
+                exclusive with `reveal` and `revision`.
```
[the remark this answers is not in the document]
kayra1: done

jubilant/secrettypes.py
```diff
@@ -0,0 +1,124 @@
...
+            return str(self)
+
+
+@dataclasses.dataclass(frozen=True)
+class Secret:
+    """Represents a secret."""
+
+    uri: SecretURI
+    revision: int
+    checksum: str | None
```
james-garner-canonical, line 36: Should `checksum` be an attribute of `RevealedSecret` instead? I believe it's only returned by Juju when the secret content is provided.

jubilant/secrettypes.py
```diff
@@ -0,0 +1,124 @@
...
+        for example "9m4e2mr0ui3e8a215n4g").
+        """
+        if '/' in self:
+            # Handle 'secret://MODEL-UUID/UNIQUE-IDENTIFIER'
+            return self.rsplit('/', maxsplit=1)[-1]
+        elif self.startswith('secret:'):
+            # Handle common case of 'secret:UNIQUE-IDENTIFIER'
+            return self[len('secret:') :]
+        else:
+            return str(self)
```
dimaqq, line 27: **minor:** doc string could be better.

**potential issue:** the class title doesn't appear to be consistent with helper property.

Does the secret URI have to to start with `secret:` or not?

jubilant/secrettypes.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+@dataclasses.dataclass(frozen=True)
+class Secret:
+    """Represents a secret."""
```
dimaqq, line 32: **clarification needed:** any secret? A Juju secret? A user secret? A charm secret?

jubilant/secrettypes.py
```diff
@@ -0,0 +1,124 @@
...
+    rotation: str | None
+    rotates: datetime.datetime | None
+    owner: str
+    description: str | None
+    name: str | None
+    label: str | None
+    created: datetime.datetime
+    updated: datetime.datetime
+    revisions: list[Revision] | None
+    access: list[Access] | None
```
dimaqq, line 47: **API design:** it feels weird that so many fields are optional.

When are they optional and when are they provided?

Note that if a field is marked smth like "omitempty" in Go source code doesn't mean it's actually optional, as some of these fields are never actually empty.

If `| None` have semantic content, perhaps some helper properties could be used, like `.is_user_secret / .is_charm_secret` or `.is_revealed` or what not. WDYT?

jubilant/secrettypes.py
```diff
@@ -0,0 +1,124 @@
...
+        return cls(
+            target=d['target'],
+            scope=d['scope'],
+            role=d['role'],
+        )
+
+
+def _datetime_from_iso(dt: str) -> datetime.datetime:
+    """Converts a Juju-specific ISO 8601 string to a datetime object."""
+    return datetime.datetime.fromisoformat(dt.replace('Z', '+00:00'))
```
dimaqq, line 124: This is great, except Python famously fails to part fractional parts if they contain unusual number of digits.

Sorry I don't recall what Python version, it may have been fixed.

In short:
- `xxx:59.123` OK
- `xxx:59.120` OK
- `xxx:59.12` Error

See https://github.com/python/cpython/issues/95221

dimaqq (commented):
A partial review from me at this point.

jubilant/secrettypes.py
tonyandrewmeyer: > **potential issue:** the class title doesn't appear to be consistent with helper property.

This class is already in Jubilant (it's being moved from one module to another), so we can't change the name without breaking compatibility.

jubilant/secrettypes.py
tonyandrewmeyer: > **API design:** it feels weird that so many fields are optional.

That's the Juju design.

> When are they optional and when are they provided?

There's quite a bit of discussion around this in the (mostly resolved) comments. But generally, things like `expires`, `description`, and `rotation` are provided when the secret has expiry or rotation set. `revisions` and `access` are provided when using `show_secret()` and not when using `secrets()` -- there's a bunch of discussion about how to handle this in the comments, too - I don't think it's worth re-opening that.

> If `| None` have semantic content, perhaps some helper properties could be used, like `.is_user_secret / .is_charm_secret` or `.is_revealed` or what not. WDYT?

I believe we've (Kayra, Ben, James, me) settled on using a subclass for revealed (so you can `isinstance(s, RevealedSecret)` although `hasattr(s, 'content')` would be more Pythonic.

For the others, I don't think a `has_description` property is better than `description != None` (same for expiry, rotation, label). `revisions` and `access` will be best served by clear docs that explain which method to use to have them provided, I believe.

I'm not sure whether a `is_user_secret` property offers much over `.owner == '<model>'`. It doesn't seem like something most tests would be checking for, either.

james-garner-canonical: I agree with @tonyandrewmeyer's comment.

> you can isinstance(s, RevealedSecret) although hasattr(s, 'content') would be more Pythonic

Though note that the intended design is that you can safely assume you have a `RevealedSecret` because you called `s = show_secret(..., reveal=True)`, allowing you to do the most Pythonic thing possible and write `s.content` and let things explode at runtime if it's not there ;)

jubilant/secrettypes.py
tonyandrewmeyer: See also [this earlier discussion](https://github.com/canonical/jubilant/pull/135#discussion_r2165088276).

kayra1: If I'm understanding the linked issue correctly, only 3 or 6 digit decimal places work, any other number of decimal digits produce an error, but only if there are decimal digits for seconds.

Right now, Juju doesn't return any decimals in the datetime string, and I'm confident that the integration tests would fail if the juju response ever changes the format of the datetime strings, so I still think it's still fine to keep it as is.

jubilant/secrettypes.py
kayra1: that seems to be true, i've moved it there.

kayra1:
> It looks like @kayra1 also needs to sign [the CLA](https://ubuntu.com/legal/contributors).

I will rebase this PR to only use my work email after getting an approval.

tests/unit/conftest.py
```diff
@@ -23,3 +23,11 @@ def time(monkeypatch: pytest.MonkeyPatch) -> Generator[mocks.Time]:
...
     monkeypatch.setattr('time.sleep', time_mock.sleep)
     yield time_mock
+
+
+@pytest.fixture
+def file(monkeypatch: pytest.MonkeyPatch) -> Generator[mocks.File]:
+    """Pytest fixture that patches tempfile.NamedTemporaryFile with mocks.File."""
+    file_mock = mocks.File()
+    monkeypatch.setattr('tempfile.NamedTemporaryFile', file_mock)
+    yield file_mock
```
james-garner-canonical, line 33: I wonder if we should just write to the temporary file, like we do in the [unit tests for run](https://github.com/canonical/jubilant/blob/c1a160e0f1bf485340b9b00016f6e2af330a4fde/tests/unit/test_run.py#L159). On the other hand, I do quite like the testing story this provides over having to mock out run to check the file contents, so maybe we shoulds tick with this approach here, and that test should be updated to use this fixture (not in this PR). WDYT @benhoyt?

tests/unit/test_add_secret.py
```diff
@@ -1,44 +1,30 @@
...
+def test_normal(run: mocks.Run, file: mocks.File):
+    run.handle(
+        ['juju', 'add-secret', 'my-secret', '--file', file.name],
+        stdout='secret:0123456789abcdefghji\n',
+    )
     juju = jubilant.Juju()
 
-    secret_uri = juju.add_secret(
-        'sec1', {'username': 'usr', 'password': 'hunter2'}, info='A description.'
+    juju.add_secret('my-secret', {'username': 'admin'})
```
james-garner-canonical, line 14: Let's not remove the return type and content assertions and the exercising of the info argument.

jubilant/secrettypes.py
```diff
@@ -0,0 +1,125 @@
...
+    rotation: str | None
+    rotates: datetime.datetime | None
+    owner: str
+    description: str | None
+    name: str | None
+    label: str | None
+    created: datetime.datetime
+    updated: datetime.datetime
+    revisions: list[Revision] | None
+    access: list[Access] | None
```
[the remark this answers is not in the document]
kayra1: The order is from Ben's initial comment. I don't mind either way, it's something users won't see.

jubilant/secrettypes.py
```diff
@@ -0,0 +1,125 @@
...
+
+    @classmethod
+    def _from_dict(cls, d: dict[str, Any]) -> Secret:
+        return Secret(
+            uri=SecretURI('secret:' + d.get('uri', '')),
```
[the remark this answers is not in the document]
kayra1: done

jubilant/secrettypes.py
```diff
@@ -0,0 +1,125 @@
...
+            description=d.get('description'),
+            created=_datetime_from_iso(d['created']),
+            updated=_datetime_from_iso(d['updated']),
+            expires=d.get('expires'),
+            access=[Access._from_dict(access) for access in d.get('access', [])]
+            if 'access' in d
+            else None,
+            revisions=[Revision._from_dict(revision) for revision in d.get('revisions', [])]
+            if 'revisions' in d
+            else None,
```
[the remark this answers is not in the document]
kayra1: done

jubilant/_juju.py
```diff
@@ -905,6 +991,39 @@ def trust(
...
+                ``{'password': 'hunter2'}``.
+            info: New description for the secret.
+            name: New name for the secret.
+            auto_prune: automatically remove revisions that are no longer tracked by any observers.
+        """
+        args = ['update-secret', identifier]
+        if name is not None:
+            args.extend(['--name', name])
+        if info is not None:
+            args.extend(['--info', info])
```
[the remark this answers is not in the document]
kayra1: done

jubilant/_juju.py
```diff
@@ -837,6 +841,88 @@ def scp(
...
+    ) -> Secret | RevealedSecret:
+        """Get the content of a secret.
+
+        Args:
+            identifier: Name or URI of the secret to return.
+            reveal: Whether to reveal the secret content.
+            revisions: Whether to include all revisions of the secret. Mutually
+                exclusive with *reveal* and *revision*.
+            revision: Revision number of the secret to reveal. If not specified,
+                the latest revision is revealed.
```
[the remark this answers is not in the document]
kayra1: done

jubilant/secrettypes.py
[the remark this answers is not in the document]
james-garner-canonical: Let's get @benhoyt's take on the ordering then, and whatever we go with, let's update the order of the arguments in the constructor to match for ease of reading.

tests/unit/test_add_secret.py
james-garner-canonical: Makes sense, good call.

benhoyt (approved):
Approving optimistically as this is very close, but a few remaining comments: one on the overloads, and several on the unit tests.

jubilant/_juju.py
```diff
@@ -837,6 +841,88 @@ def scp(
...
+        self,
+        identifier: str | SecretURI,
+        *,
+        reveal: Literal[False] = False,
+        revision: int | None = None,
+        revisions: Literal[False] = False,
+    ) -> Secret: ...
+
+    @overload
+    def show_secret(
```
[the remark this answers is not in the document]
james-garner-canonical: In addition to the two different return types, the three overloads capture the permitted combinations of the three Juju options, `--reveal`, `--revision`, and `--revisions`.

Overloads:
1. If you set `reveal=True` you may specify a `revision`, but you can't specify `revisions`.
2. With `reveal=False`, you may specify a `revision` as long as `revisions=False`;
3. but when `revisions=True` , the other two arguments must be false.

The bad combinations are then excluded by their overloads not being defined (`reveal=True` + `revisions=True`, `revision=...` + `revisions=True`) -- the type checker will complain if you use those combinations.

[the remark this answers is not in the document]
benhoyt: Ah, thanks for the explanation @james-garner-canonical  -- makes sense.

tests/unit/conftest.py
```diff
@@ -23,3 +23,11 @@ def time(monkeypatch: pytest.MonkeyPatch) -> Generator[mocks.Time]:
     monkeypatch.setattr('time.monotonic', time_mock.monotonic)
     monkeypatch.setattr('time.sleep', time_mock.sleep)
     yield time_mock
+
+
+@pytest.fixture
+def file(monkeypatch: pytest.MonkeyPatch) -> Generator[mocks.File]:
```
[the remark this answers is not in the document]
benhoyt: I wanted to link it somehow to "named temporary file" but without quite such a long name. But I'm okay with `mock_file`.

tests/unit/mocks.py
```diff
@@ -80,3 +81,32 @@ def monotonic(self) -> float:
...
     def sleep(self, seconds: float):
         self._monotonic += seconds
+
+
+class File:
+    """Mock for tempfile.NamedTemporaryFile.
+
+    Captures any writes to any temporary files. Each separate call to write
+    adds it to its own row in a log, and each flush only increments the flush count.
+    A temporary file name is also provided both to the mock and to the caller.
```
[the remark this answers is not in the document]
kayra1: done

tests/unit/mocks.py
```diff
@@ -80,3 +81,32 @@ def monotonic(self) -> float:
...
+    """
+
+    def __init__(self):
+        self.write_log: list[str] = []
+        self.flush_count: int = 0
```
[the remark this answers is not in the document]
kayra1: done

tests/unit/test_add_secret.py
```diff
@@ -1,44 +1,31 @@
...
+    )
     juju = jubilant.Juju()
 
-    secret_uri = juju.add_secret(
-        'sec1', {'username': 'usr', 'password': 'hunter2'}, info='A description.'
+    secret_uri = juju.add_secret('my-secret', {'username': 'admin'})
+
+    assert secret_uri.startswith('secret:')
+    assert len(file.write_log) == 1
+    assert file.write_log[0] == 'username: admin\n'
```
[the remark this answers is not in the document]
kayra1: done

tests/unit/test_secrets.py
```diff
@@ -0,0 +1,32 @@
...
+    assert response[0].owner == '<model>'
+    assert response[0].name == 'admin-account'
+    assert response[0].created.day == response[0].updated.day
+
+
+def test_get_empty_secrets(run: mocks.Run):
+    run.handle(['juju', 'secrets', '--format', 'json'], stdout=json.dumps({}))
+    juju = jubilant.Juju()
+
+    juju.secrets()
```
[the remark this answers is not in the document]
kayra1: i felt that it was enough to make sure the function completes without crashing, but added that too

jubilant/secrettypes.py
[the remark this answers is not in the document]
james-garner-canonical: @benhoyt, any opinion here? Want to stick with the original field ordering you suggested, or should we put all the non `| None` fields first? Or a third option?

tests/integration/test_secrets.py
```diff
@@ -14,3 +15,66 @@ def test_add_secret(juju: jubilant.Juju):
...
+
+def test_get_all_secrets(juju: jubilant.Juju):
+    secrets = juju.secrets()
+    assert len(secrets) == 2
+
+    assert secrets[0].revision == 1
+    assert secrets[0].name == 'sec1'
+    assert secrets[0].owner == '<model>'
+    assert secrets[0].description == 'A description.'
+    assert secrets[0].created.year == datetime.datetime.now().year
```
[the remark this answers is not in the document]
kayra1: done

tests/unit/test_show_secret.py
```diff
@@ -0,0 +1,103 @@
...
+def test_get_secret(run: mocks.Run):
+    run.handle(
+        ['juju', 'show-secret', 'example-charm-secret', '--format', 'json'],
+        stdout=json.dumps(SINGLE_SECRET),
+    )
+    juju = jubilant.Juju()
+    response = juju.show_secret('example-charm-secret')
+
+    uri = next(iter(SINGLE_SECRET.keys()))
+    secret = SINGLE_SECRET.get(uri)
```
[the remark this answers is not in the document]
kayra1: done

benhoyt (approved):
Only the field ordering thing left, but approving optimistically.

benhoyt:
Oh, looks like the unit tests are failing too.

jubilant/secrettypes.py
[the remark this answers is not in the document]
kayra1: done

benhoyt:
Tests all look good now. Let's get the CLA check done and merge this! Thanks a lot @kayra1!

kayra1:
> Tests all look good now. Let's get the CLA check done and merge this! Thanks a lot @kayra1!

We made it! Is the version number ok? Should I go through and make a release according to the readme or would you like to do that?

tests/unit/mocks.py
```diff
@@ -80,3 +81,32 @@ def monotonic(self) -> float:
 
     def sleep(self, seconds: float):
         self._monotonic += seconds
+
+
+class NamedTemporaryFile:
```
dimaqq, line 86: I wonder if perhaps stdlib [unittest.mock.mock_open](https://docs.python.org/3/library/unittest.mock.html#mock-open) would have been sufficient instead.

PR #148, opened 2025-06-05
fix: "juju model" doesn't use --model argument

### LINKED ISSUE #147: Juju offer can use invailid --model argument
When the Juju object refers to a specific model, when creating an offer, it will send the argument `--model` to `juju offer`. However, the `--model` argument is not valid for the offer command. See [juju offer](https://documentation.ubuntu.com/juju/3.6/reference/juju-cli/list-of-juju-cli-commands/offer/).

That is, the following code will fail:
```
    juju = jubilant.Juju(model='testing-lxd')
    juju.offer('rabbitmq-server', endpoint='amqp')
```

With an error similar to:
```
    raise CLIError(e.returncode, e.cmd, e.stdout, e.stderr) from None
jubilant._juju.CLIError: Command '['juju', 'offer', '--model', 'testing-lxd', 'rabbitmq-server:amqp']' returned non-zero exit status 2.
Stderr:
ERROR option provided but not defined: --model
```

benhoyt:
When implementing the `Juju.offer` method, I had overlooked that that command doesn't take a `--model` argument, and Juju complains when you do. Remove it.

This command doesn't have any integration tests, but update the unit tests with a negative test of this.

Fixes #147.

### DIFF
--- a/tests/unit/test_offer.py
+++ b/tests/unit/test_offer.py
@@ -11,7 +11,8 @@ def test(run: mocks.Run):
 
 
 def test_with_model(run: mocks.Run):
-    run.handle(['juju', 'offer', '--model', 'mdl', 'mysql:db'])
+    # "juju offer" isn't a model-based command
+    run.handle(['juju', 'offer', 'mysql:db'])
     juju = jubilant.Juju(model='mdl')
 
     juju.offer('mysql', endpoint='db')
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -601,7 +601,7 @@ class Juju:
         if name is not None:
             args.append(name)
 
-        self.cli(*args)
+        self.cli(*args, include_model=False)
 
     def refresh(
         self,


james-garner-canonical (approved):
Nice clean fix

tests/unit/test_offer.py
```diff
@@ -11,7 +11,8 @@ def test(run: mocks.Run):
 
 
 def test_with_model(run: mocks.Run):
-    run.handle(['juju', 'offer', '--model', 'mdl', 'mysql:db'])
+    # "juju offer" isn't a model-based command
+    run.handle(['juju', 'offer', 'mysql:db'])
```
javierdelapuente, line 15: Could it be a better option to do something like `juju offer mdl.mysql:db`, instead of `juju offer mysql:db` so the current model is specified? (see https://documentation.ubuntu.com/juju/3.6/reference/juju-cli/list-of-juju-cli-commands/offer/ for reference)

james-garner-canonical: I think the interesting question here isn't so much about the unit test, but about whether `Juju.model` should be included in the `offer` command. I haven't got my head around all this yet, but if I have two `jibilant.Juju` instances with different models, should running `offer` on each of them produce different commands?

PR #150, opened 2025-06-09
feat: add Status.get_units, include subordinates in all_* and any_*

### LINKED ISSUE #137: Feature request: easier way to get subordinate status
Dylan asked me today about getting subordinate unit status (machine model).

python-libjuju has this helper:
```py
def app_units(full_status: FullStatus, app_name: str) -> dict[str, UnitStatus]:
    """Fish out the app's units' status from a FullStatus response."""
    rv: dict[str, UnitStatus] = {}
    app = full_status.applications[app_name]
    assert isinstance(app, ApplicationStatus)

    if app.subordinate_to:
        parent_name = app.subordinate_to[0]
        parent = full_status.applications[parent_name]
        assert isinstance(parent, ApplicationStatus)
        for parent_unit in parent.units.values():
            assert isinstance(parent_unit, UnitStatus)
            for name, unit in parent_unit.subordinates.items():
                if not name.startswith(f"{app_name}/"):
                    continue
                assert isinstance(unit, UnitStatus)
                rv[name] = unit
    else:
        for name, unit in app.units.items():
            assert isinstance(unit, UnitStatus)
            rv[name] = unit

    return rv
```

We were wondering if a helper like that could be added to jubilant

benhoyt:
This adds Status.get_units to get an application's units, including the units of subordinate applications. It falls back to `apps[app].units` if the application is a principal (non-subordinate) app.

It also updates the all_* and any_* helpers like all_active and any_error to include units of subordinate apps.

The change to the all_* and any_* helpers is actually a breaking change, though I doubt it will affect users, except in a good way -- it's very likely what they expect already, and the docs say "Report whether any app or unit in status is active" without qualifiers, so it should include the units of subordinate apps.

Fixes #137

### DIFF
--- a/tests/unit/test_all_any.py
+++ b/tests/unit/test_all_any.py
@@ -8,7 +8,7 @@ import pytest
 
 import jubilant
 
-from .fake_statuses import MINIMAL_STATUS, SNAPPASS_JSON
+from .fake_statuses import MINIMAL_STATUS, SNAPPASS_JSON, SUBORDINATES_JSON
 
 
 class AllAnyFunc(Protocol):
@@ -201,3 +201,25 @@ def test_all_agents_idle():
     ] = 'error'
     status = jubilant.Status._from_dict(status_dict)
     assert not jubilant.all_agents_idle(status)
+
+
+def test_all_subordinates():
+    status_dict = json.loads(SUBORDINATES_JSON)
+    status_dict['applications']['ubuntu']['units']['ubuntu/1']['subordinates']['nrpe/1'][
+        'workload-status'
+    ]['current'] = 'active'
+    status = jubilant.Status._from_dict(status_dict)
+    assert not jubilant.all_blocked(status, 'nrpe')
+    assert status.apps['nrpe'].is_blocked
+    assert status.apps['ubuntu'].is_active
+    assert status.apps['ubun2'].is_active
+
+
+def test_any_subordinates():
+    status_dict = json.loads(SUBORDINATES_JSON)
+    status_dict['applications']['nrpe']['application-status']['current'] = 'active'
+    status = jubilant.Status._from_dict(status_dict)
+    assert jubilant.any_blocked(status)
+    assert status.apps['nrpe'].is_active
+    assert status.apps['ubuntu'].is_active
+    assert status.apps['ubun2'].is_active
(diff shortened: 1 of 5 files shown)


tests/unit/test_all_any.py
```diff
@@ -201,3 +201,25 @@ def test_all_agents_idle():
...
     status = jubilant.Status._from_dict(status_dict)
     assert not jubilant.all_agents_idle(status)
+
+
+def test_all_subordinates():
+    status_dict = json.loads(SUBORDINATES_JSON)
+    status_dict['applications']['ubuntu']['units']['ubuntu/1']['subordinates']['nrpe/1'][
+        'workload-status'
+    ]['current'] = 'active'
+    status = jubilant.Status._from_dict(status_dict)
```
james-garner-canonical, line 211: Unrelated to this PR, but a random thought about the semi-private `_from_dict` methods -- maybe we should add a public `jubilant.from_dict(Status, status_dict)` function?

james-garner-canonical (approved):
I'm on board with doing this as a bugfix release, assuming we're reasonably sure that people haven't been using `{all,any}_*` with subordinate apps and then working around the subordinate units being skipped ... though even then maybe they'd be happy to get this breakage as it would presumably simplify their code.

tests/unit/test_all_any.py
benhoyt: I would soon make `jubilant.Status.from_dict` public (but probably not all the other `_from_dict` methods on sub-status objects).

james-garner-canonical: Fair enough. If we only wanted to make `from_dict` public for the one top-level class, then I guess a function like `jubilant.status_from_dict(status_dict)` would also work.

benhoyt: For now, though, I think I'll leave it private, unless someone comes and asks with a real need. These are only used in tests, and I don't see them being used externally.

james-garner-canonical: Yeah I don't think this needs addressing in this PR at all, but I'm sure we have recommended to external users to use `_from_dict` before, so I figured I'd mention it.

tonyandrewmeyer: > For now, though, I think I'll leave it private, unless someone comes and asks with a real need. These are only used in tests, and I don't see them being used externally.

I think they could be somewhere like [juju-doctor](https://github.com/canonical/juju-doctor/blob/f8c73340049b53b25a3fe68b16a310f33f5d02d7/src/juju_doctor/artifacts.py#L79) but I agree that it's better to wait until they come asking (we already made them aware that this might be nice on their side).

PR #152, opened 2025-06-10
ci: run TIOBE checks in diagnostic mode

tonyandrewmeyer:
The output shows pylint being installed but TIOBE cannot find it. Maybe diagnostic mode will show why.

Note that we should be able to run the action against the branch so avoid merging this.

### DIFF
--- a/.github/workflows/tiobe.yaml
+++ b/.github/workflows/tiobe.yaml
@@ -18,6 +18,13 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
 
+        # Note that these are dependencies for the tools that the TIOBE action will run,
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer:
Argh! It was [definitely failing in main](https://github.com/canonical/jubilant/actions/workflows/tiobe.yaml?query=branch%3Amain) and now has magically started passing. I tried a bunch of things to fix it and then diagnose what was wrong in this branch, but it seems like none of them were necessary.

Either this was some temporary GitHub glitch, or when I contacted TIOBE about it they fixed something behind the scenes without telling me. Anyway, no need for this PR.

PR #153, opened 2025-06-12
chore: update to latest version of pyright (402)

benhoyt:
This updates to the latest version of Pyright (1.1.402). Due to https://github.com/microsoft/pyright/issues/10277 we need to silence the "Type of X is partially unknown" errors on lines like this:

```
    open_ports: list[str] = dataclasses.field(default_factory=list)
```

Which we use a lot (especially in `statustypes.py`). This is perfectly good code, and we can't use `default_factory=list[str]` in the runtime code anyway, so just silence this error.


benhoyt:
Yeah, I was wondering about that too. I don't love the noise of the ignores on all those lines, but the `# type: ignore` shortcut is not terrible. I've done that.

james-garner-canonical (approved):

PR #160, opened 2025-07-08
feat: add "overlays" parameter to Juju.deploy

### LINKED ISSUE #155: Add "overlays" arg to "deploy"
Currently `.deploy()` [doesn't take](https://canonical-jubilant.readthedocs-hosted.com/reference/jubilant/#jubilant.Juju.deploy) `overlay`.

```
> juju deploy --help | grep overlay
--overlay  (= )
    Bundles to overlay on the primary bundle, applied in order
```

It would help with updating [existing code](https://github.com/canonical/cos-lite-bundle/blob/0d7a98adb8314deea744bb8710dd3cc3ed70fae6/tests/integration/helpers.py#L42) to use jubilant.

I'd imagine it would be `overlays: Optional[List[Path | dict]] = None`.

benhoyt:
This PR adds an `overlays: Iterable[str | pathlib.Path]` parameter to `Juju.deploy` to give access to the `juju deploy --overlay` argument.

I went back and forth about whether this should be named `overlay` and also allow a single overlay, with type `str | pathlib.Path | Iterable[str, pathlib.Path]`. However, I decided that the intent of this is a list applied in order (and the example @sed-i linked to in #155 uses a list). Note "Bundles" and "applied in order" in the Juju help output:

```
$ juju deploy --help | grep -i overlay
--overlay  (= )
    Bundles to overlay on the primary bundle, applied in order
```

Fixes #155.

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -328,6 +328,7 @@ class Juju:
         constraints: Mapping[str, str] | None = None,
         force: bool = False,
         num_units: int = 1,
+        overlays: Iterable[str | pathlib.Path] = (),
         resources: Mapping[str, str] | None = None,
         revision: int | None = None,
         storage: Mapping[str, str] | None = None,
@@ -352,6 +353,7 @@ class Juju:
             constraints: Hardware constraints for new machines, for example, ``{'mem': '8G'}``.
             force: If true, bypass checks such as supported bases.
             num_units: Number of units to deploy for principal charms.
+            overlays: File paths of bundles to overlay on the primary bundle, applied in order.
             resources: Specify named resources to use for deployment, for example:
                 ``{'bin': '/path/to/some/binary'}``.
             revision: Charmhub revision number to deploy.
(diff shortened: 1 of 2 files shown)


james-garner-canonical (approved):
Looks good to me. Your reasoning for `overlays: Iterable[...]` over `overlay: str | ...` makes sense to me, and I think it fits the overall API decisions (see below). Test coverage is light but probably sufficient given that most real errors are going to come from outside `jubilant` via Juju (e.g. using overlays with a charm instead of a bundle).

---

I recall the `str | Iterable[str]` question coming up a few times, so I had a quick look through the API to see how we've handled it elsewhere.

It seems that we've most commonly explicitly allowed `str`  (5 methods) rather than raising a `TypeError` (2 methods) -- instances listed below.

However we've been very consistent in the cases where we allow `str` are exactly the cases where the Juju CLI takes `--flag a,b`, so that could be a good rule of thumb for when to use that pattern -- and I see that this method instead uses `--flag a --flag b`.

---

```py
if isinstance(..., str):
    ... = ','.join(...)
```
```py
str | Iterable[str] | None = None
```
- `add_unit` (`attach_storage`, `to`)
- `integrate` (`via`)
```py
str | Iterable[str] = ()
```
- `config` (`reset`)
- `model_config` (`reset`)
```py
arg: str | Iterable[str]
```
- `offer` (`endpoint`)

---

```py
Iterable[str] = ()
if isinstance(..., str):
    raise TypeError(...)
```
- `scp` (`scp_options`)
- `ssh` (`ssh_options`)

PR #161, opened 2025-07-08
feat: add AppStatus.leader_unit property

### LINKED ISSUE #128: Consider adding Status.get_unit method
literally all of our tests reimplement various versions of:

```
def get_unit_ip_address(juju: Juju, app_name: str, unit_no: int):
    """Return a juju unit's IP address."""
    return juju.status().apps[app_name].units[f"{app_name}/{unit_no}"].address
```

would it make sense to add it to `Status`? or as a standalone helper?

benhoyt:
This simplifies code like:

```
address = status.apps['snappass-test'].units['snappass-test/0'].address
```

to

```
address = status.apps['snappass-test'].leader_unit.address
```

Also update integration tests to use the new property.

Fixes #128

### DIFF
--- a/jubilant/statustypes.py
+++ b/jubilant/statustypes.py
@@ -254,6 +254,26 @@ class AppStatus:
         """Report whether the application status for this app is "waiting"."""
         return self.app_status.current == 'waiting'
 
+    @property
+    def leader_unit(self) -> UnitStatus:
+        """Return the leader unit's status.
+
+        For example, instead of this kind of duplication::
+
+            address = status.apps['snappass-test'].units['snappass-test/0'].address
+
+        You can simplify using ``leader_unit``:
+
+            address = status.apps['snappass-test'].leader_unit.address
+
+        Raises:
+            ValueError: if this application has no units marked as leader.
+        """
+        for unit in self.units.values():
+            if unit.leader:
+                return unit
+        raise ValueError('application has no leader unit')
+
 
 @dataclasses.dataclass(frozen=True)
 class EntityStatus:
--- a/jubilant/statustypes.py
+++ b/jubilant/statustypes.py
@@ -258,11 +258,11 @@ class AppStatus:
     def leader_unit(self) -> UnitStatus:
         """Return the leader unit's status.
 
-        For example, instead of this kind of duplication::
+        For example, instead of duplicating the application name::
 
             address = status.apps['snappass-test'].units['snappass-test/0'].address
 
-        You can simplify using ``leader_unit``:
+        You can simplify to::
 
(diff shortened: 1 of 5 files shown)


PietroPasotti:
`unit0` doesn't quite cut it because we have no guarantee that unit 0 will be elected. 
I wonder if leader_unit should error out if there is no leader yet, or if it should give None. The user could add a `jubilant.wait` waiting for it to return something truthy.

jubilant/statustypes.py
```diff
@@ -254,6 +254,26 @@ def is_waiting(self) -> bool:
         """Report whether the application status for this app is "waiting"."""
         return self.app_status.current == 'waiting'
 
+    @property
+    def leader_unit(self) -> UnitStatus:
+        """Return the leader unit's status.
+
+        For example, instead of duplicating the application name::
```
tonyandrewmeyer, line 261: I feel like "unit 0 may not be leader" is a more important reason to use this property than avoiding a little bit of duplication. Could we adjust the help to mention it?

jubilant/statustypes.py
```diff
@@ -254,6 +254,26 @@ def is_waiting(self) -> bool:
...
+
+            address = status.apps['snappass-test'].units['snappass-test/0'].address
+
+        You can simplify to::
+
+            address = status.apps['snappass-test'].leader_unit.address
+
+        Raises:
+            ValueError: if this application has no units marked as leader.
```
tonyandrewmeyer, line 270: I feel like this should be:

```suggestion
            ValueError: if this application has no unit marked as leader.
```

Since there can only be 0 or 1 leaders.

tonyandrewmeyer (approved):

jubilant/statustypes.py
```diff
@@ -254,6 +254,26 @@ def is_waiting(self) -> bool:
...
+        """
+        for unit in self.units.values():
+            if unit.leader:
+                return unit
+        raise ValueError('application has no leader unit')
```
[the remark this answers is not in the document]
dimaqq: Another possibility is that the application doesn't have a single unit.
It's apparently possible to scale an app down to 0.
Or transiently where new application got registered but no units were spooled up yet.

benhoyt:
I think Tony's right that the type of `leader_unit` should be `UnitStatus | None`. However, that's much more awkward and would just make tests longer that wait on things like `status.app['foo'].units['foo/0']`. So after further thought I don't think this pays for itself.

I'd consider adding the `leader_unit: UnitStatus | None` property later if there's a bunch of tests that show it's clearly useful.

PR #162, opened 2025-07-08
docs: add an example of temp_model, including how to set attributes

### LINKED ISSUE #134: Document how to set attributes on Juju instance returned by temp_model
Given that users would call either `temp_model()` or `Juju()`, perhaps it makes sense to replicate the keyword arguments.

benhoyt:
Fixes #134

### DIFF
--- a/jubilant/_test_helpers.py
+++ b/jubilant/_test_helpers.py
@@ -16,6 +16,16 @@ def temp_model(keep: bool = False, controller: str | None = None) -> Generator[J
 
     Provides a :class:`Juju` instance to operate on.
 
+    If you want to configure ``Juju()`` parameters such as ``wait_timeout``, set the appropriate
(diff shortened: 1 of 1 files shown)


dimaqq (approved):

jubilant/_test_helpers.py
```diff
@@ -16,6 +16,16 @@ def temp_model(keep: bool = False, controller: str | None = None) -> Generator[J
 
     Provides a :class:`Juju` instance to operate on.
 
+    If you want to configure ``Juju()`` parameters such as ``wait_timeout``, set the appropriate
```
dimaqq, line 19: There's a reference to the object Juju right above, so maybe this line can be simplified to:

```suggestion
    If you want to configure its parameters, such as ``wait_timeout``, set the appropriate
```

dimaqq (approved):

PR #164, opened 2025-07-10
feat: add consume command

### LINKED ISSUE #129: Add `consume` command
Jubilant models the `offer` command but not `consume`.  I suspect that if anyone is using `offer` in their tests, they'll want to consume that too :D It would be a nice addition

benhoyt:
This PR adds [`Juju.consume`](https://canonical-jubilant--164.com.readthedocs.build/reference/jubilant/#jubilant.Juju.consume).

I've tried to make the first argument mirror `juju.offer('mymodel.mysql')`, so it includes the model name. There are other ways we could skin the cat, but I think this one matches the existing `Juju.offer` best, and is reasonably Pythonic.

Fixes #129.

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -331,9 +331,9 @@ class Juju:
         """
         offer_path = model_and_app
         if owner is not None:
-            offer_path = owner + '/' + offer_path
+            offer_path = f'{owner}/{offer_path}'
         if controller is not None:
-            offer_path = controller + ':' + offer_path
+            offer_path = f'{controller}:{offer_path}'
         args = ['consume', offer_path]
         if alias is not None:
             args.append(alias)
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -305,6 +305,41 @@ class Juju:
 
         self.cli(*args)
 
+    def consume(
+        self,
+        model_and_app: str,
+        alias: str | None = None,
+        *,
+        controller: str | None = None,
+        owner: str | None = None,
+    ) -> None:
+        """Add a remote offer to the model.
+
+        Examples::
+
+            juju.consume('othermodel.mysql', 'sql')
+            juju.consume('othermodel.mysql', controller='ctrl2', owner='admin')
+
+        Args:
+            model_and_app: Dotted application and model name to offer endpoints for, for example
+                ``othermodel.mysql``.
+            alias: A local alias for the offer, for use with :meth:`integrate`. Defaults to the
+                application name.
+            controller: Remote offer's controller. Defaults to the current controller.
+            owner: Remote model's owner. Defaults to the user that is currently logged in to the
+                controller providing the offer.
+        """
+        offer_path = model_and_app
+        if owner is not None:
+            offer_path = owner + '/' + offer_path
+        if controller is not None:
+            offer_path = controller + ':' + offer_path
+        args = ['consume', offer_path]
+        if alias is not None:
+            args.append(alias)
+
+        self.cli(*args)
+
     def debug_log(self, *, limit: int = 0) -> str:
         """Return debug log messages from a model.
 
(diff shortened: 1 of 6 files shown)


jubilant/_juju.py
```diff
@@ -305,6 +305,41 @@ def config(
 
         self.cli(*args)
 
+    def consume(
+        self,
+        model_and_app: str,
```
james-garner-canonical, line 310: I think this is probably the right call.

My first alternative thought was `model: str, app: str, *`, since it appears we always have to specify both.

I still quite like the two required arguments approach, but I think the main argument against it is that it seems like we use the `model_and_app` dotted name form elsewhere (at least (optionally) in `offer`), and consistency is best here.

It also seems like the single string offers some extra flexibility (e.g. passing in `owner` and `controller` in the string directly just works), and easier extensibility in future.

jubilant/_juju.py
```diff
@@ -305,6 +305,41 @@ def config(
...
+                application name.
+            controller: Remote offer's controller. Defaults to the current controller.
+            owner: Remote model's owner. Defaults to the user that is currently logged in to the
+                controller providing the offer.
+        """
+        offer_path = model_and_app
+        if owner is not None:
+            offer_path = owner + '/' + offer_path
+        if controller is not None:
+            offer_path = controller + ':' + offer_path
```
james-garner-canonical, line 336: We could handle empty strings gracefully (ignoring them and treating them as the option not being specified) by testing for truthiness, avoiding sending a (presumably?) malformed string to Juju at runtime. On the other hand, maybe this has the potential to silently hide user errors.
```suggestion
        if owner:
            offer_path = owner + '/' + offer_path
        if controller:
            offer_path = controller + ':' + offer_path
```

tests/unit/test_consume.py
```diff
@@ -0,0 +1,31 @@
...
+    juju = jubilant.Juju()
+
+    juju.consume('othermodel.mysql', owner='owner')
+
+
+def test_all_args(run: mocks.Run):
+    run.handle(['juju', 'consume', 'anothercontroller:admin/othermodel.mysql', 'sql'])
+    juju = jubilant.Juju()
+
+    juju.consume('othermodel.mysql', 'sql', controller='anothercontroller', owner='admin')
```
james-garner-canonical, line 31: Suggestion: add a unit test for `controller='', owner=''` to make it harder to change the user facing behaviour by accident in future.

Suggestion: add a unit test for `juju.consume('anothercontroller:admin/othermodel.mysql')` for the same reason.

tests/unit/test_offer.py
```diff
@@ -10,14 +10,6 @@ def test(run: mocks.Run):
...
 
 
-def test_with_model(run: mocks.Run):
-    # "juju offer" isn't a model-based command
-    run.handle(['juju', 'offer', 'mysql:db'])
-    juju = jubilant.Juju(model='mdl')
-
-    juju.offer('mysql', endpoint='db')
-
-
```
james-garner-canonical, line 20: Why are we removing this test?

james-garner-canonical (approved):
Looks good. I think the `model_and_app: str` is a good call. I've made suggestions for a small behaviour change (handling `''` for `owner` and `controller`) and a couple of extra unit test cases. I'm also curious why `tests/unit/test_offer:test_with_model` is being removed.

jubilant/_juju.py
benhoyt: Yeah, it is a bit awkward, but the `model:app` syntax is to mirror what you'd write in the associated `offer()` call. See the integration tests, where the `offer()` arg matches the `consume()` arg.

benhoyt: I think I'm going to keep this as is: I think that `consume()` matching the associate `offer()` command is a pretty strong argument.

tests/unit/test_offer.py
benhoyt: I removed it as the other commands that don't take `--model` don't have this negative test. However, on reflection I think it's worth having, so I reinstated it.

jubilant/_juju.py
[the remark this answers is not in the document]
benhoyt: Done.

PR #168, opened 2025-07-18
feat: have temp_model forward "cloud" and other args to add_model

crypticC0der:
Allows users to pass the cloud option to jubilant.temp_model for instances with multiple clouds in the same controller, ie testing machine charms integrating with k8s charms.

### DIFF
--- a/jubilant/_test_helpers.py
+++ b/jubilant/_test_helpers.py
@@ -2,13 +2,20 @@ from __future__ import annotations
 
 import contextlib
 import secrets
+from collections.abc import Mapping
 from typing import Generator
 
-from ._juju import Juju
+from ._juju import ConfigValue, Juju
 
 
 @contextlib.contextmanager
-def temp_model(keep: bool = False, controller: str | None = None) -> Generator[Juju]:
+def temp_model(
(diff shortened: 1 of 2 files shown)


tonyandrewmeyer:
@benhoyt is there any reason to not have `temp_model` accept and pass through all the `add_model` arguments? Is your preference to keep it minimal because it's trivial to create a context manager just like `temp_model` but that has the extra args, and you'd rather have charms do that themselves?

tonyandrewmeyer:
> However, if we use `**kwargs` we don't get good typing/autocomplete, right? So possibly better to list them out manually (but document that args other than _keep_ are passed through)?

Yes, agreed. But then we might as well expand this PR to include `config` and `credential`. @crypticC0der would you be willing to do that? We also need to update the [tests](https://github.com/canonical/jubilant/blob/58a98f284171479d1f0ea29e4c843368c48e1cd0/tests/unit/test_test_helpers.py#L43) for the new options.

benhoyt (approved):

benhoyt:
Thanks for the contribution. Merging!

PR #171, opened 2025-07-23
chore: use mock_file for the run test

kayra1:
This PR follows up with the previous one where we introduced a mock for `NamedTemporaryFile`. It uses that mock in the test for `run`. It also includes a tiny refactor to the function to avoid using `os.remove()`.


jubilant/_juju.py
```diff
@@ -846,14 +847,12 @@ def run(
         if wait is not None:
             args.extend(['--wait', f'{wait}s'])
 
-        params_file = None
-        if params is not None:
-            with tempfile.NamedTemporaryFile(
-                'w+', delete=False, dir=self._temp_dir
-            ) as params_file:
+        with (
+            tempfile.NamedTemporaryFile('w+', dir=self._temp_dir) if params else nullcontext()
```
kayra1, line 851: it may look a bit better with `params is not None` instead.

jubilant/_juju.py
```diff
@@ -11,6 +11,7 @@
 import tempfile
 import time
 from collections.abc import Callable, Iterable, Mapping
+from contextlib import nullcontext
```
benhoyt, line 14: Our [style](https://github.com/canonical/operator/blob/main/STYLE.md#import-modules-not-objects) (apart from typing, which includes collections.abc) is to `import module` rather than `from module import thing`.
```suggestion
import contextlib
```

jubilant/_juju.py
```diff
@@ -846,14 +847,15 @@ def run(
...
-        if params is not None:
-            with tempfile.NamedTemporaryFile(
-                'w+', delete=False, dir=self._temp_dir
-            ) as params_file:
+        with (
+            tempfile.NamedTemporaryFile('w+', dir=self._temp_dir)
+            if params is not None
+            else nullcontext()
+        ) as params_file:
+            if params_file is not None:
```
benhoyt, line 855: This is fine, but it wasn't immediately clear to me (before reading the `nullcontext` docs) that `params_file` would be None here. Can we change this to `params is not None` instead?

Nice refactoring/simplification, by the way.

tests/unit/test_run.py
```diff
@@ -155,8 +151,7 @@ def test_wait_timeout(run: mocks.Run):
         juju.run('mysql/0', 'do-thing', wait=0.001)
 
 
-@pytest.mark.parametrize('cli_binary', ['/snap/bin/juju', '/bin/juju'])
-def test_params(monkeypatch: pytest.MonkeyPatch, cli_binary: str):
+def test_params(run: mocks.Run, mock_file: mocks.NamedTemporaryFile):
```
benhoyt, line 154: Due to this change, we've lost our 100% test coverage (see `make coverage-html`) for `_juju.py`, due to I'm usually not a fan of chasing coverage numbers, but going from 100% to <100% feels bad. :-) Can we add a test to `test_juju_class.py` of the `Juju._temp_dir` property directly, for the two cases?

tests/unit/test_run.py
```diff
@@ -199,5 +178,5 @@ def mock_run(args: list[str], **_: Any) -> subprocess.CompletedProcess[str]:
         results={'username': 'user', 'password': 'pass'},
     )
     assert task.success
-    assert params_path is not None
-    assert not os.path.exists(params_path)
+    assert yaml.safe_load(mock_file.writes[0]) == {'foo': 1, 'bar': ['ab', 'cd']}
```
benhoyt, line 181: Can we tweak this to:
```suggestion
    assert yaml.safe_load('\n'.join(mock_file.writes)) == {'foo': 1, 'bar': ['ab', 'cd']}
```

benhoyt (commented):
Great, thanks for this. Looks good. Couple of small things suggested.

jubilant/_juju.py
kayra1: since the nullcontext assigns an optional `None` to the `params_file` variable, I would have to add an ignore directive or assert statement here to avoid the type complaint. I could also just do `params is not None and params_file is not None`. Which one would you prefer?

jubilant/__init__.py
```diff
@@ -47,4 +47,4 @@
     'temp_model',
 ]
 
-__version__ = '1.3.0'
+__version__ = '1.3.1'
```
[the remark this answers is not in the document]
kayra1: I think that might be the best way to do versioning ever. done.

jubilant/_juju.py
benhoyt: Ah, interesting. In that case, let's just keep the code you have now, but add a comment on the `if params_file is not None:` line, something like `# not None when params is not None`.

tests/unit/test_run.py
kayra1: I'm not sure I understand what you mean correctly, but I added the two options as tests. Let me know if I got it wrong.

jubilant/_juju.py
dimaqq: Alternatively, WDYT about using a temporary directory?
- directory is created always, whether a file is needed or not
- 0 or more files are created in it and 0 or more file are used
- directory is removed after

Example: https://github.com/canonical/operator/pull/1290/files

benhoyt: Yeah, that's not a bad approach, though I'd prefer to stick with the current approach for now (nothing created unless needed).

tests/unit/test_juju_class.py
```diff
@@ -44,3 +46,17 @@ def test_method_order():
     sorted_by_alpha = sorted(method_linenos)
     sorted_by_lines = sorted(method_linenos, key=lambda k: method_linenos[k])
     assert sorted_by_lines == sorted_by_alpha, 'Please keep Juju methods in alphabetical order'
+
+
+def test_default_tempdir(monkeypatch: pytest.MonkeyPatch):
+    monkeypatch.setattr('jubilant.Juju._juju_is_snap', False)
```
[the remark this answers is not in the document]
kayra1: default tempdir would be not snap, and that would not include 'snap' in the directory, is that not correct?

tests/unit/test_juju_class.py
```diff
@@ -44,3 +46,17 @@ def test_method_order():
...
+    juju = jubilant.Juju(cli_binary='/bin/juju')
+
+    assert 'snap' not in juju._temp_dir
```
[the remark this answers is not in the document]
kayra1: default tempdir would be not snap, and that would not include 'snap' in the directory, is that not correct?

benhoyt (approved):
Excellent, thanks for this.

PR #173, opened 2025-07-29
docs: Add a security explanation doc

tonyandrewmeyer:
Adds a basic Security doc in the explanation section.

This docs covers:

* Cryptography (non) use
* Hardening (not required)
* Getting security updates
* Security risks introduced/changed when using Jubilant (none)
* Good security practices for charm test authors

The goal is for this doc to be generally useful for charm authors, but also to comply with [SEC0030](https://docs.google.com/document/d/12_hj3_xhaYWqKHsmnuRl5wQO3Es4GuHWtgrxMHG21FM/edit?tab=t.0#heading=h.jszeo1fa8onu) and [Canonical product security documentation guidance](https://docs.google.com/document/d/1QBVUxfOVaTXEbBUiiWMzYHiAxG79NIGS7YTVX_GNE4Q/edit?tab=t.0#heading=h.ikq8mvhgar8v).

### DIFF
--- a/docs/explanation/index.md
+++ b/docs/explanation/index.md
@@ -2,14 +2,10 @@
 
 These guides explain how Jubilant works and how it's developed.
 
-
-## Design goals
-
-When designing Jubilant, we had several important goals.
-
 ```{toctree}
 :titlesonly:
 :maxdepth: 1
 
 Design goals <design-goals>
+Security <security>
 ```
--- a/docs/explanation/security.md
+++ b/docs/explanation/security.md
@@ -1,0 +1,49 @@
+(security)=
+
+# Security
+
+When producing security documentation for your charm, it's important to consider the security aspects of the charm's tests. If there are any security questions that aren't answered here in terms of the use of cryptographic technology, security risks, hardening, and good practice, with regards to Jubilant, please reach out to the Charm Tech team, and we'll do our best to assist.
+
+## Cryptographic technology
+
+Jubilant does not use any cryptographic technology, hashing, or digital signatures.
+
+## Inter-process communication
+
+Jubilant communicates with Juju by running processes in the test environment (the Juju CLI).
+
+> See also:
+> - {external+juju:ref}`Juju CLI <list-of-juju-cli-commands>`
+
+## Hardening
+
+No additional steps are required to harden your integration tests when using Jubilant.
+
+> See also: {external+juju:ref}`Juju | Harden your deployment <harden-your-deployment>`
+
+## Security updates
+
+We strongly recommend restricting the version of `jubilant` in `pyproject.toml` in a way that allows picking up new compatible releases every time that you re-lock. For example, `jubilant~=1.2`. Set a minor version that includes all the features that the charm tests use.
+
+Your charm repository should have tooling configured so that any dependencies with security updates are detected automatically (such as [Dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/about-dependabot-security-updates) or [Renovate](https://www.mend.io/renovate/)), prompting to you re-lock so that the charm will be built with the latest version.
+
+For information about supported versions and how to report security issues, see [SECURITY.md](https://github.com/canonical/jubilant/blob/main/SECURITY.md).
+
+## Risks
+
+Jubilant inherits the risks of using the Juju CLI. Charm authors should be familiar with {external+juju:ref}`Juju security <juju-security>`.
+
+Charm integration tests are most commonly run in ephemeral CI environments. However, the tests may also be run in local development environments. Charm authors should be mindful of this when writing integration tests.
+
+Otherwise, Jubilant doesn't introduce any new security risks over simply running Juju CLI commands.
+
+## Good practices
+
+* Where charm tests require cloud credentials, these should be saved in an appropriate secret store (such as GitHub secrets), should be used only for the integration tests, and should provide no access to production clouds.
+* Only `deploy(trust=True)`, `refresh(trust=True)`, and `trust(remove=False)` in tests when required by the relevant charm, and ensure that the cloud credentials that the charm will gain access to is only used for integration tests.
+* When using `scp()` use a secure temporary directory for the local side.
+* Charms should follow best practices for writing secure Python tests.
+* Charms should have workflows that statically check for security issues (such as [ruff](https://docs.astral.sh/ruff/linter/) and [zizmor](https://docs.zizmor.sh/)).
+* Charm authors should exercise caution when considering adding dependencies to their charm tests.
+* Write the exact dependencies of the charm's tests into a lock file (using `uv lock`, `poetry lock`, or similar tool) and commit that lock file to source control.
+* Charm repositories should have tooling that automatically detects outdated dependencies, particularly missing security updates.


docs/explanation/index.md
```diff
@@ -2,14 +2,10 @@
 
 These guides explain how Jubilant works and how it's developed.
 
-
-## Design goals
-
-When designing Jubilant, we had several important goals.
-
```
tonyandrewmeyer, line 9: I couldn't figure out a way to keep this wording but also have a document that wasn't "design goals" in the table of contents, and I don't think we want multiple ToCs on the page.

I'm happy to put this back if someone can suggest a way that it can work with other explanation docs being included as well.

benhoyt (approved):
Looks good overall, but left a few minor comments.

docs/explanation/security.md
```diff
@@ -0,0 +1,48 @@
...
+Jubilant communicates with Juju by running processes in the test environment (the Juju CLI).
+
+> See also:
+> - {external+juju:ref}`Juju CLI <list-of-juju-cli-commands>`
+
+## Hardening
+
+No additional steps are required to harden your integration tests when using Jubilant.
+
+> See also: {external+juju:ref}`Juju | Harden your deployment <harden-your-deployment>`
```
dwilding, line 22: We'll need to add Juju as an intersphinx project (in docs/conf.py). Would it help if I push that to your branch?

docs/conf.py
```diff
@@ -202,8 +202,9 @@
 # TODO: [@dwilding DONE] Remove or adjust the ACME entry after you update the contributing guide
 
 linkcheck_ignore = [
-    "http://127.0.0.1:8000"
-    ]
+    "http://127.0.0.1:8000",
+    "https://documentation.ubuntu.com/juju/3.6/reference/juju-cli/list-of-juju-cli-commands/#list-of-juju-cli-commands",  # Actual anchor is #id1 because title contains code font
```
dwilding, line 206: The intersphinx ref to `list-of-juju-cli-commands` causes a problem for the link checker. The [target page](https://documentation.ubuntu.com/juju/3.6/reference/juju-cli/list-of-juju-cli-commands/) has a title that includes code font, so the expected anchor `#list-of-juju-cli-commands` doesn't exist, which causes the link checker to fail. I'm excluding the URL for now and checking whether there's a better solution.

tonyandrewmeyer: By the way, the page I'm trying to link to is [this one](https://documentation.ubuntu.com/juju/3.6/reference/juju-cli/). Maybe that's the wrong reference anyway? I don't see anything else in [here](https://github.com/juju/juju/tree/3.6/docs/reference/juju-cli) though, and I think there's some automation that provides the content?

tonyandrewmeyer: Ah, thanks! Not sure how I missed that - I thought I scanned through the files in the parent as well as looking at the folder, but not carefully enough it seems!

docs/explanation/security.md
```diff
@@ -0,0 +1,48 @@
...
+
+Your charm repository should have tooling configured so that any dependencies with security updates are detected automatically (such as [Dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/about-dependabot-security-updates) or [Renovate](https://www.mend.io/renovate/)), prompting to you re-lock so that the charm tests will run with the latest version.
+
+For information about supported versions and how to report security issues, see [SECURITY.md](https://github.com/canonical/jubilant/blob/main/SECURITY.md).
+
+## Risks
+
+Jubilant inherits the risks of using the Juju CLI. Charm authors should be familiar with {external+juju:ref}`Juju security <juju-security>`.
+
+Charm integration tests are most commonly run in ephemeral CI environments. However, the tests may also be run in local development environments. Charm authors should be mindful of this when writing integration tests.
```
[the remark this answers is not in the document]
tonyandrewmeyer: The point I'm trying to make is that when you're filling out your risk register for your charm (or doing threat modelling, or whatever the security process looks like at your organisation), you need to remember that if you are running tests locally, then the local system is at risk: running tests that use Jubilant is giving Jubilant the same access to your system as the user running the tests. A malicious or buggy version of Jubilant can do very bad things to your local system.

There's also a bit of a point about not doing things like installing software or changing `/etc/hosts` or other things like that, which I've seen tests do, presumably with the thought that they are running in ephemeral environments and so can do anything they like, but which are actually quite invasive on a local development system.

It's true that this is more about *running tests* in general than something that Jubilant introduces -- or, assuming that you/your team wrote the tests, then having dependencies (Jubilant or otherwise) for running tests, because *any* of those dependencies could be compromised. Perhaps it's more suited to the Ops security doc, or an Ops testing doc. I'm fine with removing it if you think that's better.

docs/explanation/security.md
```diff
@@ -0,0 +1,48 @@
+(security)=
+
+# Security
+
+When producing security documentation for your charm, it's important to consider the security aspects of the charm's tests. If you have any Jubilant-related security questions that aren't answered here in terms of the use of cryptographic technology, security risks, hardening, and good practice, please reach out to the Charm Tech team, and we'll do our best to help.
```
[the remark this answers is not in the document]
tonyandrewmeyer: I added an intro section like this to the Ops one (and then copied & adjusted it here) because Dima wanted the doc to explain why it was useful for charmers. So it was somewhat of a "here's what you can find and why you want to find it", that I'm not a huge fan of but added to try to address a review concern.

Removing the list is fine with me - the contents section really gives it to you anyway.

docs/explanation/index.md
tonyandrewmeyer: Thanks! How's this?

docs/explanation/security.md
```diff
@@ -0,0 +1,48 @@
...
+
+Otherwise, Jubilant doesn't introduce any new security risks over directly running Juju CLI commands.
+
+## Good practices
+
+* Where charm tests require cloud credentials, these should be saved in an appropriate secret store (such as GitHub secrets), should be used only for the integration tests, and should provide no access to production clouds.
+* Only use `deploy(trust=True)`, `refresh(trust=True)`, and `trust(remove=False)` in tests when required by the relevant charm. Ensure that the cloud credentials that the charm will gain access to are only used for integration tests.
+* When using `scp()` use a secure temporary directory for the local side.
+* Charms should follow best practices for writing secure Python tests.
+* Charms should have workflows that statically check for security issues (such as [ruff](https://docs.astral.sh/ruff/linter/) and [zizmor](https://docs.zizmor.sh/)).
```
dwilding, line 46:
```suggestion
* Charms should have workflows that statically check for security issues (such as [ruff](https://docs.astral.sh/ruff/linter/) and [zizmor](https://docs.zizmor.sh/)).
* Charms should follow best practices for writing secure Python tests.
* Charm tests may be run in local development environments, so charm tests should not install software or make system changes.
```

docs/explanation/security.md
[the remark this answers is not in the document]
dwilding: OK, I get you, thanks.

I think it's worth keeping, especially because of the "Otherwise" paragraph that follows. How about reducing it to:

```
Charm integration tests are most commonly run in ephemeral CI environments. However, the tests may also be run in local development environments, giving Jubilant access to these environments.
```

Then adding a best practices bullet point too. See https://github.com/canonical/jubilant/pull/173/files#r2241563045

I agree that the general advice would be good to cover in the Ops docs too.

PR #177, opened 2025-08-06
docs: add sitemap.xml, add new spelling check, remove markdown lint

dwilding:
This PR updates to the latest version of the [doc starter pack](https://github.com/canonical/sphinx-docs-starter-pack), customising to what I think is practical for Jubilant.

Overview of changes:

- Enabled sitemap.xml ([preview](https://canonical-ubuntu-documentation-library--177.com.readthedocs.build/jubilant/sitemap.xml)). This is a requirement for docs under documentation.ubuntu.com.

- Added the new spell checker to our docs CI. We had previously decided not to use the spell checker, as it required lots of exclusions, but the new spell checker seems better.

- Removed the markdown linter from our docs CI. The linter setup has changed in the latest starter pack, and now runs against the .md doc source files, raising issues that I don't think affect how the docs are rendered.

    Previously, we only used the linter for the "user-facing" markdown in the repo root, and by default the new linter doesn't check that. After exploring the configuration in [pebble#659](https://github.com/canonical/pebble/pull/659), I don't think it's worth the trouble of enabling the linter in CI. We can still run it manually with `cd docs; make lint-md`.

This PR includes other less impactful changes. The latest version of the starter pack has moved some files around and added/updated some scripts. That means some extra noise in the PR. I could get more aggressive with removing files that we probably won't use, but I'd rather not break or edit the docs Makefile (since we have no particular reason to customise it).

I've added comments to highlight the most significant changes.


benhoyt:
Why is https://documentation.ubuntu.com/jubilant/404/ in the sitemap? I notice URLs like this are in the Ops and Pebble sitemaps too, for example https://documentation.ubuntu.com/pebble/404/ -- which is actually an HTTP 200. I wonder if we can update the starter pack (and our sites) to exclude this URL from the sitemap. Ideally going to those URLs would return HTTP 404 too, not 200 -- though that's probably not a big deal.

dwilding:
Good point about the 404 page. It turns out that we can exclude URLs from the sitemap as part of the config that I've added in this PR. So I'm excluding these generated pages:

```
'404/',
'genindex/',
'py-modindex/',
'search/',
```

It would be nice if these were excluded automatically, but I think this is good for now. (I'll need to update the config of our other sites anyway, so I'll include the same exclusions when I do that.)

Re status codes: Yeah, interesting that https://documentation.ubuntu.com/jubilant/404/ returns 200. I suppose that's because it's a page that does actually exist. Other really-not-found pages return 404 as expected - [example](https://documentation.ubuntu.com/jubilant/foo/)

Issue #141, opened 2025-05-22
Add `bootstrap` command


benhoyt:
This won't be used in most integration tests, but it could be used in certain types of tests, and in addition, we've had requests of it when using Jubilant in other contexts. For example, Sunbeam's governor snap that orchestrates the rest of the world (per discussion with Eric Chen and Billy Olsen).

PR #180, opened 2025-08-08
feat: add bootstrap command

IronCore864:
Add Juju bootstrap with unit test.

[Preview of API Reference docs.](https://canonical-ubuntu-documentation-library--180.com.readthedocs.build/jubilant/reference/jubilant/#jubilant.Juju.bootstrap)

The integration test is not added because it seems unnecessary, and it is also too slow.

Closes https://github.com/canonical/jubilant/issues/141.


jubilant/_juju.py
```diff
@@ -205,6 +205,151 @@ def add_unit(
...
+        db_snap_asserts: str | pathlib.Path | None = None,
+        force: bool = False,
+        keep_broken: bool = False,
+        metadata_source: str | pathlib.Path | None = None,
+        model_default: Mapping[str, ConfigValue] | None = None,
+        no_switch: bool = False,
+        regions: str | None = None,
+        storage_pool: Mapping[str, str] | None = None,
+        to: str | None = None,
+    ) -> str:
```
benhoyt, line 237: I think we won't need many of these arguments, and we should trim it down a bit. I suggest getting rid of the ones having to do with customising/developing the controller code:

- agent_version
- auto_upgrade
- bootstrap_image
- build_agent 
- controller_charm_channel
- controller_charm_path
- db_snap
- db_snap_asserts
- metadata_source

In addition, I think we should drop these:

- clouds (this is a completely different mode, and should be done via `juju clouds --all` or similar instead; let's just drop it here)
- regions (similar to *clouds*)
- keep_broken: I don't think this one will be useful in an automated context

jubilant/_juju.py
```diff
@@ -205,6 +205,151 @@ def add_unit(
...
+        controller_charm_channel: str | None = None,
+        controller_charm_path: str | pathlib.Path | None = None,
+        credential: str | None = None,
+        db_snap: str | pathlib.Path | None = None,
+        db_snap_asserts: str | pathlib.Path | None = None,
+        force: bool = False,
+        keep_broken: bool = False,
+        metadata_source: str | pathlib.Path | None = None,
+        model_default: Mapping[str, ConfigValue] | None = None,
+        no_switch: bool = False,
```
benhoyt, line 233: We'll have to think about what we want to do with `no_switch`. For the similar arg in `add_model`, we always add `--no-switch`, with this reasoning:

> To avoid interfering with CLI users, this won't switch the Juju CLI to the
> newly-created model. However, because :attr:`model` is set to the name of the new
> model, all subsequent operations on this instance will use the new model.

So I think we should try to do the same here. However, it's a bit tricky -- I guess we should set `self.model` to the `f'{controller}:'` (note the trailing colon indicating no model name) or `f'{controller}:{add_model}'` if *add_model* is provided. However, what do we do if *controller* isn't provided? Try to parse it from the stdout/stderr? (A bit yucky.) Or always require both cloud and controller name?

In addition, we may have to update handling of `self.model` in other functions to maintain the controller name prefix.

Adding @tonyandrewmeyer, given that he worked on [multiple controller support](https://github.com/canonical/jubilant/pull/165) -- thoughts?

tests/unit/test_bootstrap.py
```diff
@@ -0,0 +1,118 @@
+import jubilant
+
+from . import mocks
+
+
+def test_defaults_interactive(run: mocks.Run):
```
benhoyt, line 6: Per above, we should drop interactive support. So this would become just `test_defaults` and be something like `juju.bootstrap('microk8s')`.

tests/unit/test_bootstrap.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def test_defaults_with_cloud(run: mocks.Run):
+    run.handle(['juju', 'bootstrap', 'lxd'])
+    juju = jubilant.Juju()
+
+    juju.bootstrap('lxd')
+
+
+def test_defaults_with_add_model(run: mocks.Run):
```
benhoyt, line 34: I think this can just be handled in `test_all_args`.

jubilant/_juju.py
```diff
@@ -205,6 +205,151 @@ def add_unit(
...
+            model_default: Configuration options to set for all models, unless otherwise
+                specified.
+            no_switch: If true, do not switch to the newly created controller.
+            regions: Print the available regions for the specified cloud.
+            storage_pool: Options for an initial storage pool as key-value pairs. ``name``
+                and ``type`` are required, plus any additional attributes.
+            to: Placement directive indicating an instance to bootstrap.
+
+        Returns:
+            The standard output of the Juju CLI command.
```
benhoyt, line 289: Note that the standard output of `juju bootstrap` is empty. The stderr contains the interesting stuff (I just checked with `juju bootstrap lxd lxd2 >out 2>err`). So if anything we should return stderr.

However, I think maybe we should just return nothing, as it's human-readable and has to be parsed anything, so is not particularly useful. It may be useful for diagnostics, though, so how about we log it and return nothing?

```python
_, stderr = self._cli(*args, include_model=False)
logger.info('bootstrap output:\n%s', stderr)
```

jubilant/_juju.py
```diff
@@ -205,6 +205,151 @@ def add_unit(
...
+        db_snap: str | pathlib.Path | None = None,
+        db_snap_asserts: str | pathlib.Path | None = None,
+        force: bool = False,
+        keep_broken: bool = False,
+        metadata_source: str | pathlib.Path | None = None,
+        model_default: Mapping[str, ConfigValue] | None = None,
+        no_switch: bool = False,
+        regions: str | None = None,
+        storage_pool: Mapping[str, str] | None = None,
+        to: str | None = None,
```
benhoyt, line 236: This should probably be `str | Iterable[str] | None` like `deploy` and `add_unit`.

jubilant/_juju.py
```diff
@@ -205,6 +205,151 @@ def add_unit(
...
+            config: Controller configuration options. Model config keys only affect the
+                controller model.
+            constraints: Set model constraints, for example, ``{'mem': '8G', 'cores': '4'}``.
+                If used, its values will be set as the default constraints for all future
+                workload machines in the model, exactly as if the constraints were set with
+                ``juju set-model-constraints``.
+            controller_charm_channel: The Charmhub channel to download the controller charm
+                from (if not using a local charm).
+            controller_charm_path: Path to a locally built controller charm.
+            credential: Credentials to use when bootstrapping.
```
benhoyt, line 273: I believe this is the name of a previously added credential. So we should probably clarify, like we do in the `add_model` docstring:

```suggestion
            credential: Name of cloud credential to use when bootstrapping.
```

benhoyt (changes requested):
Looks like a good start, thanks. I think we should considerably cut back on the number of options, at least for this initial version.

jubilant/_juju.py
tonyandrewmeyer: Yeah, I agree that we want to force `--no-switch` for the same reasons as we had for `add-model`.

I think if we don't get something for `--add-model` then we shouldn't be changing the instance's model - I'm not a fan of setting the model name to be the empty string. I'm also not keen on parsing the human-readable output to find the controller name. I think requiring a name would probably be ok - if you want to have a random name, then it's not difficult to do that in Python rather than having Juju do it.

I think my preference would be that `bootstrap` doesn't alter the current instance. If you wanted to, there are a few options; one would be:

```python
juju.bootstrap(cloud="k8s", controller="mycontroller")
juju.add_model("mymodel", "k8s", controller="mycontroller")
```

It is inconsistent with the `add_model` behaviour, but `bootstrap` feels like a different kind of command anyway.

My second preference would be to only set `.model` if both the controller name and a model to create are provided. I feel like this is cleanest, and could be expanded to not require the controller name if some future version of Juju had a `--format=json` option for `bootstrap`, but it's not ideal that the behaviour differs depending on the arguments.

> In addition, we may have to update handling of self.model in other functions to maintain the controller name prefix.

This should have been done in #165. I reviewed all the commands in the Jubilant API and all of the CLI commands accept `controller:model` other than where we special-case (like `offer`). It's possible I missed something, but I feel it's a bug if `controller:model` doesn't work somewhere (other than `.cli()`).

benhoyt: After team discussion, doing this:

- make both *cloud* and *controller* required (so controller name is known)
- not have the `add_model` argument on `Juju.bootstrap` at all, but just let people use `juju.add_model()`, as per Tony's code snippet above.

PR #188, opened 2025-09-02
fix: add timeout to temp_model destroy-model to fix microk8s issues

benhoyt:
The integration-k8s tests are hanging ([example here](https://github.com/canonical/jubilant/actions/runs/17312074131/job/49297452945)) just after the second-to-last test (test_secrets.test_get_all_secrets) and then timing out after 6h. Juju hangs on cleanup -- after talking to the Juju team, this is a Microk8s issue, so work around it here.

We could only do this for microk8s, but it seems like when using a temp model, destroying it should never take longer than a reasonable time (10 minutes seems ample), so just add a timeout to the `juju destroy-model` in `temp_model()` always.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ install-dev:  # Install development tooling
 	uv tool install zizmor
 
 integration-k8s:  # Run K8s integration tests on Juju, eg: make integration ARGS='-k test_deploy'
-	uv run pytest tests/integration -vv --log-level=INFO --log-format="%(asctime)s %(levelname)s %(message)s" -m 'not machine' $(ARGS)
+	uv run pytest tests/integration -vv --log-cli --log-cli-level=INFO --log-format="%(asctime)s %(levelname)s %(message)s" -m 'not machine' $(ARGS)
 
 integration-machine:  # Run machine integration tests on Juju, eg: make integration-machine ARGS='-k test_ssh'
 	uv run pytest tests/integration -vv --log-level=INFO --log-format="%(asctime)s %(levelname)s %(message)s" -m machine $(ARGS)
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ integration-k8s:  # Run K8s integration tests on Juju, eg: make integration ARGS
 	uv run pytest tests/integration -vv --log-cli-level=INFO --log-format="%(asctime)s %(levelname)s %(message)s" -m 'not machine' $(ARGS)
 
 integration-machine:  # Run machine integration tests on Juju, eg: make integration-machine ARGS='-k test_ssh'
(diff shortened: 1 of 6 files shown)


tonyandrewmeyer:
Potentially [this bug](https://github.com/juju/juju/issues/19869) (adding so there's a link).

jubilant/_juju.py
```diff
@@ -313,6 +318,7 @@ def _cli(
                 capture_output=True,
                 encoding='utf-8',
                 input=stdin,
+                timeout=timeout,
```
tonyandrewmeyer, line 321: It feels a bit odd that we wrap `CalledProcessError` but don't wrap `TimeoutExpired`. I guess the difference is that there's no public way to provide the timeout, but it does feel a bit off.

jubilant/_test_helpers.py
```diff
@@ -49,4 +53,14 @@ def juju():
     finally:
         if not keep:
             assert juju.model is not None
-            juju.destroy_model(juju.model, destroy_storage=True, force=True)
+            try:
```
tonyandrewmeyer, line 56: I think we should have a comment here that explains why we aren't using the provided method, so people don't need to blame to understand that.

I wonder if the doc for `temp_models should also explicitly say that it'll wait up to 10m to destroy.

tonyandrewmeyer (approved):
I'm +1 on adding a timeout arg to `_cli()`. I think it would be reasonable in `cli()1s as well, but I'm fine waiting for someone to ask for that.

I agree it's better to patch the timeout into the helper than to `destroy_model`.

Makefile
```diff
@@ -29,10 +29,10 @@ install-dev:  # Install development tooling
 	uv tool install zizmor
 
 integration-k8s:  # Run K8s integration tests on Juju, eg: make integration ARGS='-k test_deploy'
-	uv run pytest tests/integration -vv --log-level=INFO --log-format="%(asctime)s %(levelname)s %(message)s" -m 'not machine' $(ARGS)
+	uv run pytest tests/integration -vv --log-cli-level=INFO --log-format="%(asctime)s %(levelname)s %(message)s" -m 'not machine' $(ARGS)
```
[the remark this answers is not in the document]
benhoyt: Yeah, I had that in there initially and it didn't work. Maybe it's only for the config file version?

jubilant/_juju.py
benhoyt: Yeah, I am happy wrapping it if/when we make `timeout` public.

jubilant/_test_helpers.py
benhoyt: Both good points - added.

PR #189, opened 2025-09-02
chore: run integration tests against Juju 4/beta

dimaqq:
- adds matrix over Juju versions
- adds Juju channel 4/beta
- adds integration test timeout
- don't fail the overall job if 4/beta branch fails

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -53,6 +53,11 @@ jobs:
       group: ${{ github.workflow }}-${{ github.ref }}-k8s
       cancel-in-progress: true
 
+    strategy:
+      fail-fast: false
+      matrix:
+        juju-channel: ['3/stable', '4/beta']
+
     steps:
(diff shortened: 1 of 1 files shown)


dimaqq:
Findings:
- can't tests against Juju 4 on machines, because storage is not implemented
- Heisenbug against Juju 3, stuck at 96%, killed by timeout
- Heisenbug against Juju 4, stuck at 7%, killed by timeout

I suspect that more tests were run than 7%, that it's just the amount of output we get.

benhoyt:
The Juju 3 Heisenbug aside, I think it's probably too early to run these against Juju 4 on every PR. They'll just fail every time, and it'll be noise.

dimaqq:
I agree.
Juju 4 vm not ready for this workload.
Juju 4 k8s could perhaps be used to figure out the tooling around Heisenbugs, but that would be a separate PR.

Issue #178, opened 2025-08-06
Add version command


benhoyt:
You can currently spell this `Juju.cli('version', include_model=False)`, but the *include_model* arg is a bit awkward, and it returns a plain string. Would be useful to wrap the `version` command and return a nice namedtuple/dataclass, something like this:

```python
class Juju:
    def version(self) -> Version: ...

@dataclasses.dataclass
class Version:
    major: int
    minor: int
    patch: int  # or perhaps "micro" like sys.version_info
    suffix: str  # not sure what form this should take, or if there's be multiple fields

    def semver_tuple(self) -> tuple[int, int, int]: ...
```

We need to determine whether this supports `--all` (or always includes `--all` information).

PR #193, opened 2025-10-02
feat: add Juju 2.9 compatibility

tonyandrewmeyer:
Adds compatibility with Juju 2.9. The API remains the same as the regular Jubilant (meaning the Juju 3.x API) with the code translating to Juju 2.9 CLI calls where required, if the Juju version is 2.9.

This is essentially the same (and replaces) https://github.com/tonyandrewmeyer/jubilant-backports, but because it's in a branch of the actual Jubilant code rather than a separate package, it's much simpler because instead of lots of subclassing there are more surgical changes.

There is one new public name, `ExecTask`, because in Juju 2.9 `exec` and `run-action` have quite different return types.

The `Juju` class loads the CLI version at instantiation time, which does mean there is one extra subprocess call per `Juju` object created, but that seems ok for this backports package (it could be moved to a lazy property if required). We are expecting to [load the version in a future release of the main package](https://github.com/canonical/jubilant/issues/178), but if that is `.version` it would not clash with this.

The major differences between 2.9 and 3.x are:
* 2.9 does not have secrets. We let the Juju "not a juju command" error surface naturally.
* `base` arguments are translated to `series` in `bootstrap()`, `deploy()`, and `refresh()`.
* `exec()` has a different return type, which means that the signature changes slightly, to `Task | ExecTask`.
* `run` works somewhat differently in Juju 2.9.

The same status classes are used for 2.9 and 3.x. They are mostly the same, and where there are differences the fields will be empty in 2.9 or 3.x. The most significant difference is in `AppStatus`, where `relations` is a `dict[str, list[AppStatusRelation]]` but in 2.9 none of the fields of the `AppStatusRelation`s are filled out. There are two new public names: `MeterStatus` and `BranchStatus`.

The existing unit tests remain generally unchanged, except that the unit tests are set to behave as if they are running under Juju 3. A new set of unit tests that explicitly test both 2.9 and 3.x where the behaviour differs is added.

The existing integration tests remain generally unchanged. A new set of integration tests that explicitly test both 2.9 and 3.x where the behaviour differs is added.

The intention is **not** to release this as a new `jubilant` package release. Instead a follow-up PR will adjust `pyproject.toml` and some documentation to release this as a new `jubilant-backports` package release. We also intend to only provide bug and security fixes for the 2.9 support: most particularly, if any new features are added to the main Jubilant releases, they will not be backported to the `jubilant-backports` branch.


tonyandrewmeyer:
@benhoyt sorry, I tagged you as a reviewer but this is still in draft until I verify the tests (I need to adjust the CI, I think). I'll ping you when it's ready for review.

jubilant/statustypes.py
```diff
@@ -174,7 +202,7 @@ class AppStatus:
     address: str = ''
     life: str = ''
     app_status: StatusInfo = dataclasses.field(default_factory=StatusInfo)
-    relations: dict[str, list[AppStatusRelation]] = dataclasses.field(default_factory=dict)  # type: ignore
+    relations: dict[str, list[AppStatusRelation | str]] = dataclasses.field(default_factory=dict)  # type: ignore
```
benhoyt, line 205: Per discussion, let's use just `list[AppStatusRelation]` but only fill in the `interface` field.

tests/unit/test_add_model.py
```diff
@@ -5,7 +5,7 @@
 
 def test_defaults(run: mocks.Run):
     run.handle(['juju', 'add-model', '--no-switch', 'new'])
-    juju = jubilant.Juju(model='initial')
+    juju = jubilant.Juju(model='initial', cli_version='3.6.9')
```
benhoyt, line 8: What if instead of `cli_version` property we have an `Juju.is_juju_3` property and default it to `True`. That would get rid of all these.

tests/integration/test_execution.py
```diff
@@ -109,9 +111,9 @@ def test_ssh_and_scp(juju: jubilant.Juju):
     output = juju.ssh('snappass-test/0', 'ls', '/charm/containers')
     assert output.split() == ['redis', 'snappass']
     output = juju.ssh('snappass-test/0', 'ls', '/charm/container', container='snappass')
-    assert 'pebble' in output.split()
+    assert 'pebble.socket' in output.split()
     output = juju.ssh('snappass-test/0', 'ls', '/charm/container', container='redis')
-    assert 'pebble' in output.split()
+    assert 'pebble.socket' in output.split()
```
tonyandrewmeyer, line 116: In Juju 3.6 there is a "pebble" directory and a "pebble.socket" (socket) file. In Juju 2.9, there are two socket files "pebble.socket" and "pebble.socket.untrusted". It seems like the simplest change here is to look for the socket file rather than the directory.

tests/unit/test_pretty.py
```diff
@@ -37,6 +37,7 @@ def test_medium():
     version='3.6.1',
     region='localhost',
     model_status=StatusInfo(current='available', since='24 Feb 2025 12:02:57+13:00'),
+    sla='unsupported',
```
tonyandrewmeyer, line 40: This is present in 2.9 but not in 3.0. For the purposes of checking the pretty functionality, it seems simplest (in this branch) to just always have a value.

jubilant/_juju.py
```diff
@@ -791,10 +839,25 @@ def refresh(
...
                 args.extend(['--storage', f'{k}={v}'])
-        if trust:
+        if trust and not self._is_juju_2:
             args.append('--trust')
 
-        self.cli(*args)
+        mgr = (
+            contextlib.nullcontext()
+            if (not self._is_juju_2 or config is None)
+            else tempfile.TemporaryFile('w')  # noqa: SIM115
```
benhoyt, line 848: Should this be `NamedTemporaryFile`? I don't think `TemporaryFile` is guaranteed to have a name. Also, I think it should probably have a `dir=self._temp_dir` argument like the other calls (due to the snap being unable to access the `/tmp` directory).

jubilant/_juju.py
```diff
@@ -1277,6 +1392,24 @@ def _temp_dir(self) -> str:
...
+        '14.04': 'trusty',
+        '16.04': 'xenial',
+        '18.04': 'bionic',
+        '20.04': 'focal',
+        '22.04': 'jammy',
+        '24.04': 'noble',
+        '24.10': 'oracular',
+        '25.04': 'plucky',
+        '25.10': 'questing',
+    }[cycle]
```
benhoyt, line 1410: Style nit: I would prefer this "constant" dict was extracted to a module-lev var to avoid building it every time. `_SERIES_NAMES` or `_SERIES_BY_BASE` or something.

benhoyt (approved):
Just two very small things, but approving optimistically. Thanks for this!

jubilant/_juju.py
tonyandrewmeyer: I believe `TemporaryFile` is guaranteed to have a name, but that name might not be visible to the file system. But that is indeed an issue here, because the separate `juju` process needs to be able to see the file.

tonyandrewmeyer:
Merging with the 3.x MicroK8s integration tests timing out because of a known issue. Still planning on having those pass before there's an actual release.

PR #195, opened 2025-10-03
docs: another wait() example, and link to it from ref and how-to

### LINKED ISSUE #151: More docs and examples of custom wait functions
When Jon was writing the integration tests for the https://github.com/canonical/ubuntu-manpages-operator charm, he found there weren't many examples of different types of custom `Juju.wait` "ready" functions. We should add more of these or make the ones in the migration how-to more prominent, or something.

This is one of the most interesting/complex parts of an integration test, so it would be good to have more here.

benhoyt:
Now that we have the tutorial (and how-to), there are actually quite a few examples of this, so this only adds one example. But it also shows how to use a stand-alone `def` function with type annotations (which can be better for autocompletion), and links to the examples from the `Juju.wait` reference docs and the migration how-to.

[**Preview**](https://canonical-ubuntu-documentation-library--195.com.readthedocs.build/jubilant/tutorial/getting-started/#use-a-custom-wait-condition)

Fixes #151.

### DIFF
--- a/docs/how-to/migrate-from-pytest-operator.md
+++ b/docs/how-to/migrate-from-pytest-operator.md
@@ -267,6 +267,8 @@ juju.wait(
 )
 ```
 
+For more examples, see [Tutorial | Use a custom wait condition](#use_a_custom_wait_condition).
+
 
 ### Integrating two applications
 
--- a/docs/tutorial/getting-started.md
+++ b/docs/tutorial/getting-started.md
@@ -93,6 +93,7 @@ def juju():
 ```
 
 
+(use_a_custom_wait_condition)=
 ## Use a custom `wait` condition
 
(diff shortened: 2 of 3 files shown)


james-garner-canonical (approved):
These examples are very helpful. I wonder if it would be worth adding one using the `is_*` convenience methods too, but doesn't need to block this PR by any means.

benhoyt:
@james-garner-canonical The `test_custom_wait` example uses `is_active` and `is_blocked`, did you see that one?

```python
def test_custom_wait(juju: jubilant.Juju):
    juju.deploy('myapp')
    juju.deploy('yourapp')

    def ready(status: jubilant.Status) -> bool:
        return (
            status.apps['myapp'].is_active and
            status.apps['yourapp'].is_blocked and
            'waiting' in status.apps['yourapp'].app_status.message
        )

    juju.wait(ready, error=jubilant.any_error)
```

james-garner-canonical:
Ah you're right, I was focusing on the `'waiting'` condition. That's perfect then.

PR #197, opened 2025-10-06
ci: publish the jubilant-backports branch to the jubilant-backports PyPI package

tonyandrewmeyer:
The PR:
* Changes the project name (in this branch) to `jubilant-backports`
* Changes the version (in this branch) to `1.0.0` in preparation of the next release (the latest one from tonyandrewmeyer/jubilant-backports is 1.0.0a1).
* Adjusts the README to reflect the above.

Note that the only differences between the existing jubilant-backports and the current one, in terms of usage, are:
* Instead of `import jubilant_backports as jubilant` you just do `import jubilant`
* `Juju.cli_version` and `Juju.cli_major_version` have been removed.
* The status and task classes are exactly the same types as in `jubilant`, rather than having the same name but being a different class. If anyone was doing something like `jubilant.Status | jubilant_backports.Status` that's not required any more. This does mean that there are a couple of fields that will be empty in 2.9 (like "base") and some that will be empty in 3.x (like "series"), but the 3.x ones could already be empty.
* The `TaskError` class handles both `Task` and `ExecTask`, so `pytest.raises(TaskError)` works in 2.9 and 3.x.

This PR also includes d18f143693bd1db2523956f44c3072341dd97c21 and adab08c4a96c6f23493e847310fe94f145f4c3bf, because I started it from main@HEAD instead of the revision that the branch came from. I've opened https://github.com/canonical/jubilant/pull/198 to do that separately and will rebase when that's merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Because Jubilant-backports calls the Juju CLI, you'll also need to [install Juju
 To use Jubilant-backports in Python code:
 
 ```python
-import jubilant_backports as jubilant
+import jubilant
 
 juju = jubilant.Juju()
 juju.deploy('snappass-test')
@@ -35,7 +35,7 @@ Below is an example of a charm integration test. First we define a module-scoped
 
 ```python
 # conftest.py
-import jubilant_backports as jubilant
+import jubilant as jubilant
 
 @pytest.fixture(scope='module')
 def juju():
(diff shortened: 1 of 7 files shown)


benhoyt (approved):
Looks good to me, thanks!

PR #202, opened 2025-10-08
refactor: install jubilant-backports into the jubilant_backports namespace

tonyandrewmeyer:
Rather than installing `jubilant-backports` into the `jubilant` namespace, install into `jubilant_backports`. This increases compatibility with 1.0.0a1, and avoids issues where installing both `jubilant` and `jubilant-backports` would conflict.

The PR:
* Brings back to the README the suggestion to use `import jubilant_backports as jubilant` for easier movement to the regular package when Juju 2.9 support is no longer required.
* Moves `jubilant` to `jubilant_backports`
* Adjusts the tests to import `jubilant_backports as jubilant` (and cleans up a few places that were using both, leftover from the original fork).
* Adjusts some machinery (like the Makefile) to handle the new name.

I've also bumped the version from 1.0.0 to 1.4.0. I think keeping the major.minor aligned with regular Jubilant is probably worth doing, as it hints at which Jubilant features are here. We probably won't have any new minor versions here, so it doesn't make much difference, and it leaves the last number free to handle any bug fixes specific to the branch.


benhoyt (approved):
Looks good to me, thanks. The docs build is failing, but I presume that's intended.

Speaking of docs, are you planning to add a note about this to our main docs?

tonyandrewmeyer:
> Looks good to me, thanks. The docs build is failing, but I presume that's intended.

Yes, I don't think branch should publish to the stable docs, and I don't think it's worth having a docs version for the branch, since the API is almost identical and you're meant to use it as a drop-in replacement.

@dwilding might know how to adjust things so readthedocs doesn't try to build on this branch?

> Speaking of docs, are you planning to add a note about this to our main docs?

I hadn't thought about that, but I guess there should be somewhere. I'll find a place and open a very small PR after this is released.

PR #205, opened 2025-10-08
feat: make Juju.temp_dir public, improve docs

benhoyt:
This makes the `Juju._temp_dir` helper public, for use by tests that do `scp` and the like, when Juju is running as a snap.

I'm not *entirely* sure this is a good idea, or necessarily the API we want, so would like feedback on that.

Fixes #201.

PR #207, opened 2025-10-16
feat: copy files to juju-accessible place for deploy, refresh, scp

### LINKED ISSUE #201: Juju scp: no such file or directory
This issue describe a problem with the Jubilant `Juju.scp` function (wrapping juju's [scp](https://documentation.ubuntu.com/juju/3.6/reference/juju-cli/list-of-juju-cli-commands/scp/) command), where it cannot locate local files.

### How to reproduce

<details>

```python
import os
import pathlib
import tempfile

import jubilant
from jubilant import Juju
from jubilant.statustypes import UnitStatus

MYSQL_APP_NAME = "mysql"


def _get_juju_keys() -> tuple[str, str]:
    """Get Juju public and private keys."""
    config_dir = pathlib.Path("~/.local/share/juju")
    if juju_data := os.getenv("JUJU_DATA"):
        config_dir = pathlib.Path(juju_data)

    return (
        str(config_dir.expanduser().resolve() / "ssh" / "juju_id_rsa.pub"),
        str(config_dir.expanduser().resolve() / "ssh" / "juju_id_rsa"),
    )


def get_app_units(juju: Juju, app_name: str) -> dict[str, UnitStatus]:
    """Get the units for the given application."""
    model_status = juju.status()
    app_status = model_status.apps[app_name]
    return app_status.units


def scp_unit_file(juju: Juju, source: str, target: str) -> None:
    """Copy a file from source to target."""
    juju_keys = _get_juju_keys()
    juju.scp(source, target, host_key_checks=False, scp_options=["-B", "-i", juju_keys[1]])


if __name__ == "__main__":
    juju = Juju(model="testing")
    
    juju.deploy(
        MYSQL_APP_NAME,
        app=MYSQL_APP_NAME,
        base="ubuntu@22.04",
        config={"profile": "testing"},
        num_units=1,
    )
    juju.wait(
        ready=lambda status: all((
            jubilant.all_agents_idle(status),
            jubilant.all_active(status, MYSQL_APP_NAME),
        )),
        timeout=20 * 60,
    )
    
    for unit_name in get_app_units(juju, MYSQL_APP_NAME):
        with tempfile.NamedTemporaryFile(mode="w") as temp_file:
            temp_file.write("data")
            temp_file.flush()
            scp_unit_file(
                juju=juju,
                source=f"{temp_file.name}",
                target=f"{unit_name}:/tmp/file",
            )
```
</details>

### Traceback:

<details>

```log
juju scp --model testing --no-host-key-checks -- -B -i /root/.local/share/juju/ssh/juju_id_rsa /tmp/tmp12d6m2pr mysql/0:/tmp/file
ERROR exit status 255 (scp: stat local "/tmp/tmp12d6m2pr": No such file or directory)
```
</details>

### Work-around

<details>

```diff
- def scp_unit_file(juju: Juju, source: str, target: str) -> None:
-     """Copy a file from source to target."""
-     juju_keys = _get_juju_keys()
-     juju.scp(source, target, host_key_checks=False, scp_options=["-B", "-i", juju_keys[1]])

+ import subprocess
+
+ def get_unit_ip(juju: Juju, app_name: str, unit_name: str) -> str:
+     """Get the application unit IP."""
+     model_status = juju.status()
+     app_status = model_status.apps[app_name]
+     for name, status in app_status.units.items():
+         if name == unit_name:
+             return status.public_address
+
+     raise Exception("No application unit found")
+
+
+ def scp_file_to_unit(juju: Juju, app_name: str, unit_name: str, source_path: str) -> None:
+     """Copy a file from a given unit."""
+     juju_keys = _get_juju_keys()
+     unit_address = get_unit_ip(juju, app_name, unit_name)
+     unit_username = "ubuntu"
+
+     subprocess.check_output([
+         "scp",
+         "-B",
+         "-o",
+         "StrictHostKeyChecking=no",
+         "-i",
+         juju_keys[1],
+         f"{source_path}",
+         f"{unit_username}@{unit_address}:/tmp/file",
+     ])
```
</details>

### Versions

- Jubilant: `1.4.0` (latest).
- Python: `3.10`
- Juju: `3.6.9` (latest)
- LXD: `5.21.4` (latest)
- Ubuntu: `24.04` (latest LTS).

benhoyt:
This is an alternative to #205 (making `Juju.temp_dir` public): it changes `deploy`, `refresh`, and `scp`, when used with local files, to copy the file to (or from) a snap-accessible location before (or after) running the command.

It's best to view the diff with "Hide whitespace" turned on, as much of the changes to `_juju.py` are just indentation changes.

Fixes #201.

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -16,6 +16,7 @@ from typing import Any, Generator, Literal, Union, overload
 
 from . import _pretty, _yaml
 from ._task import Task
+from ._version import Version
 from .secrettypes import RevealedSecret, Secret, SecretURI
 from .statustypes import Status
 
@@ -1215,6 +1216,12 @@ class Juju:
             file.flush()
             args.extend(['--file', file.name])
             self.cli(*args)
+
+    def version(self) -> Version:
+        """Return the parsed Juju CLI version."""
+        stdout = self.cli('version', '--format', 'json', '--all', include_model=False)
+        version_dict = json.loads(stdout)
+        return Version._from_dict(version_dict)
 
     def wait(
         self,
--- a/jubilant/__init__.py
+++ b/jubilant/__init__.py
@@ -17,6 +17,7 @@ from ._all_any import (
 from ._juju import CLIError, ConfigValue, Juju, WaitError
 from ._task import Task, TaskError
 from ._test_helpers import temp_model
+from ._version import Version
 from .secrettypes import RevealedSecret, Secret, SecretURI
 from .statustypes import Status
 
@@ -30,6 +31,7 @@ __all__ = [
     'Status',
     'Task',
     'TaskError',
+    'Version',
     'WaitError',
     'all_active',
     'all_agents_idle',
--- a/jubilant/_version.py
+++ b/jubilant/_version.py
@@ -1,0 +1,94 @@
+from __future__ import annotations
+
+import dataclasses
+import re
+from typing import Any
+
+# Per https://github.com/juju/juju/blob/afed87bbe98189805bc81e1af5d5fac70195fc86/core/semversion/version.go#L86-L91,
+# support version strings in the following formats:
+#
+# 1.2-release-arch
+# 1.2.3-release-arch
+# 1.2.3.4-release-arch
+# 1.2-alpha3-release-arch
+# 1.2-alpha3.4-release-arch
+majmin = r'^(\d+)\.(\d+)'
+relarch = r'([^-]+)-([^-]+)$'
+majmin_re = re.compile(rf'{majmin}-{relarch}')
+majmin_patch_re = re.compile(rf'{majmin}\.(\d+)-{relarch}')
+majmin_patch_build_re = re.compile(rf'{majmin}\.(\d+)\.(\d+)-{relarch}')
+majmin_tag_re = re.compile(rf'{majmin}-([a-z]+)(\d+)-{relarch}')
+majmin_tag_build_re = re.compile(rf'{majmin}-([a-z]+)(\d+)\.(\d+)-{relarch}')
+
+
+@dataclasses.dataclass(frozen=True)
+class Version:
+    """Parsed Juju CLI version as returned by ``juju version --format=json --all``.
+
+    The version from Juju is normally in the form ``major.minor.patch-release-arch``.
+    You can compare versions using the :attr:`tuple` property::
+
+        if juju.version().tuple >= (3, 6, 11):
+            ...  # Juju CLI is at least 3.6.11
+    """
+
+    major: int
+    minor: int
+    patch: int
+    release: str = ''
+    arch: str = ''
+
+    tag: str | None = None
+    build: int | None = None
+
+    git_commit: str | None = None
+
+    def __str__(self) -> str:
+        """The string representation of the parsed version.
+
+        This recreates the version string from Juju.
+        """
+        prefix = f'{self.major}.{self.minor}'
+        if self.tag is not None:
+            prefix += f'-{self.tag}{self.patch}'
+        elif self.patch != 0:
+            prefix += f'.{self.patch}'
+        if self.build is not None:
+            prefix += f'.{self.build}'
+        return f'{prefix}-{self.release}-{self.arch}'
+
+    @property
+    def tuple(self) -> tuple[int, int, int]:
+        """The tuple ``(major, minor, patch)``."""
+        return self.major, self.minor, self.patch
+
+    @classmethod
+    def _from_dict(cls, d: dict[str, Any]) -> Version:
+        version = d['version']
+        patch = 0
+        tag = None
+        build = None
+
+        if match := majmin_re.match(version):
+            major, minor, release, arch = match.groups()
+        elif match := majmin_patch_re.match(version):
+            major, minor, patch, release, arch = match.groups()
+        elif match := majmin_patch_build_re.match(version):
+            major, minor, patch, build, release, arch = match.groups()
+        elif match := majmin_tag_re.match(version):
+            major, minor, tag, patch, release, arch = match.groups()
+        elif match := majmin_tag_build_re.match(version):
+            major, minor, tag, patch, build, release, arch = match.groups()
+        else:
+            raise ValueError(f'invalid version {version!r}')
+
+        return Version(
+            major=int(major),
+            minor=int(minor),
+            patch=int(patch),
+            release=release,
+            arch=arch,
+            tag=tag,
+            build=int(build) if build is not None else None,
+            git_commit=d.get('git-commit'),
+        )
--- a/tests/integration/test_deployment.py
+++ b/tests/integration/test_deployment.py
@@ -70,3 +70,9 @@ def test_refresh_path(juju: jubilant.Juju):
     juju.wait(
         lambda status: status.apps['testdb'].units['testdb/0'].workload_status.current == 'unknown'
     )
+
+
+def test_version(juju: jubilant.Juju):
+    version = juju.version()
+    raw_version = juju.cli('version', include_model=False).strip()
+    assert str(version) == raw_version
--- a/tests/unit/test_version.py
+++ b/tests/unit/test_version.py
@@ -1,0 +1,68 @@
+import json
+
+import pytest
+
+import jubilant
+
+from . import mocks
+
+
+def test_simple(run: mocks.Run):
+    version_dict = {
+        'version': '3.6.11-genericlinux-amd64',
+        'git-commit': '17876b918429f0063380cdf07dc47f98a890778b',
+    }
+    run.handle(['juju', 'version', '--format', 'json', '--all'], stdout=json.dumps(version_dict))
+
+    juju = jubilant.Juju()
+    version = juju.version()
+
+    assert version == jubilant.Version(
+        3,
+        6,
+        11,
+        release='genericlinux',
+        arch='amd64',
+        git_commit='17876b918429f0063380cdf07dc47f98a890778b',
+    )
+    assert version.tuple == (3, 6, 11)
+
+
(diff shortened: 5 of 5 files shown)


jubilant/_juju.py
```diff
@@ -478,52 +478,66 @@ def deploy(
         if isinstance(overlays, str):
             raise TypeError('overlays must be an iterable of str or pathlib.Path, not str')
 
-        args = ['deploy', str(charm)]
-        if app is not None:
-            args.append(app)
+        charm = str(charm)
+        temp_needed = charm.startswith(('.', '/')) and self._juju_is_snap
```
dimaqq: From in person discussion: for vm charm that need a resource, the latter too could be a local file.

benhoyt: Good idea -- now done!

jubilant/_juju.py
```diff
@@ -1314,3 +1340,42 @@ def _status_line_ok(line: str) -> bool:
...
+            assert resources is not None
+            resources_temp = {}
+            for k, v in resources.items():
+                if not v.startswith(('.', '/')):
+                    resources_temp[k] = v
+                    continue
+                temp = os.path.join(temp_dir, k)
+                shutil.copy(v, temp)
+                resources_temp[k] = temp
+            resources = resources_temp
```
dimaqq: **minor**
This could be a little shorter like so:

```suggestion
        if resources_needs_temp:
            assert resources is not None
            resources = {**resources}
            for k, v in resources.items():
                if v.startswith(('.', '/')):
                    shutil.copy(v, temp := os.path.join(temp_dir, k))
                    resources[k] = temp
```

resources can be updated in place (values overwritten only at existing keys), though it must be a copy (shallow ok) so as not to clobber function arg.

jubilant/_juju.py
```diff
@@ -489,52 +489,54 @@ def deploy(
         if isinstance(overlays, str):
             raise TypeError('overlays must be an iterable of str or pathlib.Path, not str')
 
-        args = ['deploy', str(charm)]
-        if app is not None:
-            args.append(app)
+        with _deploy_tempdir(self, charm, resources) as (_charm, resources):
```
dimaqq: **minor** I think that if it needs "self", it may as well be a method, though I could be convinced otherwise.

```suggestion
        with self._deploy_tempdir(charm, resources) as (_charm, resources):
```

tests/integration/test_execution.py
```diff
@@ -137,6 +139,11 @@ def test_ssh_and_scp(juju: jubilant.Juju):
     passwd = pathlib.Path('passwd').read_text()
     assert 'redis:' in passwd
 
+    # Test a round trip
+    juju.scp(__file__, 'snappass-test/0:/tmp/foobar.py')
+    juju.scp('snappass-test/0:/tmp/foobar.py', 'foobar.py')
+    assert pathlib.Path('foobar.py').read_text() == pathlib.Path(__file__).read_text()
```
dimaqq: Nice trick, but I don't think this really tests temp dir for snaps.
IIRC github actions checkout is done to home directory, which snapped juju can read.

To test required tempifictation, have your **local path** under `/tmp/`.

jubilant/_juju.py
```diff
@@ -997,17 +1000,40 @@ def scp(
...
+                args.append(source)
+                args.append(file_temp.name)
+                self.cli(*args)
+                shutil.copy(file_temp.name, destination)
 
-        self.cli(*args)
+            else:
+                args.append(source)
+                args.append(destination)
+                self.cli(*args)
```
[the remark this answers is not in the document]
benhoyt: Done

jubilant/_juju.py
```diff
@@ -997,17 +1000,41 @@ def scp(
...
+        ) as file_temp:
+            args = ['scp']
+            if container is not None:
+                args.extend(['--container', container])
+            if not host_key_checks:
+                args.append('--no-host-key-checks')
+            args.append('--')
+            args.extend(scp_options)
+
+            if file_temp is not None and ':' not in source:
```
tonyandrewmeyer: You can `juju scp` with a local source and local destination. I'm not sure why someone would do that, but it seems like we should have it work. So I think these should be switched so that it handles adding the source arg, then handles adding the destination arg, then outside of that runs the command (and then does the destination copy if needed).

jubilant/_juju.py
```diff
@@ -1297,6 +1324,43 @@ def _temp_dir(self) -> str:
...
+    # is running as a snap (in which case /tmp is not accessible).
+    @contextlib.contextmanager
+    def _deploy_tempdir(
+        self,
+        charm: str | pathlib.Path | None,
+        resources: Mapping[str, str] | None,
+    ) -> Generator[tuple[str | None, Mapping[str, str] | None]]:
+        if charm is not None:
+            charm = str(charm)
+        charm_needs_temp = charm is not None and charm.startswith(('.', '/'))
```
tonyandrewmeyer: A path could also start with `~`, maybe other things as well? Perhaps we should [parse as a URL](https://github.com/juju/juju/blob/c19c662002d32b5710120084bd3bc2df4cf7c4b2/internal/charm/url.go#L139)? I'm also ok with starting simple and adjusting if charmers come to us with issues.

tonyandrewmeyer (approved):
Thanks - I think this will be a nice developer experience improvement, particularly given how confusing the error messages (or lack thereof) are from Juju.

jubilant/_juju.py
```diff
@@ -786,29 +788,30 @@ def refresh(
...
+                args.extend(['--base', base])
+            if channel is not None:
+                args.extend(['--channel', channel])
+            if config is not None:
+                for k, v in config.items():
+                    args.extend(['--config', _format_config(k, v)])
+            if force:
+                args.extend(['--force', '--force-base', '--force-units'])
+            if path is not None:
+                args.extend(['--path', path])
```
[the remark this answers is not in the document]
benhoyt: No, that's not correct. This is the `str | None` returned by the `_deploy_tempdir` context manager, and it can't be None.

jubilant/_juju.py
benhoyt: I think it's the shell that expands `~`, so you wouldn't be expecting it in this context. Also, I believe `juju deploy` determines whether a charm is a local file by whether it starts with `.` or `/`, so I'm following that logic here. But you're right, if we need to adjust later we can.

jubilant/_juju.py
```diff
@@ -997,17 +1000,41 @@ def scp(
...
-            args.extend(['--container', container])
-        if not host_key_checks:
-            args.append('--no-host-key-checks')
-        args.append('--')
-        args.extend(scp_options)
-        args.append(str(source))
-        args.append(str(destination))
+        source = str(source)
+        destination = str(destination)
+        temp_needed = (':' not in source or ':' not in destination) and self._juju_is_snap
```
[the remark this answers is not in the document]
benhoyt: Good call, thx -- I've updated this logic.

jubilant/_juju.py
```diff
@@ -997,17 +1000,41 @@ def scp(
...
+            elif file_temp is not None and ':' not in destination:
+                # Remote source, local destination
+                args.append(source)
+                args.append(file_temp.name)
+                self.cli(*args)
+
+                shutil.copy(file_temp.name, destination)
 
-        self.cli(*args)
+            else:
```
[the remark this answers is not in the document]
benhoyt: This has been reworked, per above

jubilant/_juju.py
tonyandrewmeyer: > I think it's the shell that expands `~`, so you wouldn't be expecting it in this context.

Ah, true.

> Also, I believe `juju deploy` determines whether a charm is a local file by whether it starts with `.` or `/`, so I'm following that logic here.

I could not find any Juju code that did that. I believe it's the code I linked above, which is parsing as a URL.

jubilant/_juju.py
```diff
@@ -489,52 +489,54 @@ def deploy(
         if isinstance(overlays, str):
             raise TypeError('overlays must be an iterable of str or pathlib.Path, not str')
 
-        args = ['deploy', str(charm)]
-        if app is not None:
-            args.append(app)
+        with self._deploy_tempdir(charm, resources) as (_charm, resources):
```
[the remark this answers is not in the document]
tonyandrewmeyer: I don't think there's any reason `resources` needs to be in the middle.

```python
        args = ['deploy']
        if app is not None:
            args.append(app)

        if attach_storage:
            if isinstance(attach_storage, str):
                args.extend(['--attach-storage', attach_storage])
            else:
                args.extend(['--attach-storage', ','.join(attach_storage)])
        if base is not None:
            args.extend(['--base', base])
        if bind is not None:
            if not isinstance(bind, str):
                bind = ' '.join(f'{k}={v}' for k, v in bind.items())
            args.extend(['--bind', bind])
        if channel is not None:
            args.extend(['--channel', channel])
        if config is not None:
            for k, v in config.items():
                args.extend(['--config', _format_config(k, v)])
        if constraints is not None:
            for k, v in constraints.items():
                args.extend(['--constraints', f'{k}={v}'])
        if force:
            args.append('--force')
        if num_units != 1:
            args.extend(['--num-units', str(num_units)])
        for overlay in overlays:
            args.extend(['--overlay', str(overlay)])
        if revision is not None:
            args.extend(['--revision', str(revision)])
        if storage is not None:
            for k, v in storage.items():
        if to:
            if isinstance(to, str):
                args.extend(['--to', to])
            else:
                args.extend(['--to', ','.join(to)])
        if trust:
            args.append('--trust')
        with self._deploy_tempdir(charm, resources) as (_charm, resources):
            assert _charm is not None
            args.insert(1, _charm)

            if resources is not None:
                for k, v in resources.items():
                    args.extend(['--resource', f'{k}={v}'])

            self.cli(*args)
```

Issue #211, opened 2025-10-21
Magical secret URI can't be passed to an action directly


dimaqq:
Code I tried to run:

```
    secrets = juju.secrets()
    assert secrets
    assert secrets[0].owner == 'test-secrets'
    good_secret = secrets[0].uri

    params['secretid'] = good_secret
    rv = juju.run(leader, 'set-secret-flow', params=params)
```

Crashed with:

```py
test/integration/test_secrets.py:119:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/code/operator/.venv/lib/python3.14/site-packages/jubilant/_juju.py:944: in run
    _yaml.safe_dump(params, params_file)
/code/operator/.venv/lib/python3.14/site-packages/jubilant/_yaml.py:45: in safe_dump
    return yaml.dump(data, stream=stream, Dumper=_safe_dumper)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/code/operator/.venv/lib/python3.14/site-packages/yaml/__init__.py:253: in dump
    return dump_all([data], stream, Dumper=Dumper, **kwds)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/code/operator/.venv/lib/python3.14/site-packages/yaml/__init__.py:241: in dump_all
    dumper.represent(data)
/code/operator/.venv/lib/python3.14/site-packages/yaml/representer.py:27: in represent
    node = self.represent_data(data)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
/code/operator/.venv/lib/python3.14/site-packages/yaml/representer.py:48: in represent_data
    node = self.yaml_representers[data_types[0]](self, data)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/code/operator/.venv/lib/python3.14/site-packages/yaml/representer.py:207: in represent_dict
    return self.represent_mapping('tag:yaml.org,2002:map', data)
[...]

PR #213, opened 2025-10-23
feat: add SecretURI.__repr__ to make type obvious in error messages

benhoyt:
See rationale in #211.


james-garner-canonical (approved):
Nice improvement, looks good to me!

A small suggestion for generality, but probably not needed since we're not expecting subclasses.

PR #230, opened 2025-11-27
fix: fix Version.__str__ and strictify parsing

benhoyt:
Currently '4.0.0-genericlinux-amd64' parsed and str'd gives '4.0-genericlinux-amd64', which isn't correct (and not what Juju's `semversion.Number.String` does). So fix that.

In addition, switch to Juju's "strict" parsing, which is what Juju's semversion.Parse does by default. Specifically, this means that "1.2-release-arch" is not supported. I think this is the right thing to do here.


benhoyt:
> I assume that "strict" mode is what juju version is always going to provide, so that seems correct to me too.

Yeah, `juju version` uses `Number.String` which always includes the patch version.

PR #236, opened 2025-12-19
feat: add model_constraints method to get/set machine constraints

### LINKED ISSUE #227: Enhance temp_model to allow setting arch as a model constraint
Can we enhance `jubilant.temp_model()` to optionally allow for specifying the architecture, and setting this as a model constraint before returning? 

This would simplify creating models for testing across different archs, and omit the need to do a `juju.cli("set-model-constraints", f"arch={arch}")`. 
(As a follow-up: should we consider creating a dedicated method for `set-model-constraints`? I'm not sure how heavily this is used though)

Soundarya03:
Defines `model_constraints` method that maps to `set-model-constraints` and `model-constraints` commands in the Juju CLI.

Fixes #227

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -1522,8 +1522,6 @@ def _format_config(k: str, v: ConfigValue) -> str:
 
 
 def _format_constraint(k: str, v: ConstraintValue) -> str:
-    if v is None:  # type: ignore
-        raise TypeError(f'unexpected None value for constraint key {k!r}')
     if isinstance(v, bool):
         v = 'true' if v else 'false'
     return f'{k}={v}'
--- a/tests/unit/test_model_constraints.py
+++ b/tests/unit/test_model_constraints.py
@@ -1,5 +1,3 @@
-import pytest
-
 import jubilant
 from tests.unit import mocks
 
@@ -7,12 +5,18 @@ from tests.unit import mocks
 def test_get(run: mocks.Run):
     run.handle(
         ['juju', 'model-constraints', '--format', 'json'],
-        stdout='{"arch":"amd64","cores":8,"mem":16384}',
+        stdout='{"arch":"amd64","cores":8,"mem":16384,"root-disk":1638.4,"allocate-public-ip":false}',
     )
 
     juju = jubilant.Juju()
     values = juju.model_constraints()
-    assert values == {'arch': 'amd64', 'cores': 8, 'mem': 16384}
+    assert values == {
+        'arch': 'amd64',
+        'cores': 8,
+        'mem': 16384,
+        'root-disk': 1638.4,
+        'allocate-public-ip': False,
+    }
 
 
 def test_set(run: mocks.Run):
@@ -22,19 +26,20 @@ def test_set(run: mocks.Run):
             'set-model-constraints',
             'arch=amd64',
             'cores=8',
-            'mem=16384',
+            'mem=16G',
+            'root-disk=1638.4',
             'allocate-public-ip=false',
         ]
     )
 
     juju = jubilant.Juju()
     retval = juju.model_constraints(
-        {'arch': 'amd64', 'cores': 8, 'mem': '16384', 'allocate-public-ip': False}
+        {
+            'arch': 'amd64',
+            'cores': 8,
+            'mem': '16G',
+            'root-disk': 1638.4,
+            'allocate-public-ip': False,
+        }
     )
     assert retval is None
-
-
-def test_format_constraint_type_error():
-    juju = jubilant.Juju()
-    with pytest.raises(TypeError):
-        juju.model_constraints({'foo': None})  # type: ignore


dimaqq:
Nevermind the failure against Juju 4 / k8s; that an issue in the master branch too.

jubilant/_juju.py
```diff
@@ -805,6 +805,43 @@ def model_config(
 
         self.cli(*args)
 
+    def model_constraints(
+        self,
+        model: str | None = None,
```
benhoyt: We should remove this argument and make it operate on the current model (`self.model`) like the other commands. Note that `_cli` does this automatically unless you specify `include_model=False`.

jubilant/_juju.py
```diff
@@ -805,6 +805,43 @@ def model_config(
 
         self.cli(*args)
 
+    def model_constraints(
```
benhoyt: I think it's the right choice to make this a single method, instead of two, even though it's two separate `juju` commands (`model-constraints` and `set-model-constraints`). At some point we had a spec that proposed merging those two commands so it'd be consistent with `juju config` and `juju model-config`, but that was never implemented. In any case, I think it was the intent and it's good for consistency with `Juju.config()` and `Juju.model_config()`. So thanks.

However, we should provide two typing `@overload`s, like we do with those methods.

jubilant/_juju.py
```diff
@@ -805,6 +805,43 @@ def model_config(
...
+        self,
+        model: str | None = None,
+        constraints: Mapping[str, bool | int | float | str] | None = None,
+    ) -> Mapping[str, bool | int | float | str] | None:
+        """Gets or sets machine constraints on a model.
+
+        Sets new constraints if *constraints* argument is provided.
+        If not, gets you the current details.
```
benhoyt:
```suggestion
        If called with no arguments, get the model constraints. If called with the
        *constraints* argument, set the model constraints and return None.
```

jubilant/_juju.py
```diff
@@ -805,6 +805,43 @@ def model_config(
...
+
+        Args:
+            model: Name of the model or ``controller:model``. If omitted,
+                this is taken to be the current model.
+            constraints: Named constraints to be set.
```
benhoyt:
```suggestion
            constraints: Model constraints to set, for example, ``{'mem': '8G', 'cores': 4}``.
```

jubilant/_juju.py
```diff
@@ -805,6 +805,43 @@ def model_config(
...
+            args.append('set-model-constraints')
+            args.extend(_format_config(k, v) for k, v in constraints.items())
+            if model is not None:
+                args.extend(['--model', model])
+                self.cli(*args, include_model=False)
+            else:
+                self.cli(*args)
+
+            return None
+        else:
```
benhoyt: Because there's a `return None` in the `if constraints` block, let's remove this `else` (similar to `config`). Though I'd prefer to reverse the blocks and have the "get" case first, as with `config` too.

benhoyt (changes requested):
Thanks for this -- looks like a good start. Several minor changes requested. It'd also be good to add an integration test of both the get/set case. Perhaps in `test_machine.py` setting `{"mem": "1G", "cores": 4}` or something.

tests/unit/test_model_constraints.py
```diff
@@ -0,0 +1,20 @@
...
+    values = juju.model_constraints()
+    assert values == {'arch': 'amd64', 'cores': 8, 'mem': 16384}
+
+
+def test_set(run: mocks.Run):
+    run.handle(['juju', 'set-model-constraints', 'arch=amd64', 'cores=8', 'mem=16384'])
+
+    juju = jubilant.Juju()
+    values = juju.model_constraints(constraints={'arch': 'amd64', 'cores': '8', 'mem': '16384'})
+    assert values is None
```
benhoyt: Let's also add a test of the `_format_constraint` None check (similar to how we do with config). Something like this:

```python
def test_format_constraint_type_error():
    juju = jubilant.Juju()
    with pytest.raises(TypeError):
        juju.model_constraints({'foo': None})  # type: ignore

**Comment by benhoyt** [comment]
Can you also please update the branch by merging from main?

**Review by benhoyt — COMMENTED** [review]
Looks good to me. However, there are a couple of `docs` files included in the PR when they shouldn't be. I'll also request a second review from @tonyandrewmeyer (who's only back at work next week) in case he has any concerns about the API.

File: jubilant/_juju.py [hunk 7]
```diff
@@ -1486,6 +1515,14 @@ def _format_config(k: str, v: ConfigValue) -> str:
     return f'{k}={v}'
 
 
+def _format_constraint(k: str, v: ConstraintValue) -> str:
+    if v is None:  # type: ignore
```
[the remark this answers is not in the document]
benhoyt: This is defensive programming and intentional

docs/conf.py
```diff
@@ -222,13 +222,13 @@
...
-# A regex list of URLs that are ignored by 'make linkcheck'
-#
-# TODO: [@dwilding DONE] Remove or adjust the ACME entry after you update the contributing guide
+# During linkcheck, if a target URL matches any of these regexes (using re.match)
+# then the URL is skipped.
 
 linkcheck_ignore = [
-    "http://127.0.0.1:8000"
-    ]
+    # Excluded because the pages don't contain elements with an ID matching the URL fragment.
```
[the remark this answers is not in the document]
benhoyt: This change shouldn't be included in this PR, so ignoring.

tonyandrewmeyer (changes requested):
Overall looks good to me, thanks for doing this!

One very small test expansion suggestion.

Also, I re-opened a copilot suggestion, where I disagree with @benhoyt - if either of you could convince me why we need to handle None, that would be great, thanks!

jubilant/_juju.py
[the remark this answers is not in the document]
benhoyt: That's fair push-back. The `None` check was originally added to `_format_config` because the config commands originally accepted None (it meant "reset", but now we have a separate *reset* argument). James [suggested](https://github.com/canonical/jubilant/pull/125#discussion_r2065050629) adding a None check because it was a (pre-1.0) breaking change:

> Worth making _format_config do something sensible with None values? Maybe raise an error? I know we normally rely on the type checker for cases like this, but maybe it's worthwhile here since we're making a breaking change.

Which I think made sense at the time, but doesn't now. I think you're right, and it's totally fine to remove the None checks now.

@Soundarya03 Given the above, let's please get rid of the None check in both helper functions (and remove the associated tests).

PR #242, opened 2026-01-15
ci: install Copilot files from the copilot-collections tool

tonyandrewmeyer:
This PR completes onboarding of a Charm Tech repository into the [copilot-collections](https://github.com/canonical/copilot-collections) tool (Jubilant is a test case to explore whether we want to go all-in on the copilot-collections system).

Adds configuration to load four "collections": `charm-tech-core`, `charm-tech-python-library`, `charm-tech-docs`, and `charm-tech-make` (all of these are defined in the copilot-collections repository).

Also adds a GitHub workflow to keep the files in the Jubilant repository in sync with the ones upstream.

The files added by doing a sync are also included (via running a local sync). If we want to make improvements to these, that needs to be done in a PR in the copilot-collections repository (any changes here will be undone via the automation).


tonyandrewmeyer:
> I'm okay with this for experimentation, but overall quite skeptical about how much "hidden documentation for machines" there is in all of this.

I do feel like the quantity of instructions in the existing copilot-collections is generally on the too-large side. For example, code comments are something that agents do very badly, so instructions are needed (more so than for humans, I think), but the [existing guide](https://github.com/canonical/copilot-collections/blob/main/assets/instructions/python/code-commenting.instructions.md) is 165 lines, 4.7K and that feels way too much to me (not just in terms of how much content gets maintained, but also in terms of consuming context space).

My plan is to get one of our repos (I picked Jubilant because it's the last main repo that doesn't already have instructions in progress) using this system, and then work with the upstream to get the actual instructions tighter.

> I'm worried it'll mean we don't improve our regular README and STYLE guides and whatnot, or that it won't be kept up to date.

This is another way in which Jubilant feels like an interesting place to try this system out to me. At the moment, Jubilant is extremely minimal in terms of style and contributing guidelines (a very small bit in the README, an implicit rule that the operator repo STYLE.md applies, some high level design guidelines in the docs, the enforcement via tooling). We've had a few contributions to Jubilant outside of Charm Tech, and had mixed success with this minimal approach, I think.

If we expanded the guidance for humans to contribute, then there would perhaps be less use for the guidance for machines (but I don't think you can really do a good job of writing for both in one document). If we keep the guidance for humans more minimal, then more machine guidance will be useful, I think.

In terms of keeping the AI documentation up to date, that's what the automation is for. If we have a small set of files that are used across all our repos, and updates are automatically pulled into each repo when updated in the upstream repo, then that should make it much more likely that we are consistent and don't get out-of-date.

> Maybe part of our larger Copilot discussion.

Added to the agenda.

> My own request here is: can we avoid yet another file in the root directory? (`.copilot-collections.yaml`)

Opened https://github.com/canonical/copilot-collections/issues/13.

benhoyt:
Great, thanks for the response. I'm happy to experiment here, and yeah, we'll discuss further in a couple of weeks.

dwilding (approved):
Realised that I didn't approve yesterday. I think my concerns are more to do with copilot-collections, and I don't want to block experimentation

tonyandrewmeyer:
> If I understand correctly, all the added files in `.github` (except the workflow?) were added by the local sync, and will be maintained by the workflow.

Yes, the [first commit](https://github.com/canonical/jubilant/pull/242/commits/dfd29ab5795cb3ea6a50e2325dd66176aba60f1a) has what's really being added here, and the second commit has everything that the sync system brings (and the third commit is just a regular merge).

> I have a couple of concerns:
> 
> * Will we/others remember not to edit these files?

This is similar to the files we have in canonical/operator that are maintained by workflows (the compatibility workflow pinning, the best practices list). If someone does edit them by hand then when the workflow next runs it'll delete those changes in the PR that it opens. I'd expect that the people doing the review of that PR would question anything being deleted that shouldn't be.

That being said, there is some risk still. However, like your later comment says, this is really an issue with the copilot-collections system.

In practice, if the copilot-collections system is successful, I would expect that we use it across all the Charm Tech repos, and so it's habit to update everything there, and we'd probably have that in the "how Charm Tech works" document.

> * The ultimate source of each file is painful to track down, as the copilot-collections scripts seem to bundle together standard instructions with our own custom instructions in that repo.

By "standard" do you mean ones that are in the copilot-collections "core" and "common" groups? This is again (and again as your later comment said) more something to raise in copilot-collections, I think. The intention certainly is, from what I understand, to have files that are used across teams and within teams in the copilot-collections repository (we only have ones used in Jubilant right now, with this branch, but the intention is to expand that if we are satisfied with it).

We are using:
 - charm-tech-core
 - charm-tech-python-library
 - charm-tech-docs
 - charm-tech-make

The "charm-tech-" prefix should be a strong indicator to look in copilot-collections groups/charm-tech. In there, many of the files have the exact path in the `dst` field. The `includes` makes it more complicated because you need to keep following those until there are none left. We could have a Charm Tech policy of not using `includes`, and/or we could argue upstream that `includes` is an unnecessary feature at this point and it would be better to start simpler.

> There's not a huge amount we can do in this repo I suppose, as we're limited by the copilot-collections tooling.

Yes, but we can open issues and/or PRs there.

> I wonder whether each file in that repo could be stamped with its location (in the frontmatter) so that we can (1) more easily identify imported files that shouldn't be edited directly, and (2) more easily figure out where to propose edits.

Perhaps, but there's no way to do this without that content going into the LLM as well, which unnecessarily consumes tokens and potentially confuses the LLM.

> **Tangent:** Is it really necessary to duplicate all the instructions in this repo? I'm imagining some kind of tool that Copilot/agents could dynamically use to pull down the latest relevant instructions. Something like `uvx canonical-instructions charm-tech-docs`. Would a layer of indirection make the system less effective?

It's an interesting idea. I'm not sure that Copilot reviewing would always run an initial command, and I'm not sure whether a local system would do so either.

I can experiment with this, though. If it does work then it should be recommend upstream, I think.

tonyandrewmeyer:
> > **Tangent:** Is it really necessary to duplicate all the instructions in this repo? I'm imagining some kind of tool that Copilot/agents could dynamically use to pull down the latest relevant instructions. Something like `uvx canonical-instructions charm-tech-docs`. Would a layer of indirection make the system less effective?
> 
> It's an interesting idea. I'm not sure that Copilot reviewing would always run an initial command, and I'm not sure whether a local system would do so either.
> 
> I can experiment with this, though.

I committed an AGENTS.md to main in my fork that says only:

````md
It is **CRITICAL** that the first thing you do when doing any work in this repository, including code review, is run the command:

```
curl -sL https://raw.githubusercontent.com/canonical/copilot-collections/main/scripts/local_sync.sh | bash
```

That will replace the contents of this file, so you should reload it after that is complete.
````

With the `.copilot-collections.yaml` file from this PR. I created an issue for Copilot to work on in a cloud session, and interestingly it did do as I said:

Screenshot 2026-01-20 at 8 47 30 PM

I think it will try to commit all those files in the branch before opening the PR, but if they were added to `.gitignore` that would avoid that.

I don't think it was really using the instructions, although I'd have to look more closely (but it failed to use the knowledge about using `make` for example). I guess the best test would be to do the same ticket but with all the files available locally.

I'll try to do a similar review experiment as well, and perhaps a CLI one, and perhaps a "in VS Code" one too. I might move this experiment out of this PR's thread, though.

tonyandrewmeyer:
I'm going to merge this as-is and continue the "what if we just download the files every time" experiment elsewhere. I'll keep the team informed :)

PR #243, opened 2026-01-19
feat: provide additional arguments for model destruction

### LINKED ISSUE #239: Timeout in destroy_model method
Hello,

Do you consider allowing client code to pass `timeout` to the `juju.destroy_model` method? I understand that we can always resort to the `juju.cli`, but it would be nicer to have it in the `destroy_model` if possible.

wood-push-melon:
This pull request adds the `release_storage`, `timeout`, and `no_wait` to the `destroy-model` method.

Fixes #239

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -623,6 +623,9 @@ class Juju:
         *,
         destroy_storage: bool = False,
         force: bool = False,
+        no_wait: bool = False,
+        release_storage: bool = False,
+        timeout: float | None = None,
     ) -> None:
         """Terminate all machines (or containers) and resources for a model.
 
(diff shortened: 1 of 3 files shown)


jubilant/_juju.py
```diff
@@ -622,7 +622,10 @@ def destroy_model(
         model: str,
         *,
         destroy_storage: bool = False,
+        release_storage: bool = False,
         force: bool = False,
+        timeout: float | None = None,
+        no_wait: bool = False,
```
benhoyt: In other functions, I've tried to keep the keyword-only arguments in alphabetical order, to keep things consistent and avoid an arbitrary order. Can we do that here too (and in the docstrings "Args" section, and in the code where possible too)?

tests/unit/test_destroy_model.py
```diff
@@ -21,10 +21,25 @@ def test_destroy_other(run: mocks.Run):
     assert juju.model == 'initial'
 
 
-def test_args(run: mocks.Run):
-    run.handle(['juju', 'destroy-model', 'bad', '--no-prompt', '--destroy-storage', '--force'])
+def test_all_args(run: mocks.Run):
```
benhoyt: I think we shouldn't test invalid combinations of arguments here: 1) because it's not good as an example, and 2) if we do add typing overloads to narrow this down, the test will break.

So I think we should have 2-4 named tests here, that each test a different valid combination of arguments.

wood-push-melon:
> Thanks for this. Some minor comments on the documentation and argument order, and a suggested change to the tests.

Thanks @benhoyt, will take a look.

PR #256, opened 2026-02-11
docs: WaitError is raised when error is true, not false.

tonyandrewmeyer:
We raise `WaitError` when the provided method returns a truthy value, not False. The other docs already have this the correct way around.


dwilding (approved):
Good catch, thanks!

PR #262, opened 2026-02-24
test: remove requests dependency

### LINKED ISSUE #248: Bump up testing urllib3 version to >2.5.0
This is to fix a security issue. Currently only used by tests, so not critical.

Per discussion with Dima, this will likely require a bump to Python 3.10+.

dimaqq:
Fixes #248 

Repo tooling complains that the locked `urllib3` version has a vulnerability.
That dep is brought in by `requests` which are only used in integration tests.
We run unit tests on wide array of Pythons, but integration tests only on 3.12.

In case we want to run integration tests on Python 3.8, this PR replaces `requests` usage with Python standard library.

### DIFF
--- a/tests/integration/test_deployment.py
+++ b/tests/integration/test_deployment.py
@@ -22,7 +22,7 @@ def test_deploy(juju: jubilant.Juju):
 
     with urllib.request.urlopen(f'http://{address}:5000/', timeout=10) as resp:
         assert 200 <= resp.status < 300
-        body = resp.read().decode('utf-8', errors='replace')
+        body = resp.read().decode()
 
     assert '<title>' in body
     assert 'snappass' in body.lower()


tests/integration/test_deployment.py
```diff
@@ -18,10 +19,13 @@ def test_deploy(juju: jubilant.Juju):
     # Setup has already done "juju deploy", this tests it.
     status = juju.status()
     address = status.apps['snappass-test'].units['snappass-test/0'].address
-    response = requests.get(f'http://{address}:5000/', timeout=10)
-    response.raise_for_status()
-    assert '<title>' in response.text
-    assert 'snappass' in response.text.lower()
+
+    with urllib.request.urlopen(f'http://{address}:5000/', timeout=10) as resp:
+        assert 200 <= resp.status_code < 300
```
benhoyt: I believe this is spelled `resp.status` (`resp.status_code` gives me an error).

james-garner-canonical (approved):
The change looks reasonable to me, and nice to drop this dependency since we use it for so little.

PR #264, opened 2026-03-05
refactor: add the ssh key earlier, so that no retry is needed

tonyandrewmeyer:
In https://github.com/juju/juju/issues/21425#issuecomment-3998249626, Heather explains that if we add the SSH key to the model before deploying any application, the key will be there when the application is deployed.

This PR rearranges the machine integration tests slightly so that the SSH key is added earlier, and removes the retrying in the SSH test. It does mean the key is available for other tests, but it's just ignored so that should be fine.

The PR also works around a [Juju bug](https://github.com/juju/juju/issues/21927) where exec stdout has newlines trimmed. The tests will break when the Juju bug is fixed, so we will know that the workaround should be removed.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -28,6 +28,12 @@ def juju(request: pytest.FixtureRequest) -> Generator[jubilant.Juju]:
 
 
 @pytest.fixture(scope='module')
+def juju_version(juju: jubilant.Juju) -> jubilant.Version:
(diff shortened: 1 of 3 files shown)


tonyandrewmeyer:
The newline issue is indeed a change in Juju (4.0.2 to 4.0.3 I believe, since the tests were passing with 4.0.1 and I think 4.0.2). Opened https://github.com/juju/juju/issues/21927

Issue #170, opened 2025-07-21
juju offer on non default model


crypticC0der:
juju.offer will always run on the default model (the one you get when you do jubilant.Juju()). This means for tests with multiple models or even just using the temp_model and then trying to offer an integration in there you get an error that the charm cannot be found

This can be reproduced easily with
```
import jubilant
j = jubilant.Juju()
j.add_model("default")
a = jubilant.Juju()
a.add_model("a")
a.deploy("self-signed-certificates")
a.wait(jubilant.all_active)
a.offer("self-signed-certificates", endpoint="certificates")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ubuntu/.local/lib/python3.12/site-packages/jubilant/_juju.py", line 604, in offer
    self.cli(*args, include_model=False)
  File "/home/ubuntu/.local/lib/python3.12/site-packages/jubilant/_juju.py", line 227, in cli
    stdout, _ = self._cli(*args, include_model=include_model, stdin=stdin)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/.local/lib/python3.12/site-packages/jubilant/_juju.py", line 247, in _cli
    raise CLIError(e.returncode, e.cmd, e.stdout, e.stderr) from None
jubilant._juju.CLIError: Command '['juju', 'offer', 'self-signed-certificates:certificates']' returned non-zero exit status 1.
Stderr:
ERROR getting offered application self-signed-certificates: application "self-signed-certificates" not found
```

crypticC0der:
my point was more that the model connected to the jubilant.Juju should be included in the juju.offer command and the default being what it is differs from the design pattern that the rest of the library follows.

dwilding:
I got hit by this today and I don't think the behaviour is sufficiently documented. I've opened #269 to explain it more clearly.

PR #269, opened 2026-03-13
docs: fix docstring of consume()

dwilding:
This PR:

- Fixes a copy-paste error (presumably) in the docstring of `consume()`
- ~Makes a couple of additions to the docstring of `offer()` to clarify that `self.model` is not used. See #170.~ We decided not to do this. See discussion below


tonyandrewmeyer (approved):
I just re-read #170, #148, and #82, and, combine with this PR, that makes me wonder if @benhoyt and I were wrong in #170. The original addition of `offer()` did (theoretically) have the more natural behaviour, but had a bug (#148), and lost it in that fix.

Perhaps we should really be behaving like pretty much all the other Jubilant methods and transparently handling the model unless explicitly provided, and not relying on the active `juju` model. Something like if the user doesn't provide a `model:app` style app, then we add `{self.model}:` to the provided app.

That would be a breaking change, though, and I'm not sure it could really be justified as a breaking bug fix, because it was an explicit design choice that was later confirmed when questioned. So that perhaps takes it off the table for now.

Assuming we'll stick with the current behaviour, then the PR works for me, with a couple of minor notes, thanks!

PR #272, opened 2026-03-18
feat: Allow metadata_source to be passed down to the CLI via the bootstrap command

moisesbenzan:
## Summary

Add support for passing Juju agent metadata source through `Juju.bootstrap()`.

## What changed
- Added optional `metadata_source: str | None` parameter to `Juju.bootstrap()`.
- When provided, it now appends `--metadata-source <value>` to the bootstrap CLI args.
- Updated the `bootstrap()` docstring to document the new parameter.
- Extended unit coverage in `tests/unit/test_bootstrap.py::test_all_args` to assert the new flag is passed correctly.

## Why
   
   This enables callers to bootstrap using a custom metadata endpoint (e.g., mirrors or controlled environments) without bypassing Jubilant’s API.

## Tests

- Updated unit test: `tests/unit/test_bootstrap.py`

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -365,7 +365,7 @@ class Juju:
                 args.extend(['--to', to])
             else:
                 args.extend(['--to', ','.join(to)])
-        if metadata_source:
+        if metadata_source is not None:
             args.extend(['--metadata-source', metadata_source])
 
         _, stderr = self._cli(*args, include_model=False)


moisesbenzan:
> I would prefer -- and it's more in keeping with the design of Jubilant -- that we add proper named arguments, like `metadata_source: str | None = None`. Are there any other bootstrap arguments you need?

Sure thing. For the moment, only `--metadata-source` is needed for Openstack based deployments in our testing.

I'll make the change and send another commit up

tonyandrewmeyer (approved):
A nice, simple, change -- makes sense to me.

Do you also want to re-review @benhoyt ?

jubilant/_juju.py
```diff
@@ -363,6 +365,8 @@ def bootstrap(
                 args.extend(['--to', to])
             else:
                 args.extend(['--to', ','.join(to)])
+        if metadata_source:
```
benhoyt: We've used `is not None` explicitly for similar args (eg: `credential`). I think we should do the same here:
```suggestion
        if metadata_source is not None:
```

benhoyt (approved):
Thanks. Nice and simple. One minor tweak requested.

Issue #266, opened 2026-03-06
Jubilant scp directory does not work with -r when using the Juju snap


rpbritton:
```
$ python3 -c "
import jubilant
juju = jubilant.Juju()
juju.scp('validators/', 'kratos/0:/var/lib/validators', scp_options=['-r'])
"
Traceback (most recent call last):
  File "<string>", line 4, in <module>
  File "/home/ubuntu/.cache/pypoetry/virtualenvs/charm-integration-testing-meta-10MWES9l-py3.10/lib/python3.10/site-packages/jubilant/_juju.py", line 1119, in scp
    shutil.copy(source, file_temp.name)
  File "/usr/lib/python3.10/shutil.py", line 417, in copy
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/lib/python3.10/shutil.py", line 254, in copyfile
    with open(src, 'rb') as fsrc:
IsADirectoryError: [Errno 21] Is a directory: 'validators/'
```

tonyandrewmeyer:
Fixed in #275

PR #275, opened 2026-03-21
fix: correct behaviour when transparently handling Juju's snap confin…

### LINKED ISSUE #273: local resources lose their file extension, which leads to deploy errors
## Description

Current `_deploy_tempdir` context manager does not keep file extensions, which lead to deploy errors in scenarios where file extension is checked by juju

## Steps to reproduce

Clone & build the test charm in the `kafka-connect-operator` repo

```bash
git clone https://github.com/canonical/kafka-connect-operator
cd kafka-connect-operator/tests/integration/source-integrator-charm/
charmcraft pack

juju add-model testing
```

Then:

```python
import os
import jubilant

cwd = os.getcwd()
os.system(f"tar -cf {cwd}/dummy.tar --files-from /dev/null")

juju = jubilant.Juju(model="testing")
juju.deploy("./source-integrator_ubuntu@22.04-amd64.charm", resources={"connect-plugin": f"{cwd}/dummy.tar"})
```

deploy will fail:

```
INFO:jubilant:cli: juju deploy --model testing /root/snap/juju/common/tmph5duo5_w/_temp.charm --resource connect-plugin=/root/snap/juju/common/tmph5duo5_w/connect-plugin

ERROR resource "connect-plugin": Put https://10.160.219.65:17070/model/2b170d3b-c2c2-4e9d-831b-fadd4b1b55ba/applications/source-integrator/resources/connect-plugin?pendingid=e69387da-c1b0-4e87-8710-5d7df35573fc: incorrect extension on resource upload "/root/snap/juju/common/tmph5duo5_w/connect-plugin", expected ".tar"
```

## Additional context

`_deploy_tempdir` should keep file extensions for resources: 

https://github.com/canonical/jubilant/blob/01b93d65a3277891cfa73499d20f7bf397d3f5f5/jubilant/_juju.py#L1497-L1528

tonyandrewmeyer:
This PR fixes a pair of issues with the Jubilant functionality that transparently handles confinement of the Juju snap (by copying to a confinement-accessible location).

Firstly, when using `scp`, only single files were handled properly, but it's also possible to `scp` a directory. The PR checks the type of the object at the given path and switches between copying a file and copying the directory as needed.

Note that machine models support copying directories, but K8s models only support copying files. If a user tries to copy a directory on a K8s model, they'll get an appropriate error from Juju (which will bubble up from Jubilant in the usual way).

Secondly, when deploying with local resources, our copying names the files based on the resource name (this avoids clashes), but Juju needs the files to have particular extensions in some cases. The PR keeps the existing naming system, but appends the extension of the original file.

For the integration tests, the `testapp` charm gains a new resource (unused other than to ensure that deploying with one works). It seemed better to re-use one of the existing charms than to add a third just for this, and these small charms are simpler than deploying something from Charmhub with a local resource.

For the unit tests, a new mock class `NamedTemporaryDirectory` is added, which is similar to the existing `NamedTemporaryFile`. It is constructed with an existing directory path (most likely from the real `TemporaryDirectory`) and can then be used instead of that, so that the path is known and can be used in asserts with `mocks.Run`.

This does *not* work for the remote side. I think we would need to add some sort of side-effect functionality to `mocks.Run` to handle that, and it seems cleaner to just have a separate mock for the two tests that require it.

Fixes #273, #266.


benhoyt:
@tonyandrewmeyer Was going to add someone else for review here. I'm okay with this change at a high level, let's just make sure it's well reviewed and tested so we don't accidentally make the issue worse.

james-garner-canonical (approved):
The changes look good to me. The integration test coverage looks good.

dimaqq (commented):
Weirdly my comments from last week (?) didn't get posted.
Sorry for the delay.

tests/integration/conftest.py
```diff
@@ -39,3 +41,13 @@ def model2(request: pytest.FixtureRequest) -> Generator[jubilant.Juju]:
...
+
+
+@pytest.fixture
+def empty_tar() -> Generator[str]:
+    """Pytest fixture that creates an empty .tar file and returns its path."""
+    with tempfile.TemporaryDirectory() as temp:
+        path = f'{temp}/empty.tar'
+        with tarfile.open(path, 'w'):
+            pass
+        yield path
```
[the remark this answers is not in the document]
tonyandrewmeyer: The extension is [validated by Juju](https://github.com/juju/juju/blob/c0f9fc7e2cf915661c620db02df025ed5746f651/apiserver/internal/handlers/resources/resources.go#L288), which is the cause of the original report. It has to match the filename that's in the metadata. So tar isn't really special, but we have to use something and it felt like matching the report was nice.

I believe an empty file with the `.tar` extension would also work; it seemed cleaner to make it a real tar file.

I could change this to `.txt` or something if preferred.

tests/unit/test_deploy.py
```diff
@@ -158,3 +158,44 @@ def mock_run(args: list[str], **_: Any):
...
+    with tempfile.TemporaryDirectory() as temp:
+        snap_dir = pathlib.Path(temp) / 'snap'
+        snap_dir.mkdir()
+        monkeypatch.setattr('tempfile.TemporaryDirectory', mocks.TemporaryDirectory(str(snap_dir)))
+
+        src_dir = pathlib.Path(temp) / 'src'
+        src_dir.mkdir()
+        (src_dir / 'my.charm').write_text('CH')
+        (src_dir / 'archive.tar').write_text('TAR')
+        (src_dir / 'rawfile').write_text('RAW')
```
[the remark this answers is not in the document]
tonyandrewmeyer: In the unit tests we don't have a charm that's already packed, so we would need to either commit a packed charm or subprocess out to packing one. It would also be more work to verify that the right files were copied -- we presumably wouldn't want to load all the content of the charm zip into memory, so we'd need to hash it and check the hash.

That seems considerably more complicated for not much benefit.

tests/unit/test_deploy.py
```diff
@@ -158,3 +158,44 @@ def mock_run(args: list[str], **_: Any):
         )
 
     assert num_calls == 1
+
+
+def test_tempdir_preserves_resource_extension(run: mocks.Run, monkeypatch: pytest.MonkeyPatch):
+    monkeypatch.setattr('shutil.which', lambda _: '/snap/bin/juju')  # type: ignore
+
+    with tempfile.TemporaryDirectory() as temp:
```
james-garner-canonical: Can we use the `tmp_path` fixture instead of this?

jubilant/_juju.py
```diff
@@ -1542,10 +1523,11 @@ def _deploy_tempdir(
...
 
-        with tempfile.TemporaryDirectory(dir=self._temp_dir) as temp_dir:
+        with tempfile.TemporaryDirectory(dir=self._temp_dir) as td:
+            temp_dir = pathlib.Path(td)
             if charm_needs_temp:
                 assert charm is not None
-                temp = os.path.join(temp_dir, '_temp.charm')
+                temp = str(temp_dir / '_temp.charm')
                 shutil.copy(charm, temp)
                 charm = temp
```
james-garner-canonical: Minor: seems slightly easier to follow -- sorry for the belated suggestion.
```suggestion
                temp = temp_dir / '_temp.charm'
                shutil.copy(charm, temp)
                charm = str(temp)
```

tests/unit/test_deploy.py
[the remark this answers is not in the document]
dimaqq: sure.

dimaqq (approved):

tests/unit/test_deploy.py
tonyandrewmeyer: We don't use it anywhere else in Jubilant's tests, but yes, we probably can.

james-garner-canonical: Thanks, this makes it a lot easier for me to follow where we're using our custom mock `TemporaryDirectory`.

PR #276, opened 2026-03-22
feat!: respect self.model in offer method

dwilding:
This PR changes the behaviour of `juju.offer('myapp', endpoint='foo')`. This call now uses the model from `juju.model`, if set. Currently, this call uses the model selected in the Juju CLI, which matches how `juju offer` works, but is surprising in the context of other Jubilant methods (see #170).

I'm changing the behaviour of `offer()` in the case that the `app` arg is a bare app name (not `mymodel.myapp`):
- If `controller` is not specified, we insert the model (and controller if specified) from `juju.model`.
- If `controller` is specified, we raise an error because the caller's intention is unclear. We raise regardless of whether `juju.model` is set.

**[Preview of updated docs](https://canonical-ubuntu-documentation-library--276.com.readthedocs.build/jubilant/reference/jubilant/#jubilant.Juju.offer)**

This change is technically a breaking change. However, I can't find any examples of affected code in the wild:
- Everyone who calls `juju.offer()` with a bare app name already has `juju.model` set to the model selected in the Juju CLI.
- Nobody calls `juju.offer()` with a bare app name and `controller` set.

Mainly, this change will let people simplify their code. For example:

- [jenkins-k8s-operator](https://github.com/canonical/jenkins-k8s-operator/blob/main/tests/integration/test_auth_proxy.py#L128-L129) - No need to insert `juju.model` into app name when calling `offer()`.
- [spark-k8s-bundle](https://github.com/canonical/spark-k8s-bundle/blob/f98baa529961414063950022823f31e4ce36c386/python/tests/integration/conftest.py#L388-L389) - No need to switch model before calling `offer()`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,6 +8,8 @@ on:
   workflow_call:
   workflow_dispatch:
 
+permissions: {}
+
 jobs:
   lint:
     runs-on: ubuntu-latest
@@ -15,9 +17,11 @@ jobs:
     steps:
       - name: Check out repo
         uses: actions/checkout@v4
+        with:
+          persist-credentials: false
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v5
+        uses: astral-sh/setup-uv@d4b2f3b6ecc6e67c4457f6d3e41ec42d3d0fcb86 # v5
 
       - name: Run linting and static type checking
         run: make lint
@@ -34,6 +38,8 @@ jobs:
     steps:
       - name: Check out repo
         uses: actions/checkout@v4
+        with:
+          persist-credentials: false
 
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v5
@@ -41,7 +47,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v5
+        uses: astral-sh/setup-uv@d4b2f3b6ecc6e67c4457f6d3e41ec42d3d0fcb86 # v5
 
       - name: Run unit tests
         run: make unit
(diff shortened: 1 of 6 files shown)


jubilant/_juju.py
```diff
@@ -863,24 +863,59 @@ def offer(
...
+                raise ValueError(
+                    'controller is specified but app does not include a dotted model name'
+                )
+                # This error message implies that we reject this usage regardless of self.model.
+                # That's intentional. If self.model is None, we allow this usage because it maps
+                # directly to the Juju CLI, but we discourage it (see the docstring).
         app_endpoint = f'{app}:{endpoint}'
         args = ['offer', app_endpoint]
-        if controller:
+        if controller is not None:
```
dimaqq: I'm not sure about this change, because Juju CLI kindof supports Go "null values":

```command
⋊> dima@bb ⋊> /c/hexanator on main ⨯ juju offer aa:a,b,c --controller ""                                                                                             09:23:20
ERROR no current model, use juju switch to select a model on which to operate
⋊> dima@bb ⋊> /c/hexanator on main ⨯ juju offer aa:a,b,c                                                                                                             09:23:29
ERROR no current model, use juju switch to select a model on which to operate
```

In the above, not passing "--controller" and passing it with an empty string has the same meaning!

I think perhaps we should forbid specifying an empty string controller in the API:


```suggestion
        if controller == "": raise ValueError(...)
        if controller is not None:
```

Or, perhaps more defensively:

```suggestion
        if not controller and controller is not None: raise ValueError(f'bad {controller=}')
        if controller is not None:
```

james-garner-canonical: > In the above, not passing "--controller" and passing it with an empty string has the same meaning!

Given this, how about we just stick with the logic on `main` of
```py
if controller:
```
We can always change to pass `--controller '"` in future if that acquires Juju meaning.

dwilding: Resolved in my latest commit

jubilant/_juju.py
```diff
@@ -863,21 +863,49 @@ def offer(
...
 
         Args:
-            app: Application name to offer endpoints for. May include a dotted model name, for
-                example ``mymodel.mysql``.
-            controller: Name of controller to operate in. If not specified, use the current
-                controller.
+            app: Application name to offer endpoints for. If *app* includes a dotted model name,
+                this method ignores ``self.model`` when determining the model and controller.
+            controller: Name of controller to operate in. Only specify *controller* if *app*
+                includes a dotted model name.
```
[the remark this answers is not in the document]
dwilding: You mean the asterisks? That matches elsewhere (per my other comment)

PR #282, opened 2026-03-27
test: don't rely on Juju.secrets() order

dimaqq:
I don't know if `juju secrets` output was ever intended to be in a stable order.
I guess we got stable order before, because this test somehow worked so far.
Now it's failing in an unrelated PR: https://github.com/canonical/jubilant/actions/runs/23633104321/job/68836635117?pr=276

### DIFF
--- a/tests/integration/test_secrets.py
+++ b/tests/integration/test_secrets.py
@@ -33,7 +33,7 @@ def test_update_secret(juju: jubilant.Juju):
 
 
 def test_get_all_secrets(juju: jubilant.Juju):
-    secrets = sorted(juju.secrets(), key=lambda s: s.name)
+    secrets = sorted(juju.secrets(), key=lambda s: s.name or s.uri)
     assert len(secrets) == 2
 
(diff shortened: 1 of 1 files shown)


tests/integration/test_secrets.py
```diff
@@ -33,7 +33,7 @@ def test_update_secret(juju: jubilant.Juju):
 
 
 def test_get_all_secrets(juju: jubilant.Juju):
-    secrets = juju.secrets()
+    secrets = sorted(juju.secrets(), key=lambda s: s.name or s.uri)
```
dimaqq: Alternatively this could be:

```suggestion
    secrets = juju.secrets()
    # These two secrets always have a name, s.uri is to keep linter happy
    secrets.sorted(key=lambda s: s.name or s.uri)
```

Or

```suggestion
    secrets = juju.secrets()
    secrets.sort(key=lambda s: s.name)  # type: ignore  # we have names
```

dimaqq:
Closing in favour of Tony's approach

PR #283, opened 2026-03-28
feat: implement add_cloud and update_cloud methods

### LINKED ISSUE #156: Add juju add-cloud and update-cloud commands
Juju CLI accepts clouds as files `juju add-cloud -f <file.yaml>`, what would be the jubilant way of doing the same thing?

This is used in two places:
- sunbeam snap (sets up the stack)
- integration tests across heterogenous clouds

Tony-WLB:
Add support for `add_cloud` and `update_cloud` methods in the `Juju` class, in line with the `juju add-cloud/update-cloud` commands.

Fixes #156

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -106,6 +106,82 @@ class Juju:
     # about where to put each new method.
 
     @overload
+    def add_cloud(
+        self,
+        cloud: str,
+        definition: str | pathlib.Path | Mapping[str, Any],
+        *,
+        client: Literal[True],
+        controller: str | None = None,
+        credential: str | None = None,
+        force: bool = False,
+        target_controller: str | None = None,
+    ) -> None: ...
+
+    @overload
+    def add_cloud(
+        self,
+        cloud: str,
+        definition: str | pathlib.Path | Mapping[str, Any],
+        *,
+        client: bool = False,
+        controller: str,
+        credential: str | None = None,
+        force: bool = False,
+        target_controller: str | None = None,
+    ) -> None: ...
+
+    def add_cloud(
+        self,
+        cloud: str,
+        definition: str | pathlib.Path | Mapping[str, Any],
+        *,
+        client: bool = False,
+        controller: str | None = None,
+        credential: str | None = None,
+        force: bool = False,
+        target_controller: str | None = None,
+    ) -> None:
+        """Add a cloud definition to a controller or to a client (or both).
+
+        Args:
+            cloud: Name for the cloud.
+            definition: A YAML file or a mapping containing the cloud definition.
+            client: If true, save the cloud to the client for use with future controllers.
+                You must specify ``client=True`` or provide *controller* (or both).
+            controller: If specified, save the cloud to the named controller.
+            credential: Credential to use for the new cloud.
+            force: If true, force add cloud to the controller, bypassing checks.
+            target_controller: Name of a JAAS managed controller to add the cloud to.
+        """
+        if not client and controller is None:
+            raise TypeError('"client" must be true or "controller" must be specified (or both)')
+
+        args = ['add-cloud', cloud]
+
+        if client:
+            args.append('--client')
+        if controller is not None:
+            args.extend(['--controller', controller])
+        if credential is not None:
+            args.extend(['--credential', credential])
+        if force:
+            args.append('--force')
+        if target_controller is not None:
+            args.extend(['--target-controller', target_controller])
+
+        if isinstance(definition, (str, pathlib.Path)):
+            with self._path_for_juju(definition) as definition_path:
+                args.extend(['--file', definition_path])
+                self.cli(*args, include_model=False)
+        else:
+            with tempfile.NamedTemporaryFile('w+', dir=self._temp_dir) as temp_file:
+                _yaml.safe_dump(definition, temp_file)
+                temp_file.flush()
+                args.extend(['--file', temp_file.name])
+                self.cli(*args, include_model=False)
+
+    @overload
     def add_credential(
         self,
         cloud: str,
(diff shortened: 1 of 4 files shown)


tests/integration/test_cloud.py
```diff
@@ -0,0 +1,107 @@
...
+    'unrecognized command',
+    'not implemented',
+    'permission denied',
+)
+
+
+def _skip_if_unsupported(exc: jubilant.CLIError) -> None:
+    message = str(exc).lower()
+    if any(token in message for token in UNSUPPORTED_TOKENS):
+        pytest.skip('cloud management commands are not supported in this environment')
```
benhoyt: When does this happen?

jubilant/_juju.py
```diff
@@ -105,6 +105,55 @@ def __repr__(self) -> str:
...
+        target_controller: str | None = None,
+    ) -> None:
+        """Add a cloud definition to Juju.
+
+        Args:
+            cloud: Name for the cloud.
+            definition: Path to a YAML file containing the cloud definition, or a mapping
+                containing the parsed cloud definition YAML.
+            client: If true, save the cloud to the client for use in future controllers.
+            controller: If specified, save the cloud to the named controller.
```
benhoyt: From the TypeError below it looks like you need to specify one of `--client` or `--controller`)? I think we should probably say that here, something like:
```suggestion
            controller: If specified, save the cloud to the named controller. You must
                specify either ``client=True`` or provide a *controller*.
```

tests/integration/test_cloud.py
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    _add_cloud_or_skip(juju, cloud_name, cloud_definition)
+
+    try:
+        show_cloud = _show_cloud(juju, cloud_name)
+        assert endpoint in show_cloud
+    finally:
+        _remove_cloud_if_present(juju, cloud_name)
```
benhoyt: I'd prefer we try to avoid the "or skip" and "if present" as much as possible, to have simple, linear test flow. If we need to skip a test (on machine vs K8s) we should do it at the top of the test, and have the rest of the test run simply.

tests/integration/test_cloud.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def _skip_if_unsupported(exc: jubilant.CLIError) -> None:
+    message = str(exc).lower()
+    if any(token in message for token in UNSUPPORTED_TOKENS):
+        pytest.skip('cloud management commands are not supported in this environment')
+
+
+def _remove_cloud_if_present(juju: jubilant.Juju, cloud_name: str) -> None:
+    with contextlib.suppress(jubilant.CLIError):
```
benhoyt: Can we avoid suppressing the exception, and just remove the cloud if we've created it? See also below.

tests/unit/test_add_cloud.py
```diff
@@ -0,0 +1,72 @@
...
+        credential='mycred',
+        force=True,
+        target_controller='target-ctrl',
+    )
+
+
+def test_requires_client_or_controller():
+    juju = jubilant.Juju()
+
+    with pytest.raises(TypeError, match='"client" must be true or "controller" must be specified'):
```
benhoyt: Let's avoid testing the specific error message in the test, so it's easier to tweak later (we don't guarantee the message content as part of the API).

benhoyt (commented):
Thanks for this, it's nice clean Python, and it looks like a good start! I've added some comments (mostly about the docstrings and the integration tests).

tests/integration/test_cloud.py
Tony-WLB: Done

jubilant/_juju.py
```diff
@@ -1377,6 +1399,44 @@ def trust(
...
+
+        args = ['update-cloud', cloud]
+
+        if client:
+            args.append('--client')
+        if controller is not None:
+            args.extend(['--controller', controller])
+
+        if isinstance(definition, (str, pathlib.Path)):
+            args.extend(['-f', str(definition)])
```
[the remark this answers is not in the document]
Tony-WLB: I have also included a link to the Juju code. But I am not sure whether we should use the permalink in the comment here.

tests/unit/test_add_cloud.py
```diff
@@ -0,0 +1,72 @@
...
+import jubilant
+
+from . import mocks
+
+
+def test_with_path_str(run: mocks.Run):
+    run.handle(['juju', 'add-cloud', 'mycloud', '--client', '--file', '/path/to/cloud.yaml'])
+    juju = jubilant.Juju()
+
+    juju.add_cloud('mycloud', '/path/to/cloud.yaml', client=True)
```
tonyandrewmeyer: All of the unit tests, also for update-cloud, have `client=True`. I don't think we need to have variations of them all with `client=False`, but I do think we should have at least one (for each of add- and update-) that has `client=False`, and perhaps also that uses the default value for `client`.

jubilant/_juju.py
```diff
@@ -105,6 +105,82 @@ def __repr__(self) -> str:
...
+        credential: str | None = None,
+        force: bool = False,
+        target_controller: str | None = None,
+    ) -> None:
+        """Add a cloud definition to Juju.
+
+        Args:
+            cloud: Name for the cloud.
+            definition: Path to a YAML file containing the cloud definition, or a mapping
+                containing the parsed cloud definition YAML.
```
tonyandrewmeyer: It's not necessarily the parsed YAML, it's just the definition.

```suggestion
            definition: a YAML file or mapping containing the cloud definition.
```

tonyandrewmeyer (approved):
I wonder if we should also be adding clouds, remove-cloud, and update-credential as the "set" of cloud management functionality. It seems like you might want to check `clouds` first and only if the one you want isn't there do `add-cloud` (look before you leap) but perhaps just accepting an error (better to ask forgiveness) is fine. Similarly, it seems like maybe you'd want to explicitly clean up and so need `remove-cloud`. (You even see this a bit in the need to use `cli()` in the integration test. I'm also fine just leaving these until someone asks for them.

Everything looks good to me, and worked with my light local testing (and seems good based on the integration tests, particularly for 3 & 4). I've added a few minor comments, but I'm fine assuming those will be resolved in some way.

tests/unit/test_add_cloud.py
Tony-WLB: Tests with default client (False) added.

jubilant/_juju.py
```diff
@@ -105,6 +105,56 @@ def __repr__(self) -> str:
...
+        self,
+        cloud: str,
+        definition: str | pathlib.Path | Mapping[str, Any],
+        *,
+        client: bool = False,
+        controller: str | None = None,
+        credential: str | None = None,
+        force: bool = False,
+        target_controller: str | None = None,
+    ) -> None:
```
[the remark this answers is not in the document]
benhoyt: Okay, let's leave here for now.

PR #287, opened 2026-04-01
docs: restructure homepage and update intro to Jubilant

dwilding:
This PR updates the homepage of the Jubilant docs to use a similar style to [Ops](https://documentation.ubuntu.com/jubilant/).

Although in fact the most significant change is that I've reworked the intro - especially the advice about charm integration testing, as we now want to recommend `pytest-jubilant` and direct people to the Ops docs. Given that, it would be best for https://github.com/canonical/operator/pull/2406 to merge before this PR.

**[Preview of updated homepage](https://canonical-ubuntu-documentation-library--287.com.readthedocs.build/jubilant/)**

I decided that since we have so few doc pages, it's overkill to have "In this documentation" with an itemised table plus "How this documentation is organised". I tried it, but it looked simultaneously too sparse and too noisy with repeated links. For now, I think a simple "In this documentation" list is better.

---

Merge this after https://github.com/canonical/operator/pull/2406

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -24,7 +24,7 @@ You should switch to Jubilant if your integration tests currently use [pytest-op
 
 ## In this documentation
 
-- [Getting started with Jubilant](tutorial/getting-started)
+- [Tutorial: Getting started with Jubilant](tutorial/getting-started)
 - [Design goals](explanation/design-goals)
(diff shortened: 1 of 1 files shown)


james-garner-canonical (approved):
Looks good to me!

PR #301, opened 2026-04-13
ci: run CI (especially integration tests) every Monday morning

### LINKED ISSUE #265: We should run our tests whenever Juju is updated
This will let us see when things break, like the 4.0.2 breakage detected in #264

The simplest way to handle this is probably to run on a schedule (perhaps daily or weekly).

james-garner-canonical:
This PR updates CI to run on a schedule, every Monday morning. The goal is to ensure that we're running against new Juju releases in a relatively timely manner, so we don't run into breakages in the middle of other PRs (as we did in #264). The current [Juju release cadence](https://documentation.ubuntu.com/juju/latest/releasenotes/juju_4.0.x/) appears to be roughly monthly, but is very variable. Resolves #265.

This PR also removes the now-unnecessary workaround step for the resolved `concierge` issue https://github.com/canonical/concierge/issues/30

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,7 +20,7 @@ on:
           - edge
   schedule:
     # min (0-59) / hour (0-23) / day (1-31) / month (1-12) / weekday (SUN-SAT)
-    - cron: "0 10 * * MON"
+    - cron: "43 6 * * MON"
       timezone: "Pacific/Auckland"
 
 permissions: {}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -67,7 +67,7 @@ jobs:
       - name: Run unit tests
         run: make unit
 
-  init:
+  init-integration:
     runs-on: ubuntu-latest
 
     outputs:
@@ -87,13 +87,13 @@ jobs:
           echo "juju-channels=[\"3/$RISK\", \"4/$RISK\"]" >> "$GITHUB_OUTPUT"
 
   integration-k8s:
-    needs: init
+    needs: init-integration
     runs-on: ubuntu-latest
 
     strategy:
       fail-fast: false
       matrix:
-        juju-channel: ${{ fromJSON(needs.init.outputs.juju-channels) }}
+        juju-channel: ${{ fromJSON(needs.init-integration.outputs.juju-channels) }}
 
     concurrency:
       group: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.juju-channel }}-k8s
@@ -121,13 +121,13 @@ jobs:
         run: make integration-k8s
 
   integration-machine:
-    needs: init
+    needs: init-integration
     runs-on: ubuntu-latest
 
     strategy:
       fail-fast: false
       matrix:
-        juju-channel: ${{ fromJSON(needs.init.outputs.juju-channels) }}
+        juju-channel: ${{ fromJSON(needs.init-integration.outputs.juju-channels) }}
 
     concurrency:
       group: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.juju-channel }}-machine
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -42,7 +42,7 @@ jobs:
         run: make lint
 
   unit:
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
 
     strategy:
       fail-fast: false
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,7 +17,6 @@ on:
           - stable
           - candidate
           - beta
-          - edge
   schedule:
     # min (0-59) / hour (0-23) / day (1-31) / month (1-12) / weekday (SUN-SAT)
     - cron: "43 6 * * MON"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -42,7 +42,7 @@ jobs:
         run: make lint
 
   unit:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
 
     strategy:
       fail-fast: false
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -69,11 +69,6 @@ jobs:
       cancel-in-progress: true
 
     steps:
-      - name: Work around https://github.com/canonical/concierge/issues/30
-        run: |
-          sudo apt-get remove -y docker-ce docker-ce-cli containerd.io
-          sudo rm -rf /run/containerd
-
       - name: Check out repo
         uses: actions/checkout@v6
         with:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,6 +7,10 @@ on:
   pull_request:
   workflow_call:
   workflow_dispatch:
+  schedule:
+    # min (0-59) / hour (0-23) / day (1-31) / month (1-12) / weekday (SUN-SAT)
+    - cron: "0 10 * * MON"
+      timezone: "Pacific/Auckland"
 
 permissions: {}
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -153,3 +153,21 @@ jobs:
 
       - name: Run integration tests
         run: make integration-machine
+
+  open-issue-on-failure-if-scheduled:
+    if: ${{ failure() && github.event_name == 'schedule' }}
+    needs: [integration-k8s, integration-machine]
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -65,11 +69,6 @@ jobs:
       cancel-in-progress: true
 
     steps:
-      - name: Work around https://github.com/canonical/concierge/issues/30
-        run: |
-          sudo apt-get remove -y docker-ce docker-ce-cli containerd.io
-          sudo rm -rf /run/containerd
-
```
tonyandrewmeyer, line 72: Nice, thanks!

.github/workflows/ci.yaml
```diff
@@ -7,6 +7,10 @@ on:
   pull_request:
   workflow_call:
   workflow_dispatch:
+  schedule:
+    # min (0-59) / hour (0-23) / day (1-31) / month (1-12) / weekday (SUN-SAT)
+    - cron: "0 10 * * MON"
+      timezone: "Pacific/Auckland"
```
tonyandrewmeyer, line 23: TIL that you can put a timezone here.

.github/workflows/ci.yaml
```diff
@@ -7,6 +7,10 @@ on:
   pull_request:
   workflow_call:
   workflow_dispatch:
+  schedule:
+    # min (0-59) / hour (0-23) / day (1-31) / month (1-12) / weekday (SUN-SAT)
+    - cron: "0 10 * * MON"
```
tonyandrewmeyer: What about doing this before any of us would start? 7 a.m. or earlier, for example. I think we also try to avoid `0` as a minute to try to spread work out on GitHub runners (but if you tell me you rolled a d60 and got a 0, fine with me).

tonyandrewmeyer (approved):
I feel like it would be nice if this was triggered by a new version, or earlier-exited if there wasn't a new version, but I guess we don't really have a mechanism for either of those (at least without having some sort of state storage).

It seems like it would be nicer to be testing non-stable (candidate? beta?) so that we could report issues to the Juju team before there was a stable release. What about something like this? (via Claude):

```yaml
config:
    runs-on: ubuntu-latest
    outputs:
      channels: ${{ steps.set.outputs.channels }}
    steps:
      - id: set
        run: |
          if [ "${{ github.event_name }}" = "schedule" ]; then
            echo 'channels=["3/beta", "4/beta"]' >> "$GITHUB_OUTPUT"
          else
            echo 'channels=["3/stable", "4/stable"]' >> "$GITHUB_OUTPUT"
          fi

  build:
    needs: config
    strategy:
      matrix:
        channel: ${{ fromJson(needs.config.outputs.channels) }}
```

I think it's fine to have `lint` and `unit` run - they're both very fast so consume minimal resources and should always be passing anyway.

I'm also fine with weekly, since I assume this will very rarely fail, so usually be silent.

I don't recall exactly who gets notified when there's a failure in a scheduled action. Is it anyone with a particular access level (lower than admin?) or something that people subscribe to? If this only goes to admins, for example, then I think some sort of explicit failure notification might be good. But if it's going to (for example) all Charm Tech, then that seems fine.

.github/workflows/ci.yaml
james-garner-canonical: Good ideas:
```suggestion
    - cron: "43 6 * * MON"
```

james-garner-canonical:
> I feel like it would be nice if this was triggered by a new version, or earlier-exited if there wasn't a new version, but I guess we don't really have a mechanism for either of those (at least without having some sort of state storage).

Yeah, agreed.

> It seems like it would be nicer to be testing non-stable (candidate? beta?) so that we could report issues to the Juju team before there was a stable release. What about something like this? (via Claude):
> 
> ```yaml
> config:
>     runs-on: ubuntu-latest
>     outputs:
>       channels: ${{ steps.set.outputs.channels }}
>     steps:
>       - id: set
>         run: |
>           if [ "${{ github.event_name }}" = "schedule" ]; then
>             echo 'channels=["3/beta", "4/beta"]' >> "$GITHUB_OUTPUT"
>           else
>             echo 'channels=["3/stable", "4/stable"]' >> "$GITHUB_OUTPUT"
>           fi
> 
>   build:
>     needs: config
>     strategy:
>       matrix:
>         channel: ${{ fromJson(needs.config.outputs.channels) }}
> ```

Good idea, though I think we probably then want to be able to rerun with the beta releases on `workflow_dispatch` somehow too. Would you be on board with extending `workflow_dispatch` with an input that chooses between `stable` and `beta`, for example?

> I think it's fine to have `lint` and `unit` run - they're both very fast so consume minimal resources and should always be passing anyway.

:+1: 

> I'm also fine with weekly, since I assume this will very rarely fail, so usually be silent.

Yeah I figured you wouldn't want any more frequent than that.

> I don't recall exactly who gets notified when there's a failure in a scheduled action. Is it anyone with a particular access level (lower than admin?) or something that people subscribe to? If this only goes to admins, for example, then I think some sort of explicit failure notification might be good. But if it's going to (for example) all Charm Tech, then that seems fine.

Hilariously, according to [the docs](https://docs.github.com/en/actions/concepts/workflows-and-actions/notifications-for-workflow-runs):

> Notifications for scheduled workflows are sent to the user who initially created the workflow.
> - If a different user updates the `cron` syntax, in the schedule event in the workflow file, subsequent notifications will be sent to that user instead.
> - If a scheduled workflow is disabled and then re-enabled, notifications will be sent to the user who re-enabled the workflow rather than the user who last modified the cron syntax.

tonyandrewmeyer:
> Good idea, though I think we probably then want to be able to rerun with the beta releases on `workflow_dispatch` somehow too. Would you be on board with extending `workflow_dispatch` with an input that chooses between `stable` and `beta`, for example?

Sure.

> Hilariously, according to [the docs](https://docs.github.com/en/actions/concepts/workflows-and-actions/notifications-for-workflow-runs):
> 
> > Notifications for scheduled workflows are sent to the user who initially created the workflow.

Oh, that seems terrible. A shame we don't have a team email any more. Posting to Matrix or MM is maybe too much work? Any ideas for something other than "whoever last touched the schedule is it"?

james-garner-canonical:
> > Hilariously, according to [the docs](https://docs.github.com/en/actions/concepts/workflows-and-actions/notifications-for-workflow-runs):
> > > Notifications for scheduled workflows are sent to the user who initially created the workflow.
> 
> Oh, that seems terrible. A shame we don't have a team email any more. Posting to Matrix or MM is maybe too much work? Any ideas for something other than "whoever last touched the schedule is it"?

Looks like we should be able to [create an issue using `gh` in CI](https://docs.github.com/en/actions/tutorials/manage-your-work/schedule-issue-creation), how about we do that if we fail and it was a scheduled run?

james-garner-canonical:
I've made the changes we discussed. I'm going to try to test the issue creation by making some changes in a branch in my fork to open issues on workflow dispatch failures, and fast fail the integration tests for testing purposes.

james-garner-canonical:
> I've made the changes we discussed. I'm going to try to test the issue creation by making some changes in a branch in my fork to open issues on workflow dispatch failures, and fast fail the integration tests for testing purposes.

I've tested issue creation in my fork: [workflow with fast failing tests and issue creation on workflow_dispatch](https://github.com/james-garner-canonical/jubilant/blob/26-04%2Bci%2Btest-scheduled-tests-issue-creation/.github/workflows/ci.yaml), [successfully created issue](https://github.com/james-garner-canonical/jubilant/issues/1).

.github/workflows/ci.yaml
```diff
@@ -7,6 +7,21 @@ on:
...
+      channel:
+        description: "Juju channel risk (scheduled runs use beta, everything else uses stable)"
+        required: true
+        type: choice
+        default: stable
+        options:
+          - stable
+          - candidate
+          - beta
+          - edge
```
tonyandrewmeyer: I wonder if we should drop this. My understanding is that images aren't published for edge, so you can't just use it like you can with the other levels.

tonyandrewmeyer (approved):
Nice improvements, thanks!

I wonder if we'll want to adjust the failure issue details at some point, but I think that can be left until we actually get some.

.github/workflows/ci.yaml
james-garner-canonical: Is your suggestion just to drop edge?
```suggestion
```

tonyandrewmeyer: Yes, unless I'm remembering this incorrectly. I think it just won't work in this situation.

james-garner-canonical: Sounds good, will do. Wasn't sure if you were proposing more changes due to how the context showed up for me.

james-garner-canonical:
Would have been great if I had merged this last week so we would have seen the first scheduled run this morning, but I guess we'll see it next Monday instead.

PR #303, opened 2026-04-15
ci: pin the versions of zizmor and codespell when run by pre-commit

tonyandrewmeyer:
Pinning the versions that pre-commit uses protects those that use pre-commit from running versions that find different issues to our regular CI, and also against installing versions that haven't been vetted yet.


tonyandrewmeyer (commented):
Discussion with Ben, maybe we should just remove spellcheck since it perhaps isn't providing enough value for the cost of having a dependency. Double-check whether we still use it in any other Charm Tech repos.

dimaqq (approved):
main needs to be merged into this branch, maybe we'd re-lock at the same time.

.pre-commit-config.yaml
```diff
@@ -62,4 +62,3 @@ repos:
        types: [text]
        args: []
        require_serial: true
-       additional_dependencies: [tomli]
```
dimaqq, line 65: Nice! Happy to see it gone.

tonyandrewmeyer:
> maybe we should just remove spellcheck since it perhaps isn't providing enough value for the cost of having a dependency. Double-check whether we still use it in any other Charm Tech repos.

I was mis-remembering things here. Ops still has it and so do the charmcraft profiles, so for consistency it should probably stay for now, and we could reconsider more holistically in the future if we wanted to drop a dependency.

(It still seems to be that it's unlikely that it'll find anything that pyright or unit tests or human review or agent review wouldn't find.)

PR #305, opened 2026-04-17
feat: add add-machine command

Snehil-Shah:
Wraps `juju add-machine` CLI command.

Resolves #291

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -246,6 +246,65 @@ class Juju:
                 args.extend(['--file', temp_file.name])
                 self.cli(*args, include_model=False)
 
+    def add_machine(
+        self,
+        to: str | None = None,
+        *,
+        base: str | None = None,
+        constraints: Mapping[str, ConstraintValue] | None = None,
+        disks: str | Iterable[str] | None = None,
+        num_machines: int = 1,
+        private_key: str | pathlib.Path | None = None,
+        public_key: str | pathlib.Path | None = None,
+    ) -> None:
+        """Provision a new machine or assign one to the model.
+
+        Args:
+            to: Machine or container to allocate the new machine on. For example, to
+                allocate a new LXD container on machine 25, use ``lxd:25``. Unavailable
+                in Kubernetes clouds.
+            base: Operating system base to install on the new machine(s), for example,
+                ``'ubuntu@22.04'``.
+            constraints: Machine constraints to overwrite those available from
+                :meth:`model_constraints` and provider's defaults, for example,
+                ``{'mem': '8G', 'cores': 4}``.
+            disks: Storage directives for disks to attach to the machine(s), for
+                example, ``'ebs,1T,2'``.
+            num_machines: Number of machines to add.
+            private_key: Path to the private key to use during the connection.
+            public_key: Path to the public key to add to the remote authorized keys.
+        """
+        args = ['add-machine']
+        if to is not None:
+            args.append(to)
+        if base is not None:
+            args.extend(['--base', base])
+        if constraints is not None:
+            for k, v in constraints.items():
+                args.extend(['--constraints', _format_config(k, v)])
+        if disks is not None:
+            if isinstance(disks, str):
+                args.extend(['--disks', disks])
+            else:
+                args.extend(['--disks', ' '.join(disks)])
+        if num_machines != 1:
+            args.extend(['-n', str(num_machines)])
+
+        private_ctx = (
+            self._path_for_juju(private_key)
+            if private_key is not None
+            else contextlib.nullcontext()
+        )
+        public_ctx = (
+            self._path_for_juju(public_key) if public_key is not None else contextlib.nullcontext()
+        )
+        with private_ctx as private_path, public_ctx as public_path:
+            if private_path is not None:
+                args.extend(['--private-key', private_path])
+            if public_path is not None:
+                args.extend(['--public-key', public_path])
+            self.cli(*args)
+
     def add_model(
         self,
         model: str,
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -248,7 +248,7 @@ class Juju:
 
     def add_machine(
         self,
-        address_or_to: str | None = None,
+        target: str | None = None,
         *,
         base: str | None = None,
         constraints: Mapping[str, ConstraintValue] | None = None,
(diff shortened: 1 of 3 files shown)


tests/unit/test_add_machine.py
```diff
@@ -0,0 +1,103 @@
...
+    juju = jubilant.Juju()
+
+    juju.add_machine('ssh:user@10.10.0.3')
+
+
+def test_to_provider(run: mocks.Run):
+    run.handle(['juju', 'add-machine', 'host.internal'])
+    juju = jubilant.Juju()
+
+    juju.add_machine('host.internal')
```
Snehil-Shah, line 47: @benhoyt These three tests effectively test the same string argument handling and are technically redundant, but I still added them for visibility as per the docs [examples](https://documentation.ubuntu.com/juju/3.6/reference/juju-cli/list-of-juju-cli-commands/add-machine/#examples). Tell me if you want them removed.

jubilant/_juju.py
```diff
@@ -246,6 +246,65 @@ def add_credential(
...
+        )
+        public_ctx = (
+            self._path_for_juju(public_key) if public_key is not None else contextlib.nullcontext()
+        )
+        with private_ctx as private_path, public_ctx as public_path:
+            if private_path is not None:
+                args.extend(['--private-key', private_path])
+            if public_path is not None:
+                args.extend(['--public-key', public_path])
+            self.cli(*args)
```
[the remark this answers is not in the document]
benhoyt: We've discussed elsewhere and aren't going to do this - resolving.

jubilant/_juju.py
```diff
@@ -246,6 +246,65 @@ def add_credential(
...
+        if constraints is not None:
+            for k, v in constraints.items():
+                args.extend(['--constraints', _format_config(k, v)])
+        if disks is not None:
+            if isinstance(disks, str):
+                args.extend(['--disks', disks])
+            else:
+                args.extend(['--disks', ' '.join(disks)])
+        if num_machines != 1:
+            args.extend(['-n', str(num_machines)])
```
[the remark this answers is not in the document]
benhoyt: `juju add-machine` doesn't seem to support `--num-machines`, only `n`, so we'll leave this as is.

jubilant/_juju.py
```diff
@@ -246,6 +246,65 @@ def add_credential(
                 args.extend(['--file', temp_file.name])
                 self.cli(*args, include_model=False)
 
+    def add_machine(
+        self,
+        to: str | None = None,
```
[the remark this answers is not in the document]
benhoyt: I don't think this is a plain machine ID, but more like `lxd:4`, so I don't think `int` applies.

jubilant/_juju.py
```diff
@@ -246,6 +246,65 @@ def add_credential(
...
+                ``{'mem': '8G', 'cores': 4}``.
+            disks: Storage directives for disks to attach to the machine(s), for
+                example, ``'ebs,1T,2'``.
+            num_machines: Number of machines to add.
+            private_key: Path to the private key to use during the connection.
+            public_key: Path to the public key to add to the remote authorized keys.
+        """
+        args = ['add-machine']
+        if to is not None:
+            args.append(to)
```
dwilding:
```suggestion
        if address_or_to is not None:
            args.append(address_or_to)
```

jubilant/_juju.py
```diff
@@ -246,6 +246,65 @@ def add_credential(
...
+        num_machines: int = 1,
+        private_key: str | pathlib.Path | None = None,
+        public_key: str | pathlib.Path | None = None,
+    ) -> None:
+        """Provision a new machine or assign one to the model.
+
+        Args:
+            to: Machine or container to allocate the new machine on. For example, to
+                allocate a new LXD container on machine 25, use ``lxd:25``. Unavailable
+                in Kubernetes clouds.
```
dwilding: I think the name `to` is good for consistency with elsewhere in the API, but Juju's design of `add-machine` packs even more into this arg. I recommend expanding the name and description. Something like this:


```suggestion
            address_or_to: Address of a network-accessible computer to allocate as a machine, or a
                placement directive for a new machine. Examples: ``ssh:user@10.10.0.3`` allocates
                the specified computer. ``lxd:25`` allocates a new LXD container on machine 25.
                ``lxd`` allocates a new LXD container on a new machine instance, resulting in two
                additional machines. If not specified, a new machine instance is allocated.
```

jubilant/_juju.py
```diff
@@ -246,6 +246,65 @@ def add_credential(
...
+        to: str | None = None,
+        *,
+        base: str | None = None,
+        constraints: Mapping[str, ConstraintValue] | None = None,
+        disks: str | Iterable[str] | None = None,
+        num_machines: int = 1,
+        private_key: str | pathlib.Path | None = None,
+        public_key: str | pathlib.Path | None = None,
+    ) -> None:
+        """Provision a new machine or assign one to the model.
```
dwilding: I might make the summary more generic and let the first arg description cover the different behaviours. Something like:


```suggestion
        """Allocate a machine to the model. Unavailable in Kubernetes clouds.
```

tests/unit/test_add_machine.py
dwilding: I appreciated the extra tests, even though the code doesn't poke into the string. I think it would be good to also add a test for a bare `lxd` arg.

dwilding (changes requested):
This looks great overall - thanks a lot.

Snehil-Shah:
@dwilding Thanks for the review! I have updated the nomenclatures and added the test.

jubilant/_juju.py
```diff
@@ -246,6 +246,67 @@ def add_credential(
                 args.extend(['--file', temp_file.name])
                 self.cli(*args, include_model=False)
 
+    def add_machine(
+        self,
+        address_or_to: str | None = None,
```
benhoyt: @dwilding I realise you proposed this name in the first place, but I'm not in love with this name -- seems unwieldy and unnecessarily long. `to` is common for this in other similar functions (including `add_unit`). We'd add a machine *to* an address as well, so `to` works for both cases. In addition, what if there are other things allowed, should we say `address_or_lxd_machine_name_or_other_thing`?

dwilding: I disagree that this is semantically the same as in other places. We're either adding a machine _from_ an address, or adding a container _to_ an existing machine instance. This is strictly more functionality than in other places we use the name `to`. Really, we shouldn't bundle both behaviours into one arg - but splitting the arg might be diverging too much from the Juju CLI. Point taken about the length of the name, but I quite dislike `to` as the name. I'll have a think about shorter names that could work instead.

benhoyt: Fair. What about `target`?

dwilding (changes requested):
One last change @Snehil-Shah - thanks for your patience and work on this! After discussion with Ben, let's go with `target` instead of `address_or_to` as the argument name.

Snehil-Shah:
@dwilding Done.

benhoyt (approved):
I'm happy with this, thanks @Snehil-Shah.

@dwilding Could you please review+approve or make any final tweaks yourself and go ahead and merge?

dwilding (approved):
Great, thanks again! Merging now

PR #308, opened 2026-04-27
feat: add show_unit method

Guillaumebeuzeboc:
This PR adds the `show_unit` method, corresponding to the `juju show-unit` command. It also adds a new UnitInfo type and `unittypes.py` module (code generated from Juju structs like we did with the other Juju types).

This is useful to inspect the content of relations, for example.

It only supports a single unit name, for simplicity, like other similar Jubilant methods. And doesn't support any of the filtering arguments, as users can just use the fields or filter as needed.


benhoyt:
@Guillaumebeuzeboc I updated unittypes.py per my Go types generator from the Juju types. Only minor changes, but some fields are "required" according to the Juju definition, and this keeps it in line with statustypes.py and modeltypes.py.

I'll add a few other review comments later, but this is looking pretty good, thanks.

tests/unit/test_show_unit.py
```diff
@@ -0,0 +1,71 @@
...
+    info = juju.show_unit('mysql/0')
+
+    assert info == jubilant.UnitInfo(
+        opened_ports=['3306/tcp'],
+        charm='mysql',
+        leader=False,
+    )
+
+
+def test_model_attribute(run: mocks.Run):
```
benhoyt: This is normally called `test_with_model` is other test modules.

tests/integration/test_show_unit.py
```diff
@@ -0,0 +1,38 @@
+from __future__ import annotations
+
+import jubilant
+
+from . import helpers
+
+
+def test_show_unit(juju: jubilant.Juju, juju_version: jubilant.Version, empty_tar: str):
```
benhoyt: Could we add this test to `test_relations.py` instead, to avoid deploying again and slowing down the integration tests?

tests/integration/test_show_unit.py
```diff
@@ -0,0 +1,38 @@
...
+    juju.deploy(helpers.find_charm('testdb'))
+    juju.deploy(helpers.find_charm('testapp'), resources={'test-file': empty_tar})
+
+    juju.integrate('testdb', 'testapp')
+    status = juju.wait(jubilant.all_active)
+
+    unit_name = next(iter(status.apps['testapp'].units))
```
benhoyt: The unit name is always going to be testapp/0, right? Can we just hard-code it here, like the other integration tests do in similar situations?

tests/integration/test_show_unit.py
```diff
@@ -0,0 +1,38 @@
...
+    info = juju.show_unit(unit_name)
+
+    assert isinstance(info, jubilant.UnitInfo)
+    assert 'testapp' in info.charm
```
benhoyt: Why `'testapp' in info.charm` rather than `info.charm == 'testapp'`?

tests/integration/test_show_unit.py
```diff
@@ -0,0 +1,38 @@
...
+    assert info.leader in (True, False)
```
benhoyt: What does this test do? A bool is always going to be either True or False. Presumably it'll always be true here? Or we can just drop the test, or perhaps `assert isinstance(info.leader, bool)` if we really want to check the type.

tests/integration/test_show_unit.py
```diff
@@ -0,0 +1,38 @@
...
+    assert info.machine or info.provider_id
+    assert info.life == 'alive'
+
+    if juju_version.major < 4:
```
benhoyt: Why are the Juju 3 and Juju 4 code paths different here? Could we make them the same? Or is this pointing to a bug in Juju 4?

benhoyt (changes requested):

tests/integration/test_show_unit.py
Guillaumebeuzeboc: I had never used juju 4 before. I ran the test that were passing on Juju 3 and it failed on Juju 4: https://github.com/canonical/jubilant/actions/runs/25000710225/job/73209761593

Juju 4 doesn't seem to populate `related_units` the way Juju 3 does. I am not sure if it's intended.

tests/unit/test_show_unit.py
Guillaumebeuzeboc: Fixed in: [ca1f653](https://github.com/canonical/jubilant/pull/308/commits/ca1f6535bd3ca1acab6965198af23d76ebc3a24c)

tests/integration/test_show_unit.py
Guillaumebeuzeboc: fixed in: [2302b3a](https://github.com/canonical/jubilant/pull/308/commits/2302b3a2733d11159e036dbac73bb1df315bc1e1)

tests/integration/test_show_unit.py
Guillaumebeuzeboc: Fixed in: [2302b3a](https://github.com/canonical/jubilant/pull/308/commits/2302b3a2733d11159e036dbac73bb1df315bc1e1)

tests/integration/test_show_unit.py
Guillaumebeuzeboc: fixed in: [2302b3a](https://github.com/canonical/jubilant/pull/308/commits/2302b3a2733d11159e036dbac73bb1df315bc1e1)

tests/integration/test_show_unit.py
Guillaumebeuzeboc: Fixed in: [2302b3a](https://github.com/canonical/jubilant/pull/308/commits/2302b3a2733d11159e036dbac73bb1df315bc1e1)

tests/integration/test_show_unit.py
benhoyt: @Guillaumebeuzeboc Are you able to re-run this test, and print out the full UnitInfo value (return value of show_unit) as well as `juju.cli('show-unit', unit, '--format', 'json')` for comparison, for both Juju 3 and Juju 4? That would help us know whether it's a Juju bug or not.

Guillaumebeuzeboc: Setup:
```
concierge prepare --preset microk8s --juju-channel=4/stable # or 3/stable
juju deploy grafana-agent-k8s --channel 2/stable --trust
juju deploy prometheus-k8s --channel 2/stable --trust
juju integrate grafana-agent-k8s:grafana-dashboards-consumer prometheus-k8s:grafana-dashboard
```

On juju 4:
```
root@juju4:~# juju show-unit grafana-agent-k8s/0
grafana-agent-k8s/0:
  workload-version: 0.40.4
  opened-ports: []
  charm: ch:amd64/grafana-agent-k8s-211
  leader: true
  life: alive
  relation-info:
  - relation-id: 1
    endpoint: peers
    related-endpoint: peers
    application-data: {}
    local-unit:
      in-scope: true
      data:
        ingress-address: 10.152.183.148
  - relation-id: 2
    endpoint: grafana-dashboards-consumer
    related-endpoint: grafana-dashboard
    application-data: {}
    local-unit:
      in-scope: true
      data:
        ingress-address: 10.152.183.148
  provider-id: grafana-agent-k8s-0
  address: 10.1.80.14
```
On juju 3:

```
root@juju3:~# juju show-unit grafana-agent-k8s/0
grafana-agent-k8s/0:
  workload-version: 0.40.4
  opened-ports: []
  charm: ch:amd64/grafana-agent-k8s-211
  leader: true
  life: alive
  relation-info:
  - relation-id: 2
    endpoint: grafana-dashboards-consumer
    related-endpoint: grafana-dashboard
    application-data: {}
    related-units:
      prometheus-k8s/0:
        in-scope: false
        data: {}
  - relation-id: 1
    endpoint: peers
    related-endpoint: peers
    application-data: {}
    local-unit:
      in-scope: true
      data:
        egress-subnets: 10.152.183.118/32
        ingress-address: 10.152.183.118
        private-address: 10.152.183.118
  provider-id: grafana-agent-k8s-0
  address: 10.1.141.205
```

As you can see juju4 doesn't have `related-units`.

benhoyt: Thanks. I've asked the Juju team if this is a bug / regression in Juju 4, or somehow by design.

SimonRichardson: This is a regression.

benhoyt: Thanks. I've opened https://github.com/juju/juju/issues/22539 to track the Juju issue.

@Guillaumebeuzeboc for this PR for now, can we just make the integration test general enough that it passes on both Juju 3 and 4. For example, for the code in question, something like:

```python
    assert any(
            relation.endpoint == 'db'
            and relation.related_endpoint == 'db'
            for relation in info.relation_info
        )
```

Guillaumebeuzeboc: done!

benhoyt (approved):

jubilant/unittypes.py
```diff
@@ -0,0 +1,82 @@
+"""Dataclasses that contain parsed output from ``juju show-unit --format=json``.
+
+These dataclasses were originally `generated from <https://github.com/juju/juju/compare/main...benhoyt:juju:unittypes-dataclasses>`_
+the Go structs in the Juju codebase, to ensure they are correct. Class names
+come from the Go struct name, whereas attribute names come from the JSON field
+names. The one exception is that "Application" has been renamed to "App"
+throughout, for brevity (and "application" to "app").
```
tonyandrewmeyer, line 7: I don't see "Application" anywhere?

jubilant/unittypes.py
```diff
@@ -0,0 +1,82 @@
+"""Dataclasses that contain parsed output from ``juju show-unit --format=json``.
+
+These dataclasses were originally `generated from <https://github.com/juju/juju/compare/main...benhoyt:juju:unittypes-dataclasses>`_
```
tonyandrewmeyer, line 3: I realise that we use this same link for modeltypes and statustypes, but I'm not sure it's a good one. Maybe we should change it universally, though. It points to a branch comparison that was done a long time ago, and was not done to generate the unit types, but for the original status types.

[The generation section of this post](https://benhoyt.com/writings/jubilant/) would be better, because it explains what is actually done. Maybe we don't want to link out to @benhoyt's site, though. I would leave it off in that case - it's already there in the other two modules, so doesn't add much value here.

I feel like linking to the Juju code would be more useful, but I think we rejected that previously.

jubilant/_juju.py
```diff
@@ -1455,6 +1456,17 @@ def show_secret(
             return RevealedSecret._from_dict(secret)
         return Secret._from_dict(secret)
 
+    def show_unit(self, unit: str) -> UnitInfo:
```
tonyandrewmeyer, line 1459: Apologies if I missed this discussion somewhere, but was there any thought about including `app`, `related-unit`, and `endpoint` arguments? If the idea is to leave them until someone asks, then I'm ok with that since I assume they would be keyword-only and backwards compatible.

I assume the choice to only do one unit at a time is to match the other methods, and that we would do a `get_units` sort of method in the future if we wanted to provide more than one.

tonyandrewmeyer (approved):
Would you mind rewriting the PR description? It gets used as the body of the commit message, so shows up in git blame and the like later on. It's better if it includes more of what was done and why and doesn't need to have the offer to change things and so on.

I think I would have preferred an xfail marker for the Juju 4 bug so that it's explicit that this data will be missing when using Jubilant with Juju 4, and we would get notified when that changed and could call it out. However, I'm ok with the simpler approach taken.

Overall, looks good to me. Thanks for the contribution! I've left a couple of questions, but they aren't blockers.

jubilant/unittypes.py
benhoyt: Yeah, it's not great. However, note that this is actually a new branch for these unit types (though branched off the previous ones). It's what I used to actually generate this file. To be honest, we can probably just get rid of these links at this point. We've edited some things by hand since using the generator (the statustypes.py file in particular).

I'd say: let's leave as is in this PR, but consider removing all of them at some point.

tonyandrewmeyer: Oh, I missed that indeed. I clicked through from the docs and it looked roughly the same, sorry!

benhoyt: Per voice discussion, I'll leave the comments as is in this PR, but open a separate PR with the tweaks and consolidation that we discussed.

jubilant/unittypes.py
benhoyt: I'll be updating these comments in a separate PR.

jubilant/_juju.py
benhoyt: Re the `--app`, `--endpoint`, and `--related-unit` args, similar to what we've done in some hookcmds, we've left them off because in the context of automation / an API (rather than a CLI tool) it's easiest just to get all the information by default, and let the user select/filter the fields as needed.

Re choice of single unit: yes, to match the other methods, and we could add `get_units` or similar later if needed.

PR #317, opened 2026-05-12
fix: add ./ for relative pathlib.Path objects in _deploy_tempdir

### LINKED ISSUE #277: juju.deploy(Path('foo.charm')) shouldn't be an error
`juju deploy` somewhat reasonable requires that a pathy charm argument must start with './' or '/' to avoid ambiguity.

Jubilant's `Juju.deploy` allows the charm to be specified as a `pathlib.Path`. There is no ambiguity about a `Path` like `Path('foo.charm')` -- that's a relative path since it doesn't start with `/`. `pathlib` even removes redundant explicit leading `./` when creating `Path` objects, so that `str(Path('./foo.charm'))` -> `'foo.charm'`.

Because Jubilant straightforwardly converts a `Path` argument with `str` before sending it to Juju, Juju will complain that `foo.charm` is ambiguous. But Jubilant knows that it's not ambiguous, and should format the path it gives to Juju to avoid this error.

The `_deploy_tempdir` function [already does its own detection of whether a charm argument is a local filesystem path](https://github.com/canonical/jubilant/blob/01b93d65a3277891cfa73499d20f7bf397d3f5f5/jubilant/_juju.py#L1502), it just does it *after* calling `str` on the argument.
```py
        if charm is not None:
            charm = str(charm)
        charm_needs_temp = charm is not None and charm.startswith(('.', '/'))
```
This could be modified to address this issue in a number of different ways, e.g.
```py
        if isinstance(charm, pathlib.Path):
            charm = str(charm.absolute())
        charm_needs_temp = charm is not None and charm.startswith(('.', '/'))
```
```py
        if isinstance(charm, pathlib.Path):
            charm = str(charm) if charm.root else f'./{charm}'
        charm_needs_temp = charm is not None and charm.startswith(('.', '/'))
```
Or even:
```py
        charm_needs_temp = (
            isinstance(charm, pathlib.Path)
            or isinstance(charm, str) and charm.startswith(('.', '/'))
        )
        if charm is not None:
            charm = str(charm)
```
Depending on what makes the most sense.

tromai:
Fixes #277 .

This PR modifies how `_deploy_tempdir` handles the `charm` parameter when it is a `pathlib.Path` object.
- If `charm` is an absolute path, we coerce it to a string, preserving its original path.
- If `charm` is a relative path, we coerce it to a string, and append `./` to the beginning.

I used `is_absolute()` instead of `root` to decide if a `pathlib.Path` object is absolute or not.

I added unit tests for deploying and refreshing from relative `pathlib.Path` objects.

**As a code reviewer, you can skip the below sections**

<details>
<summary> About the reverted commit </summary>

I added an integration test in https://github.com/canonical/jubilant/pull/317/commits/de01a1a749cf34af92ca4dba76ee74aa2e34e99d, which I later reverted:
- The test re-deploy the same `testapp` charm into the model.
- It didn't test what I wanted: all relative charm paths are converted to absolute paths in our CI, which has Juju installed as a snap https://github.com/canonical/jubilant/blob/5fd298b704724d8afb2c18828cd60828bda91fc3/jubilant/_juju.py#L1703-L1710

</details>

<details>

<summary> Discussing the suggestions in #277 </summary>

The original ticket mentioned multiple approaches. I will go through them one by one to explain why I didn't pick them:

### 1.

```python
        if isinstance(charm, pathlib.Path):
            charm = str(charm.absolute())
        charm_needs_temp = charm is not None and charm.startswith(('.', '/'))
```
This works well. However, calling `charm.absolute()` would:
- Make unit testing difficult, as `charm.absolute()` behaves based on where the unit tests are run (e.g. on my local machine `/home/tromai/.../jubilant/xyz` in Github Actions `/runner/.../zyx`).
- Forcing absolute path will break deploying/refreshing from a relative `pathlib.Path`object  (although I am not sure if anyone depends on this behavior).

### 2.
```python
        charm_needs_temp = (
            isinstance(charm, pathlib.Path)
            or 
            ( <-- Assumed brackets, not in the original text
                isinstance(charm, str) and charm.startswith(('.', '/'))
            ) <-- Assumed brackets, not in the original text
        )
        if charm is not None:
            charm = str(charm)
```
I am not sure this fix resolves the original issue, as `str(charm)` still remove redundant "./" in relative `pathlib.Path` objects.

</details>

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -1690,6 +1690,9 @@ class Juju:
         resources: Mapping[str, str] | None,
     ) -> Generator[tuple[str | None, Mapping[str, str] | None]]:
         if isinstance(charm, pathlib.Path):
+            # We add "./" to any relative pathlib.Path objects in their string form.
+            # pathlib.Path removes the redundant "./", which we need to prevent Juju
+            # from saying the path is ambiguos.
             charm = str(charm) if charm.is_absolute() else f'./{charm}'
         charm_needs_temp = charm is not None and charm.startswith(('.', '/'))
         resources_needs_temp = resources is not None and any(
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -1692,7 +1692,7 @@ class Juju:
         if isinstance(charm, pathlib.Path):
             # We add "./" to any relative pathlib.Path objects in their string form.
             # pathlib.Path removes the redundant "./", which we need to prevent Juju
-            # from saying the path is ambiguos.
+            # from saying the path is ambiguous.
             charm = str(charm) if charm.is_absolute() else f'./{charm}'
         charm_needs_temp = charm is not None and charm.startswith(('.', '/'))
         resources_needs_temp = resources is not None and any(
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -1689,8 +1689,8 @@ class Juju:
         charm: str | pathlib.Path | None,
         resources: Mapping[str, str] | None,
     ) -> Generator[tuple[str | None, Mapping[str, str] | None]]:
-        if charm is not None:
-            charm = str(charm)
+        if isinstance(charm, pathlib.Path):
+            charm = str(charm) if charm.is_absolute() else f'./{charm}'
         charm_needs_temp = charm is not None and charm.startswith(('.', '/'))
         resources_needs_temp = resources is not None and any(
             v.startswith(('.', '/')) for v in resources.values()
--- a/tests/integration/test_deployment.py
+++ b/tests/integration/test_deployment.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import pathlib
 import urllib.request
 
 import pytest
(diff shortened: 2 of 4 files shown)


jubilant/_juju.py
```diff
@@ -1689,8 +1689,11 @@ def _deploy_tempdir(
         charm: str | pathlib.Path | None,
         resources: Mapping[str, str] | None,
     ) -> Generator[tuple[str | None, Mapping[str, str] | None]]:
-        if charm is not None:
-            charm = str(charm)
+        if isinstance(charm, pathlib.Path):
+            # We add "./" to any relative pathlib.Path objects in their string form.
+            # pathlib.Path removes the redundant "./", which we need to prevent Juju
+            # from saying the path is ambiguos.
```
james-garner-canonical:
```suggestion
            # from saying the path is ambiguous.
```

james-garner-canonical (approved):
This looks clean to me, thanks.

jubilant/_juju.py
tromai: Fixed in https://github.com/canonical/jubilant/pull/317/commits/1a5691c2e61712d95987951812a77112ba2ed853. Mark this as resolved.

tonyandrewmeyer (approved):
Thanks, LGTM!

PR #321, opened 2026-05-19
test: remove Juju 4.0.2/4.0.3 exec stdout newline workaround

### LINKED ISSUE #320: Integration tests fail after Juju release fixes issues
Scheduled tests failed: https://github.com/canonical/jubilant/actions/runs/26000026405

tonyandrewmeyer:
The trailing-newline-stripping bug (juju/juju#21927) has been fixed upstream, so the `if juju_version.major == 4` branch in the exec tests now causes failures against 4.1-beta1. Drop the workaround and assert the correct stdout directly.

We shouldn't merge this until the fixes are out of beta, since they fail (correctly) in stable. I don't think we want to have temporary logic that handles both stable and beta cases, just hold off on merging.

[Run in my fork](https://github.com/tonyandrewmeyer/jubilant/actions/runs/26117582129). Note that the timeout issues are separate.

Fixes #320.

### DIFF
--- a/tests/integration/test_execution.py
+++ b/tests/integration/test_execution.py
@@ -76,32 +76,22 @@ def test_run_unit_not_found(juju: jubilant.Juju):
         juju.run('testdb/42', 'do-thing')
 
 
-def test_exec_success(juju: jubilant.Juju, juju_version: jubilant.Version):
+def test_exec_success(juju: jubilant.Juju):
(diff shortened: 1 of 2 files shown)


tromai (approved):
The changes look great.

> I don't think we want to have temporary logic that handles both stable and beta cases

I agree on this. Let's wait.

PR #325, opened 2026-05-25
ci: use sbomber to security scan and generate SBOM on release

tonyandrewmeyer:
Automates the security scanning and SBOM generation on release, like with pebble and operator.

Uses sbomber to run the secscan tool. Unlike Pebble, we know the version locally, so can run the tool with locally built versions rather than needing to download from PyPI, avoiding race conditions.


james-garner-canonical (approved):
Looks great, thanks.

PR #330, opened 2026-05-31
test: mock shutil.which in unit tests; avoid assuming juju not installed as snap

### LINKED ISSUE #328: Jubilant unit tests failed with snap Juju installed
Running Jubilant unit tests failed when Juju is installed as a snap into the environment.

## Reproduce 

1. Prepare Jubilant dev environment. 
2. Install Juju as a snap (e.g. `sudo snap install` or through [Concierge](https://github.com/canonical/concierge))
3. Run `make unit`.

## Error

There were 12 cases that failed. I will use one as an example:

```
FAILED tests/unit/test_add_cloud.py::test_with_path_str - FileNotFoundError: [Errno 2] No such file or directory: '/path/to/cloud.yaml'
```

This is the test code:
```python
def test_with_path_str(run: mocks.Run):
    run.handle(['juju', 'add-cloud', 'mycloud', '--client', '--file', '/path/to/cloud.yaml'])
    juju = jubilant.Juju()

    juju.add_cloud('mycloud', '/path/to/cloud.yaml', client=True)
```

`juju.add_cloud` will eventually call `_path_for_juju`
```
    @contextlib.contextmanager
    def _path_for_juju(self, path: str | pathlib.Path) -> Generator[str]:
        path = str(path)
        if not self._juju_is_snap:
            yield path
            return

        with tempfile.TemporaryDirectory(dir=self._temp_dir) as temp_dir:
            temp = os.path.join(temp_dir, os.path.basename(path))
            shutil.copy(path, temp)
            yield temp
```

Which copied `/path/to/cloud.yaml` to the temporary directory. `/path/to/cloud.yaml` is not a real file, which lead to `shutil.copy` raising an exception.

The error doesn’t happen in our CI because we don’t install Juju when running unit tests: [ci.yaml](https://github.com/canonical/jubilant/blob/5fd298b704724d8afb2c18828cd60828bda91fc3/.github/workflows/ci.yaml#L55-L70)

tromai:
Fixes #328 .

This PR mocks `shutil.which` with `/bin/juju` in unit tests by default, using a pytest "autouse" fixture.

Mocking it to `/bin/juju` prevents Jubilant from copying the fake path to a temporary directory which will produce an exception.

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -35,7 +35,7 @@ def mock_file(monkeypatch: pytest.MonkeyPatch) -> Generator[mocks.NamedTemporary
 
 @pytest.fixture(scope='function', autouse=True)
 def juju_non_snap(monkeypatch: pytest.MonkeyPatch) -> None:
-    """Pytest fixture that patches shutil.which to make ``Juju._juju_is_snap()`` returns ``True``.
+    """Pytest fixture that patches shutil.which to make ``Juju._juju_is_snap()`` return ``True``.
 
     Autouse because we want our unit tests to be isolated from the external environment.
     If a test relies on `Juju._juju_is_snap()` returning ``False``, it can override with
@@ -46,4 +46,3 @@ def juju_non_snap(monkeypatch: pytest.MonkeyPatch) -> None:
         monkeypatch.setattr('shutil.which', lambda _: '/snap/bin/juju')
     """
     monkeypatch.setattr('shutil.which', lambda _: '/bin/juju')  # type: ignore
-    return
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -31,3 +31,19 @@ def mock_file(monkeypatch: pytest.MonkeyPatch) -> Generator[mocks.NamedTemporary
     file_mock = mocks.NamedTemporaryFile()
     monkeypatch.setattr('tempfile.NamedTemporaryFile', file_mock)
     yield file_mock
+
+
+@pytest.fixture(scope='function', autouse=True)
+def juju_non_snap(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Pytest fixture that patches shutil.which to make ``Juju._juju_is_snap()`` returns ``True``.
+
+    Autouse because we want our unit tests to be isolated from the external environment.
+    If a test relies on `Juju._juju_is_snap()` returning ``False``, it can override with
+    monkeypatch.setattr.
+
+    Override example::
+
+        monkeypatch.setattr('shutil.which', lambda _: '/snap/bin/juju')
+    """
+    monkeypatch.setattr('shutil.which', lambda _: '/bin/juju')  # type: ignore
+    return
--- a/.github/workflows/tiobe.yaml
+++ b/.github/workflows/tiobe.yaml
@@ -38,7 +38,7 @@ jobs:
           pip install flake8 pylint -r requirements.txt --break-system-packages
 
       - name: TICS GitHub Action
-        uses: tiobe/tics-github-action@3300ccc8de3c938cb83dc940ec4d07096c1bf185  # v3.9.1
+        uses: tiobe/tics-github-action@016a4d79f3027fe7b4a2901619a58c745ab7cadf  # v3.10.0
         with:
           mode: qserver
           viewerUrl: https://canonical.tiobe.com/tiobeweb/TICS/api/cfg?name=default
--- a/docs/reference/index.md
+++ b/docs/reference/index.md
@@ -8,6 +8,7 @@ jubilant module <jubilant>
 modeltypes module <modeltypes>
 statustypes module <statustypes>
 secrettypes module <secrettypes>
+unittypes module <unittypes>
 ```
 
 These guides provide technical information about Jubilant.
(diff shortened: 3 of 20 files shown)


benhoyt (commented):
Thinking outside the square a bit: is it possible to just do this always by default, for all unit tests? (With the few tests that need to override which doing so.)

Or if that doesn't work, have a specific `which` fixture that tests can use, instead of having to monkeypatch it for every test that needs it?

tromai:
@benhoyt Thank you for the question.

I agree that this problem can be generalized. For me, the first approach is the right direction, with some challenges I don't have answer (yet).

## First approach

Ideally, unit tests should be independent from the external environment. Setting up `jubilant.Juju` as non-snap by default is good. All unit tests will behave similarly wherever they are run (in CI, in local machine).

Naive solution: an auto use fixture for `monkeypatch.setattr('shutil.which', lambda _: '/bin/juju')`. Right now, `jubilant` only use `shutil.which` at one place, so patching that should be fine. Will this be a problem in the future ?  

If a test depends on `jubilant.Juju` as a snap, the test will fail if it doesn't override <-- Preventing new tests from depending on the native `shutil.which`.

## Second approach
Looking at all invocations of `monkeypatch.setattr('shutil.which', lambda _: '...')` reveals a pattern:
```
monkeypatch.setattr('shutil.which', lambda _: '...')

[optionally do file operations]

juju = jubilant.Juju()
```
We can refactor them into fixtures: "juju-non-snap" and "juju-snap"

```
@pytest.fixture
def juju_snap(monkeypatch: pytest.MonkeyPatch) -> Generator[jubilant.Juju]:
    file_mock = mocks.NamedTemporaryFile()
    monkeypatch.setattr('shutil.which', lambda _: '/snap/bin/juju')
    juju = jubilant.Juju()
    yield juju
```

This approach doesn't prevent: writing a test that depends on the native `shutil.which` method.

benhoyt:
I think we should try the first approach and see what it looks like. What do you think, @tonyandrewmeyer?

tromai:
@benhoyt I have tried the first approach in a fork - https://github.com/tromai/jubilant/pull/4. All unit tests passed.

Overall:
- The fix is relatively simple: create an autouse fixture in `tests/unit/conftest.py`
```
@pytest.fixture(scope='function', autouse=True)
def juju_normal(monkeypatch: pytest.MonkeyPatch) -> None:
    monkeypatch.setattr('shutil.which', lambda _: '/bin/juju')  # type: ignore
    return
```
- There is _one_ test where removed the redundant `monkeypatch.setattr('shutil.which', lambda _: '/bin/juju')`
- Tests that wants Juju snap used `monkeypatch.setattr('shutil.which', lambda _: '/snap/bin/juju')` - which override the default `juju_normal` fixture.
- We should document this behavior in `CONTRIBUTING.md`.

Suggestions for that `autouse` fixure name are appreciated. I didn't want to use a negative name like `juju_is_not_snap` :thinking: .

benhoyt:
Love it, very simple!

- I don't think CONTRIBUTING.md is the right place to note this -- seems too specific. Could add a README.md to tests/unit, however, I actually think it'd be okay just left in conftest.py (I'm not in love with pytest's magic, but that's another issue -- we're using it, so people have to understand conftest.py to understand our tests).
- Perhaps `juju_bin_dir`? Though I actually think the negative is okay here, as it describes exactly what we're doing. So I'd be good with `juju_non_snap`. Then again, it doesn't matter much, especially as it's an "autouse" fixture, so won't be mentioned by name. So go with your judgement.

tromai:
Thank you!

I agree that `CONTRIBUTING.md` is too much. I will add it to the fixture's doc string.

> Then again, it doesn't matter much, especially as it's an "autouse" fixture

Yep, it is not used anywhere, and the detail has been covered by doc string. I will pick `juju_non_snap`.

benhoyt (approved):
Thanks, works for me!

tests/unit/conftest.py
```diff
@@ -31,3 +31,19 @@ def mock_file(monkeypatch: pytest.MonkeyPatch) -> Generator[mocks.NamedTemporary
     file_mock = mocks.NamedTemporaryFile()
     monkeypatch.setattr('tempfile.NamedTemporaryFile', file_mock)
     yield file_mock
+
+
+@pytest.fixture(scope='function', autouse=True)
+def juju_non_snap(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Pytest fixture that patches shutil.which to make ``Juju._juju_is_snap()`` returns ``True``.
```
tonyandrewmeyer:
```suggestion
    """Pytest fixture that patches shutil.which to make ``Juju._juju_is_snap()`` return ``True``.
```

tests/unit/conftest.py
```diff
@@ -31,3 +31,19 @@ def mock_file(monkeypatch: pytest.MonkeyPatch) -> Generator[mocks.NamedTemporary
...
+    Autouse because we want our unit tests to be isolated from the external environment.
+    If a test relies on `Juju._juju_is_snap()` returning ``False``, it can override with
+    monkeypatch.setattr.
+
+    Override example::
+
+        monkeypatch.setattr('shutil.which', lambda _: '/snap/bin/juju')
+    """
+    monkeypatch.setattr('shutil.which', lambda _: '/bin/juju')  # type: ignore
+    return
```
tonyandrewmeyer: We shouldn't need the explicit return, right?


```suggestion
```

tonyandrewmeyer (approved):
Works nicely, thanks! I repro'd the failing tests and that this fixes it.

In general I don't particularly like the magic of auto use fixtures and conftest, but this is a case where it does elegantly solve the problem.

I see we already have tests that monkeypatch for testing the opposite case.

tests/unit/conftest.py
tromai: Yep, we don't need it. I will remove it.

PR #331, opened 2026-06-01
refactor: avoid yaml.load() call in safe_load to silence scanner false positive

tonyandrewmeyer:
Pattern-based security scanners (e.g. Coverity Sigma unsafe_deserialization_core_python_yaml) flag any call to yaml.load() as unsafe deserialisation, regardless of the Loader argument. The # noqa: S506 comment only suppresses Ruff/Bandit, not these other tools.

The PR changes the code to instantiate the safe loader directly and call `get_single_data()`, which is exactly what `yaml.load()` [does internally](https://github.com/yaml/pyyaml/blob/d51d8a138f7230834fc6e95635ff09ebd329185f/lib/yaml/__init__.py#L74). This keeps the C-loader speedup and identical behaviour while no longer matching the `yaml.load()` text pattern.

I don't love changing code to avoid scanning false positives, but this seems reasonable. It's using public interfaces and avoids a method call, and is maybe more obvious to humans as well as scanners. It should bring the TIOBE security score for Jubilant from 76.72 to 100.


jubilant/_yaml.py
```diff
@@ -22,14 +22,31 @@ class _WriteStream(Protocol[_T_contra]):
...
 
+class _SafeLoader(Protocol):
+    def __init__(self, stream: _ReadStream, /) -> None: ...
+    def get_single_data(self) -> Any: ...
+    def dispose(self) -> None: ...
+
+
 # Use C speedups if available.
-_safe_loader = getattr(yaml, 'CSafeLoader', yaml.SafeLoader)
+_safe_loader: type[_SafeLoader] = getattr(yaml, 'CSafeLoader', yaml.SafeLoader)
```
benhoyt, line 32: Out of interest, do you know what `_safe_loader` needs a type annotation, but `_safe_dumper` below does not?

jubilant/_yaml.py
```diff
@@ -22,14 +22,31 @@ class _WriteStream(Protocol[_T_contra]):
...
 _safe_dumper = getattr(yaml, 'CSafeDumper', yaml.SafeDumper)
 
 
 def safe_load(stream: _ReadStream) -> Any:
-    """Same as yaml.safe_load, but use fast C loader if available."""
-    return yaml.load(stream, Loader=_safe_loader)  # noqa: S506
+    """Same as yaml.safe_load, but use fast C loader if available.
+
+    This deliberately instantiates the safe loader directly rather than calling
```
benhoyt: I think this paragraph should be a normal `#` comment, rather than part of the docstring, as it's implementation details, not part of the contract.

jubilant/_yaml.py
```diff
@@ -22,14 +22,31 @@ class _WriteStream(Protocol[_T_contra]):
...
+    ``yaml.load``, both to avoid the (false positive) "unsafe deserialization"
+    warnings that pattern-based scanners raise on any ``yaml.load`` call, and
+    because there is no point routing through ``yaml.load`` just to do this. It
+    is equivalent to ``yaml.load(stream, Loader=_safe_loader)``.
+    """
+    loader = _safe_loader(stream)
+    try:
+        return loader.get_single_data()
```
benhoyt, line 42: Where are PyYAML's `get_single_data` and `dispose` documented?

benhoyt (commented):
I'm okay with this, just a few questions first.

jubilant/_yaml.py
tonyandrewmeyer: `dispose` is on the [docs page](https://pyyaml.org/wiki/PyYAMLDocumentation), there doesn't seem to be anchors for sections so you need to search.

`get_single_data` is [only in the source](https://github.com/yaml/pyyaml/blob/d51d8a138f7230834fc6e95635ff09ebd329185f/lib/yaml/constructor.py#L47), I think. The docs for pyyaml are not great 😞.

It does say

> Warning: API stability is not guaranteed!

But that includes load and load_safe, so we seem at the same risk either way.

jubilant/_yaml.py
tonyandrewmeyer: `_safe_loader` needs the annotation because we later call `_safe_loader(stream)` and then `.get_single_data()` on the result. pyright needs the explicit type to type-check those calls.

`_safe_dumper` is only passed as `Dumper=_safe_dumper` to `yaml.dump()`, which accepts `type[Dumper]` and `Any` from `getattr()` is enough for that.

jubilant/_yaml.py
benhoyt: Yeah, that documentation sucks like a vacuum cleaner. That warning sure doesn't make me feel good, however, you're right, it applies to what we're already using. Let's roll with this. Absolute worst case they break it in a minor version and we need to do a simple fix.

benhoyt (approved):

PR #332, opened 2026-06-01
docs: expand the security page to include more sections

tonyandrewmeyer:
This PR expands the security doc to include all the required sections required by the latest revision of the Canonical SSDLC.

* Use the section titles in SEC030, to make it easier to track the requirements, and for more consistent docs across the Canonical products.
* Flesh out the architecture section more, in line with the SSDLC expectations.
* Add a logging section.
* Add a decommissioning section (the SSDLC is clearly focused on things not like Jubilant, but we can explain "decommissioning" in a "stop using jubilant" way).
* Repeat some of what's in SECURITY.md so that users don't have to jump to GitHub.

There's also a drive-by for SECURITY.md to point to this repo rather than operator. In #41, Ben said:

> Even links to that project's security advisory reporting to avoid having to set this fledgling project's one up.

However, I think Jubilant is beyond a "fledgling" project now, and security reporting is already set up anyway.

Another drive-by: add the meta description frontmatter that we're rolling out across the repos, in line with our "add when you touch a doc" practice.

[Preview](https://canonical-ubuntu-documentation-library--332.com.readthedocs.build/jubilant/explanation/security/)


docs/explanation/security.md
```diff
@@ -1,41 +1,85 @@
...
+```
+test process → jubilant → Juju CLI subprocess → Juju controller
+```
+
+- The **test process** calls Jubilant's Python API.
+- Jubilant translates each call into one or more {external+juju:ref}`Juju CLI <juju-cli>` invocations via `subprocess`. It does not communicate with Juju or any other service directly over a network.
+- The **Juju CLI** handles all communication with the Juju controller, including authentication, credential management, and TLS. Jubilant never reads, stores, or forwards Juju credentials.
 
-Jubilant does not use any cryptographic technology, hashing, or digital signatures.
+Several Jubilant operations pass structured data to the Juju CLI via temporary files (for example, YAML config passed to `add_cloud()`, `add_credential()`, `run_action()`, and `deploy()`). When the Juju CLI is installed as a snap, these files are written to `~/snap/juju/common/` so the snap can access them; otherwise the system temporary directory is used. In both cases the files are removed immediately after the CLI call returns. Jubilant retains no state between test runs.
```
dwilding: `run_action()` should be `run()`. Plus small readability suggestions.

Also, do you think it's worth linking the methods to the corresponding reference docs? Just a thought - I'm not sure it's really worth doing here.

```suggestion
Several Jubilant operations pass structured data to the Juju CLI by writing temporary files. For example, YAML config passed to `add_cloud()`, `add_credential()`, `run()`, and `deploy()`. When the Juju CLI is installed as a snap, these files are written to `~/snap/juju/common/` so the snap can access them; otherwise the system temporary directory is used. In both cases the files are removed immediately after the CLI call returns. Jubilant retains no state between test runs.
```

docs/explanation/security.md
```diff
@@ -1,41 +1,85 @@
...
+
+Jubilant is a pure-Python library with no persistent state, no network sockets, and no privileged access of its own. Its trust boundary is narrow:
+
+```
+test process → jubilant → Juju CLI subprocess → Juju controller
+```
+
+- The **test process** calls Jubilant's Python API.
+- Jubilant translates each call into one or more {external+juju:ref}`Juju CLI <juju-cli>` invocations via `subprocess`. It does not communicate with Juju or any other service directly over a network.
+- The **Juju CLI** handles all communication with the Juju controller, including authentication, credential management, and TLS. Jubilant never reads, stores, or forwards Juju credentials.
```
dwilding: This part could be read as contradicting the susequent paragraph, which mentions `add_credential()`. We say that Jubilant never forwards Juju credentials - I suppose technically this statement excludes cloud credentials, but I still think there's a possibility for confusion. How about an adjustment like below, leading into the following paragraph about passing structured data.

```suggestion
- The **Juju CLI** handles all communication with the Juju controller, including authentication, credential management, and TLS. Jubilant never stores credentials or reads credentials from Juju.
```

docs/explanation/security.md
```diff
@@ -1,41 +1,85 @@
...
+---
+
 (security)=
 
 # Security
 
 When producing security documentation for your charm, it's important to consider the security aspects of the charm's tests. If you have any Jubilant-related security questions that aren't answered here, please reach out to the Charm Tech team, and we'll do our best to help.
 
-## Cryptographic technology
+## Product architecture
```
dwilding, line 13: Do you think we ought to mention anything about secret handling? I guess it's already covered implicitly by this section - unless there's anything unusual we'd want to call out?

docs/explanation/security.md
```diff
@@ -1,41 +1,85 @@
...
-## Hardening
+### Hardening guidelines
 
-No additional steps are required to harden your integration tests when using Jubilant.
+No additional steps are required to harden your integration tests when using Jubilant. Follow the same security and hardening practices as you would if calling the Juju CLI directly.
 
 > See also: {external+juju:ref}`Juju | Harden your deployment <harden-your-deployment>`
 
-## Security updates
+### Logging and monitoring
```
dwilding, line 47: What if Jubilant crashes for some reason - could sensitive info, such as secrets passed to `add_secret()`, be exposed in the logs?

docs/explanation/security.md
```diff
@@ -1,41 +1,85 @@
...
 
-> See also:
-> - {external+juju:ref}`Juju CLI <juju-cli>`
+## Configuring and operating securely
 
-## Hardening
+### Hardening guidelines
 
-No additional steps are required to harden your integration tests when using Jubilant.
+No additional steps are required to harden your integration tests when using Jubilant. Follow the same security and hardening practices as you would if calling the Juju CLI directly.
```
dwilding, line 43: "No additional steps are required", but we do recommend some good practices. Maybe we should include a link to the "Good practices" section in the "See also" block.

docs/explanation/security.md
```diff
@@ -1,41 +1,85 @@
...
-Jubilant inherits the risks of using the Juju CLI. Charm authors should be familiar with {external+juju:ref}`Juju security <juju-security>`.
+You may also send email to security@ubuntu.com. If you want to encrypt your email, follow [Canonical's reporting instructions](https://ubuntu.com/security/disclosure-policy#contact-us).
 
-Charm integration tests are most commonly run in ephemeral CI environments. However, the tests may also be run in local development environments, giving Jubilant access to these environments.
+Known vulnerabilities and past advisories are listed at [github.com/canonical/jubilant/security/advisories](https://github.com/canonical/jubilant/security/advisories).
 
-Otherwise, Jubilant doesn't introduce any new security risks over directly running Juju CLI commands.
+The [Ubuntu Security disclosure and embargo policy](https://ubuntu.com/security/disclosure-policy) contains more information about what you can expect when you contact us, and what we expect from you.
 
 ## Good practices
```
dwilding, line 87: What do you think about adding a good practice related to `cli()`? E.g., do people need to take care with how they escape arguments that are passed to the Juju CLI?

docs/explanation/security.md
```diff
@@ -1,41 +1,85 @@
...
+```python
+import logging
+logging.getLogger("jubilant").setLevel(logging.DEBUG)
+```
+
+For security-relevant events (login failures, credential errors, controller access), consult the Juju controller logs and the Juju CLI's own output rather than Jubilant's logs.
+
+## Decommissioning securely
+
+Removing Jubilant from a project requires no special decommissioning steps. Remove `jubilant` from your `pyproject.toml` and re-lock. Jubilant stores no credentials, configuration, or logs of its own. Any Juju environments provisioned during testing should be destroyed using the Juju CLI (`juju destroy-controller`, `juju destroy-model`) independently of removing Jubilant.
```
dwilding: Just to be totally clear:

```suggestion
Removing Jubilant from a project requires no special decommissioning steps. Remove `jubilant` from your `pyproject.toml` and re-lock dependencies. Jubilant stores no credentials, configuration, or logs of its own. Any Juju environments provisioned during testing should be destroyed using the Juju CLI (`juju destroy-controller`, `juju destroy-model`) independently of removing Jubilant.
```

docs/explanation/security.md
```diff
@@ -1,41 +1,85 @@
...
+
+## Security lifecycle
+
+Jubilant follows [semantic versioning](https://semver.org/). Security updates are released as patch releases on all major versions that have had releases in the last year. See [SECURITY.md](https://github.com/canonical/jubilant/blob/main/SECURITY.md) for the supported-versions policy.
+
+**Receiving security updates.** Restrict the version of `jubilant` in `pyproject.toml` in a way that allows picking up new compatible releases every time that you re-lock. For example, `jubilant~=1.2` allows upgrades to 1.3, 1.4, and so on, while staying on the 1.x line.
```
dwilding:
```suggestion
**Receiving security updates.** Restrict the version of `jubilant` in `pyproject.toml` in a way that allows picking up new compatible releases every time that you re-lock dependencies. For example, `jubilant~=1.2` allows upgrades to 1.3, 1.4, and so on, while staying on the 1.x line.
```

docs/explanation/security.md
```diff
@@ -1,41 +1,85 @@
...
 
-We strongly recommend restricting the version of `jubilant` in `pyproject.toml` in a way that allows picking up new compatible releases every time that you re-lock. For example, `jubilant~=1.2`. Set a minor version that includes all the features that the charm tests use, and this will allow upgrades to 1.3, 1.4, and so on.
+**Detecting available updates.** Configure Dependabot or Renovate in your charm repository so that security updates are surfaced automatically as pull requests. Re-lock when prompted so that the charm tests run with the latest version.
```
dwilding:
```suggestion
**Detecting available updates.** Configure Dependabot or Renovate in your charm repository so that security updates are surfaced automatically as pull requests. Re-lock dependencies when prompted so that the charm tests run with the latest version.
```

docs/explanation/security.md
```diff
@@ -1,41 +1,85 @@
...
+> See also: {external+juju:ref}`Juju security <juju-security>`
+
+## Cryptographic technology
 
-Jubilant communicates with Juju by running processes in the test environment (the Juju CLI).
+Jubilant does not use any cryptographic technology, hashing, or digital signatures. All cryptographic operations (TLS, credential storage, API authentication) are handled by the Juju CLI and the Juju controller.
 
-> See also:
-> - {external+juju:ref}`Juju CLI <juju-cli>`
+## Configuring and operating securely
```
dwilding: Can we drop "securely" since the whole topic is security?

```suggestion
## Configuring and operating
```

docs/explanation/security.md
```diff
@@ -1,41 +1,85 @@
...
+To capture Jubilant's own log output (command invocations, return codes, and `stderr`), configure a handler for the `jubilant` logger in your test suite:
+
+```python
+import logging
+logging.getLogger("jubilant").setLevel(logging.DEBUG)
+```
+
+For security-relevant events (login failures, credential errors, controller access), consult the Juju controller logs and the Juju CLI's own output rather than Jubilant's logs.
+
+## Decommissioning securely
```
dwilding: Can we drop "securely" since the whole topic is security?

```suggestion
## Decommissioning
```

docs/explanation/security.md
```diff
@@ -1,41 +1,85 @@
...
+
+- The **test process** calls Jubilant's Python API.
+- Jubilant translates each call into one or more {external+juju:ref}`Juju CLI <juju-cli>` invocations via `subprocess`. It does not communicate with Juju or any other service directly over a network.
+- The **Juju CLI** handles all communication with the Juju controller, including authentication, credential management, and TLS. Jubilant never reads, stores, or forwards Juju credentials.
 
-Jubilant does not use any cryptographic technology, hashing, or digital signatures.
+Several Jubilant operations pass structured data to the Juju CLI via temporary files (for example, YAML config passed to `add_cloud()`, `add_credential()`, `run_action()`, and `deploy()`). When the Juju CLI is installed as a snap, these files are written to `~/snap/juju/common/` so the snap can access them; otherwise the system temporary directory is used. In both cases the files are removed immediately after the CLI call returns. Jubilant retains no state between test runs.
 
-## Inter-process communication
+## Secure by Design
```
dwilding:
```suggestion
## Secure by design
```

dwilding (approved):
Looks great, thanks!

My main comments are:

- `run()` instead of `run_action()`
- Potential confusion around the scope of credential forwarding

All other comments are very minor

docs/explanation/security.md
tonyandrewmeyer: I've made it explicit that temporary files are used and that the secret content is not in CLI calls.

docs/explanation/security.md
tonyandrewmeyer: I worry it would make this noisy, and I don't think people would be wanting to click through to a method from here.

docs/explanation/security.md
tonyandrewmeyer: No, only the filename of the temporary file would get exposed, which is worthless to an attacker (and by the time you see the crash, the temporary file is gone).

tonyandrewmeyer: I added a note about this.

docs/explanation/security.md
tonyandrewmeyer: No need to manage escaping, because the subprocess module handles making that safe. But people should avoid args with sensitive data, also when setting config, as they may be in logs. I'll add a note about that.

PR #336, opened 2026-06-04
docs: move dataclass codegen note from docs to comment; consolidate

benhoyt:
The copy-n-pasted documentation about the Go code generator in the `*types.py` files was getting a bit unwieldy. Plus, it should have been a comment, not in our documentation, as it's how the sausage is made and not important for users. 

Move it to a comment, consolidate in statustypes.py (and refer to there from other files), and add a few more details about how to run it.


tonyandrewmeyer (approved):
Nice! I think the docs are much cleaner without this, and the explanation will serve people better going forward. Thanks!

PR #337, opened 2026-06-05
docs: move docs to canonical.com/juju/docs/jubilant

dwilding:
This PR updates our doc config so that all URLs use `canonical.com/juju/docs/jubilant` instead of the backend URL from Read the Docs. Our docs are still hosted on Read the Docs, but we'll be proxying them under canonical.com/juju/docs.

<mark>**[Preview docs](https://canonical.com/juju/docs/jubilant/)** built from this branch. After this PR merges, I'll switch those docs to build from `main` and redirect the existing docs.</mark>

**Main changes**

- Updated Jubilant doc URLs in `docs/conf.py`.
- Updated hardcoded URLs for Jubilant, excluding `CHANGES.md`.
- Added a script `docs/_static/overwrite_links.js`, which is not strictly needed at the moment. This script makes sure that the version switcher from Read the Docs has correct URLs - but we don't use the version switcher. I'm including the script because it's a standard part of the new docs setup at Canonical.

**Extra changes**

- Updated the intersphinx URL for Ops. To account for [operator#2545](https://github.com/canonical/operator/pull/2545).

**Future work (not in this PR)**

- Update URLs for Juju docs.


docs/conf.py
```diff
@@ -116,19 +116,19 @@
...
-#       uncomment and set to the RTD slug.
-slug = "jubilant"
+# Set to the path after https://canonical.com/
+slug = "juju/docs/jubilant"
 
 #######################
 # Sitemap configuration: https://sphinx-sitemap.readthedocs.io/
 #######################
 
 # Use RTD canonical URL to ensure duplicate pages have a specific canonical URL
```
tonyandrewmeyer: Should this comment change since it's not pulling from the RTD environment variable any more?

docs/_static/overwrite_links.js
```diff
@@ -0,0 +1,28 @@
+// Our docs are proxied through to canonical.com/juju/docs/jubilant.
+// If multiple doc versions are configured, Read the Docs renders a version switcher.
+// The version switcher has the readthedocs-hosted.com URL in links, which we don't want,
+// so this script overwrites those links to use the correct public-facing URL.
```
tonyandrewmeyer, line 4: It's a bit awful we have to do this, although I understand why. I guess if you don't have JS enabled, you would just end up at a page that redirects you anyway?

But nice that it's not currently relevant for Jubilant anyway.

tonyandrewmeyer (approved):
Thanks for taking care of this! Fingers crossed this is the last move for a while 😂

docs/conf.py
```diff
@@ -297,7 +298,7 @@
 #
 intersphinx_mapping = {
     "juju": ("https://documentation.ubuntu.com/juju/3.6/", None),
```
tonyandrewmeyer, line 300: Should this also change, since Juju already moved?

README.md
tonyandrewmeyer, line 22: And maybe this one?

tonyandrewmeyer:
> **Future work (not in this PR)**
> 
> * Update URLs for Juju docs.

Oops. Only just noticed this.

docs/_static/overwrite_links.js
dwilding: Replying in https://github.com/canonical/operator/pull/2545#discussion_r3432733789, as the answer is most relevant to Ops.

PR #338, opened 2026-06-05
ci: split Juju 4 channel into 4.0 and 4.1, add edge to scheduled runs

tonyandrewmeyer:
[Juju's edge builds are now usable](https://discourse.canonical.com/t/charming-engineering-lt-notes-4th-june-2026/7578?u=tony-meyer). This PR switches from candidate to edge to get faster information about when Juju breaks Jubilant. Also adds edge as a manual dispatch option.

It was also a bit confusing recently as to whether 4.0/candidate or only 4.1/candidate had a fix, and we had to compare stable on a PR run. Adjust the scheduled run to do all of these for easy comparison, except 4.1/stable as that doesn't have a release yet.

Drive-by fix to move matrix.juju-channel through an env var on both Prepare Juju steps to clear the pre-existing zizmor template-injection warning.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,7 +9,7 @@ on:
   workflow_dispatch:
     inputs:
       channel:
-        description: "Juju channel risk (scheduled runs use beta, everything else uses stable)"
+        description: "Juju channel risk (PRs use stable; scheduled runs use stable and edge)"
         required: true
         type: choice
         default: stable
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer:
We would have to change the required checks too, because of the name change, I think.

james-garner-canonical (approved):
The change looks good and makes sense. As you say, looks like a PR is needed to update the required check names before landing this.

PR #340, opened 2026-06-08
docs: clarify behaviour when providing a user in the model value

### LINKED ISSUE #281: Document that model names can contain user names
In Juju, the full spec of a model name is `controller:user/mymodel`. Jubilant seems to be completely transparent with respect to user names. In other words, I can do `juju = jubilant.Juju(model='user/mymodel')` and nothing unexpected seems to happen. But we should verify this - and add tests if appropriate. Then document somewhere that model names can contain user names.

tonyandrewmeyer:
I validated each Jubilant method in the four forms (plain model, model and controller, model and user, controller and user). They 'just work' except in the case of add-model (and a destroy-model Jubilant issue, as below).

The PR explains this in the `model` attribute docstring and the `__init__` arg, and (for the not working case) `add_model` arg. If we want to support adding a model owned by someone else, I believe would need to add an `owner` arg that translates into `--owner`. I'm leaving that until someone asks for it.

A small issue in `destroy_model` is also fixed, where we would not properly check whether it was the same model if the user was provided (a qualified user vs. unqualified).

Fixes #281

### DIFF
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -63,7 +63,7 @@ class Juju:
     Args:
         model: If specified, operate on this Juju model, otherwise use the current Juju model.
             The model name is passed through to the Juju CLI unchanged, so the full
-            ``[<controller>:][<user>/]<model>`` form is accepted: Prefix the name with with
+            ``[<controller>:][<user>/]<model>`` form is accepted: Prefix the name with
             ``<user>/`` to operate on another user's model on the current controller. Prefix the
             name with ``<controller>:`` to operate on a model in another controller. For example,
             ``alice/my-model`` or ``mycontroller:alice/my-model``.
@@ -77,12 +77,12 @@ class Juju:
     """If not None, operate on this Juju model, otherwise use the current Juju model.
 
     The model name is passed through to the Juju CLI unchanged, so the full
-    ``[<controller>:][<user>/]<model>`` form is accepted. Prefix the name with ``<controller>:``
-    to operate on a model in another controller; prefix it with ``<user>/`` to operate on another
-    user's model on the current controller. For example::
+    ``[<controller>:][<user>/]<model>`` form is accepted. Prefix the name with ``<user>/`` to
+    operate on another user's model on the current controller; prefix it with ``<controller>:``
+    to operate on a model in another controller. For example::
 
-        juju = jubilant.Juju(model='mycontroller:my-model')
         juju = jubilant.Juju(model='alice/my-model')
+        juju = jubilant.Juju(model='mycontroller:my-model')
         juju = jubilant.Juju(model='mycontroller:alice/my-model')
     """
 
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -63,10 +63,10 @@ class Juju:
     Args:
         model: If specified, operate on this Juju model, otherwise use the current Juju model.
             The model name is passed through to the Juju CLI unchanged, so the full
-            ``[<controller>:][<user>/]<model>`` form is accepted: prefix the name with
-            ``<controller>:`` to operate on a model in another controller, and prefix it
-            with ``<user>/`` to operate on another user's model on the current controller
-            (for example ``alice/my-model`` or ``mycontroller:alice/my-model``).
+            ``[<controller>:][<user>/]<model>`` form is accepted: Prefix the name with with
+            ``<user>/`` to operate on another user's model on the current controller. Prefix the
+            name with ``<controller>:`` to operate on a model in another controller. For example,
+            ``alice/my-model`` or ``mycontroller:alice/my-model``.
         wait_timeout: The default timeout for :meth:`wait` (in seconds) if that method's *timeout*
             parameter is not specified.
         cli_binary: Path to the Juju CLI binary. If not specified, uses ``juju`` and assumes it is
--- a/jubilant/_juju.py
+++ b/jubilant/_juju.py
@@ -334,7 +334,12 @@ class Juju:
         model, all subsequent operations on this instance will use the new model.
 
         Args:
-            model: Name of model to add.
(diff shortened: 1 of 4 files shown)


jubilant/_juju.py
```diff
@@ -62,7 +62,11 @@ class Juju:
 
     Args:
         model: If specified, operate on this Juju model, otherwise use the current Juju model.
-            If the model is in another controller, prefix the model name with ``<controller>:``.
+            The model name is passed through to the Juju CLI unchanged, so the full
+            ``[<controller>:][<user>/]<model>`` form is accepted: prefix the name with
+            ``<controller>:`` to operate on a model in another controller, and prefix it
+            with ``<user>/`` to operate on another user's model on the current controller
+            (for example ``alice/my-model`` or ``mycontroller:alice/my-model``).
```
dwilding: A couple of small suggestions - breaking into short sentences and matching the explanation order with the examples order.

```suggestion
            The model name is passed through to the Juju CLI unchanged, so the full
            ``[<controller>:][<user>/]<model>`` form is accepted: Prefix the name with with
            ``<user>/`` to operate on another user's model on the current controller. Prefix the
            name with ``<controller>:`` to operate on a model in another controller. For example,
            ``alice/my-model`` or ``mycontroller:alice/my-model``.
```

jubilant/_juju.py
```diff
@@ -72,8 +76,14 @@ class Juju:
...
-    If the model is in another controller, prefix the model name with ``<controller>:``; for
-    example, ``juju = jubilant.Juju(model='mycontroller:my-model')``.
+    The model name is passed through to the Juju CLI unchanged, so the full
+    ``[<controller>:][<user>/]<model>`` form is accepted. Prefix the name with ``<controller>:``
+    to operate on a model in another controller; prefix it with ``<user>/`` to operate on another
+    user's model on the current controller. For example::
+
+        juju = jubilant.Juju(model='mycontroller:my-model')
+        juju = jubilant.Juju(model='alice/my-model')
+        juju = jubilant.Juju(model='mycontroller:alice/my-model')
```
dwilding: Maybe put the `alice/my-model` example first, then the two controller examples next? Matching the order of examples in the `model` arg description. Also maybe switch the explanation order to user then controller (as in my previous suggestion). I like this order better because user is "closer" to the model name, so it feels like it should come first, then controller second.

tests/unit/test_destroy_model.py
```diff
@@ -21,6 +21,55 @@ def test_destroy_other(run: mocks.Run):
     assert juju.model == 'initial'
 
 
+def test_destroy_user_qualified(run: mocks.Run):
```
dwilding, line 24: Did you deliberately not add a docstring to this test? I think the intention is clear enough. Just wondering because the others have a docstring.

tests/integration/test_model.py
```diff
@@ -17,3 +17,22 @@ def test_show_model(juju: jubilant.Juju):
 
     info_model = juju.show_model(juju.model)
     assert info_model.model_uuid == info.model_uuid
+
+
+def test_user_qualified_model_name_is_transparent(juju: jubilant.Juju):
+    """Jubilant forwards ``[<controller>:][<user>/]<model>`` to the Juju CLI verbatim."""
```
dwilding, line 23: This seems to be testing the round-trip rather than what goes into the CLI. Do you think we need an additional unit test in `test_show_model.py` to verify that the positional argument is set correctly? Overkill?

jubilant/_juju.py
```diff
@@ -1802,6 +1820,37 @@ def _format_config(k: str, v: ConfigValue) -> str:
...
+    if '/' in s:
+        user, s = s.split('/', 1)
+    return controller, user, s
+
+
+def _same_model(a: str | None, b: str | None) -> bool:
+    """Whether two ``[<controller>:][<user>/]<model>`` references identify the same model.
+
+    An unset controller or user on either side is treated as "the default" and
+    matches a set value on the other side; a set value on both sides must match.
```
dwilding, line 1837: Mention the similar behaviour for user (if I understand correctly)?

tests/unit/test_destroy_model.py
```diff
@@ -21,6 +21,55 @@ def test_destroy_other(run: mocks.Run):
...
+
+
+def test_destroy_unqualified_matches_qualified_self(run: mocks.Run):
+    """destroy_model('m') resets self.model when self.model == 'admin/m'."""
+    run.handle(['juju', 'destroy-model', 'm', '--no-prompt'])
+    juju = jubilant.Juju(model='admin/m')
+
+    juju.destroy_model('m')
+
+    assert juju.model is None
```
dwilding, line 51: Is it worth adding similar tests for the controller handling?

dwilding (approved):
Nice improvements, thanks!

tests/unit/test_destroy_model.py
tonyandrewmeyer: No, just habit from not always adding docstrings in tests. But it would be more consistent to do it here.

tests/integration/test_model.py
tonyandrewmeyer: The integration test already exercises the CLI end-to-end with a user-qualified name. If Jubilant screwed up the positional arg, show_model would fail to find the model and the  round-trip would never match.

A unit test would basically just be checking that the input matches the output, so not really be testing anything. I don't think it's worth adding (the lack of really testing the behaviour is why I did an integration test).

jubilant/_juju.py
tonyandrewmeyer: I think it's already there. It says "An unset controller or user on either side is treated as 'the default' and matches a set value on the other side". Or did you have something more specific in mind like adding an example?

dwilding: Yes... That's me misreading it

PR #343, opened 2026-06-10
chore: enable ruff format preview to allow cuddled braces

tonyandrewmeyer:
This PR enables `preview` for `ruff format`, which we generally use across Charm Tech Python repos.

The only change to existing code is to allow cuddling of parentheses and braces. There are two cases where a single line is split into multiple, which is perhaps less pretty, but overall this looks nicer, and is more consistent with our house style.

The easiest way to verify unintended changes haven't crept in is to use GitHub's "hide whitespace" option, which shows the cuddling, but hides the indentation changes.


benhoyt (approved):
Looks good, thanks!

PR #344, opened 2026-06-10
docs: clarify behaviour of any_* methods and fix various small issues

dwilding:
This PR clarifies how the `any_*` methods handle apps that can't be found. I think it's worth mentioning because it's not 100% obvious; maybe the methods raise an error (for example).

I've also included some fixes for small issues found by Copilot.


tonyandrewmeyer (approved):
Nice, thanks!

PR #351, opened 2026-06-26
feat: improve Jubilant logging mechanism

tromai:
This PR introduces changes to how Jubilant logs in `juju.wait()`. This change is discussed in `Spec OP087 - Improving Jubilant logs`.

### Content

- Log the verbose gron diff lines in `DEBUG` level.
- Log the app status and unit status changes at `INFO` level. For app and unit that change into 'error', they will be logged at `ERROR` level. 

### TODO list:

- [x] Log application status changes in juju.wait.
- [x] Log the gron diff lines in juju.wait as DEBUG.
- [x] Add unit tests. 
- [x] Validate e2e it with pytest-jubilant in a sample charm setup.
- [x] For the very first application status change, we need a new message format, instead of saying from "unknown". This is because "unknown" is a proper charm status reported from Juju.

### Future works
Works that will land in subsequent PRs.
- Update docs: how to configure logging, recommended setup for Github Actions workflow, and more.
- Update the Charm Tech repo with the Spec in Markdown form.


jubilant/_juju.py
```diff
@@ -14,12 +14,14 @@
 from collections.abc import Callable, Iterable, Mapping
 from typing import Any, Generator, Literal, Union, overload
 
+from jubilant._all_any import any_error
```
benhoyt: Nit: change to relative dot import like those below (and move to that section):

```python
from ._all_any import any_error
```

jubilant/_juju.py
```diff
@@ -1851,6 +1865,26 @@ def _same_model(a: str | None, b: str | None) -> bool:
     return True
 
 
+def _app_status_diff(old: AppStatus | None, new: AppStatus) -> str | None:
+    if old is None:
```
benhoyt, line 1874: What about doing this:

```python
old_current, old_message = (old.app_status.current, old.app_status.message) if old is not None else ('initial', '')
```

And then we only need the second block of logic?

Also, per voice discussion, let's use `initial` or `''` as the initial state to avoid "unknown -> unknown". And let's add a test for this case.

jubilant/_juju.py
```diff
@@ -1722,9 +1724,21 @@ def wait(
             status = Status._from_dict(result)
 
             if status != prev_status:
+                # Each app status is logged separately.
+                for name, app_status in status.apps.items():
+                    prev_app_status = prev_status.apps.get(name) if prev_status else None
+
+                    if app_diff := _app_status_diff(prev_app_status, app_status):
+                        # We treat apps transitioning to error state as an ERROR level log.
+                        if any_error(status, name):
```
benhoyt: I'm not 100% sure, but I don't think we should use `any_error()` here, because it looks at unit status too. If the charm (for some weird reason) has one unit in error but explicitly sets the app status to 'active', this should log at error level but show something like `blocked -> active`, which would be strange. I think we should just switch the level on the app status here: `if app_status.current == 'error'` or whatever it is.

benhoyt (commented):
Looks pretty good to me. A couple of things to think about.

jubilant/_juju.py
tromai: Thanks for pointing this out. I agree.

We have decided in the Spec to only switch on application status. Using `any_error` here also switches on unit status, which divert from the Spec (and make the behavior a bit more complicated as you pointed out). 

I will change it. Just one observation, the if statement would be

```python
if app_status.app_status.current == 'error':
   ...
else:
   ...
```

I think having two `app_status` would be confusing. I can definitely change it to 
```python
if new_app_status.app_status.current == 'error':
   ...
else:
   ...
```
if that works better.

jubilant/_juju.py
tromai: After some thinking, may I suggest we only print the new status if `old is None`:

```
waiting: installing agent
```

If `old is not None`, the format will be the same:

```
waiting -> active: active now
```

This behavior is consistent with _status_diff, where the first diff will only contain the first "state" 

https://github.com/canonical/jubilant/blob/d131da17b45e5281bbf77aa437be5bec29c7c2a7/jubilant/_juju.py#L1854-L1861

This `_app_status_diff` function will now be

```python
def _app_status_diff(old: AppStatus | None, new: AppStatus) -> str | None:
    old_current, old_message = (old.app_status.current, old.app_status.message) if old is not None else ('', '')

    diff_line = None
    if (
        new.app_status.current != old_current
        or new.app_status.message != old_message
    ):
        diff_line = f'{old_current} -> ' if old_current else ''
        diff_line += new.app_status.current
        if new.app_status.message:
            diff_line += f': {new.app_status.message}'

    return diff_line
```

jubilant/_juju.py
tromai: From this [discussion](https://github.com/canonical/jubilant/pull/351#discussion_r3478814415), we will not use `any_error`. This line will go away in the next push.

jubilant/_juju.py
```diff
@@ -1722,9 +1722,21 @@ def wait(
             status = Status._from_dict(result)
 
             if status != prev_status:
+                # Each app status is logged separately.
+                for name, new_app_status in status.apps.items():
```
benhoyt: Shall we do `in sorted(status.apps.items())` to ensure they're logged in key order (app name alphabetically), for consistency in the output?

tromai: I think that's a good idea, thanks! I will change it.

benhoyt:
This is a huge improvement. From our own integration tests, a huge long noisy chunk becomes concise:

```
# AFTER
tests/integration/test_config.py::test_config 
-------------------------------- live log setup --------------------------------
23:04:41 INFO cli: juju add-model --no-switch jubilant-6448bb02
23:04:41 INFO cli: juju deploy --model jubilant-6448bb02 /home/runner/snap/juju/common/tmpk9lk557e/_temp.charm
23:04:42 INFO testdb: waiting: installing agent
23:04:53 INFO testdb: waiting -> waiting: agent initialising
23:04:54 INFO testdb: waiting -> maintenance: installing charm software
23:04:55 INFO testdb: maintenance -> unknown
-------------------------------- live log call ---------------------------------
23:04:57 INFO cli: juju config --model jubilant-6448bb02 --format json testdb
23:04:58 INFO cli: juju config --model jubilant-6448bb02 testdb testoption=foobar
23:04:58 INFO cli: juju config --model jubilant-6448bb02 --format json testdb
23:04:58 INFO cli: juju config --model jubilant-6448bb02 testdb --reset testoption
23:04:58 INFO cli: juju config --model jubilant-6448bb02 --format json testdb
PASSED                                                                   [  7%]

# BEFORE
tests/integration/test_config.py::test_config 
-------------------------------- live log setup --------------------------------
22:28:32 INFO cli: juju add-model --no-switch jubilant-43279d1a
22:28:32 INFO cli: juju deploy --model jubilant-43279d1a /home/runner/snap/juju/common/tmpk1spnbkt/_temp.charm
22:28:33 INFO wait: status changed:
+ .model.name = 'jubilant-43279d1a'
+ .model.type = 'caas'
+ .model.controller = 'concierge-k8s'
+ .model.cloud = 'k8s'
+ .model.version = '3.6.24'
+ .model.model_status.current = 'available'
+ .apps['testdb'].charm = 'local:testdb-0'
+ .apps['testdb'].charm_origin = 'local'
+ .apps['testdb'].charm_name = 'testdb'
+ .apps['testdb'].charm_rev = 0
+ .apps['testdb'].exposed = False
+ .apps['testdb'].base.name = 'ubuntu'
+ .apps['testdb'].base.channel = '24.04'
+ .apps['testdb'].scale = 1
+ .apps['testdb'].app_status.current = 'waiting'
+ .apps['testdb'].app_status.message = 'installing agent'
+ .apps['testdb'].units['testdb/0'].workload_status.current = 'waiting'
+ .apps['testdb'].units['testdb/0'].workload_status.message = 'installing agent'
+ .apps['testdb'].units['testdb/0'].juju_status.current = 'allocating'
+ .apps['testdb'].endpoint_bindings[''] = 'alpha'
+ .apps['testdb'].endpoint_bindings['db'] = 'alpha'
22:28:42 INFO wait: status changed:
+ .apps['testdb'].units['testdb/0'].provider_id = 'testdb-0'
22:28:43 INFO wait: status changed:
- .apps['testdb'].app_status.current = 'waiting'
- .apps['testdb'].app_status.message = 'installing agent'
- .apps['testdb'].units['testdb/0'].workload_status.current = 'waiting'
- .apps['testdb'].units['testdb/0'].workload_status.message = 'installing agent'
- .apps['testdb'].units['testdb/0'].juju_status.current = 'allocating'
+ .apps['testdb'].app_status.current = 'maintenance'
+ .apps['testdb'].app_status.message = 'installing charm software'
+ .apps['testdb'].units['testdb/0'].workload_status.current = 'maintenance'
+ .apps['testdb'].units['testdb/0'].workload_status.message = 'installing charm software'
+ .apps['testdb'].units['testdb/0'].juju_status.current = 'executing'
+ .apps['testdb'].units['testdb/0'].juju_status.message = 'running install hook'
+ .apps['testdb'].units['testdb/0'].juju_status.version = '3.6.24'
+ .apps['testdb'].units['testdb/0'].leader = True
22:28:44 INFO wait: status changed:
+ .apps['testdb'].provider_id = '2c813063-e26c-43c4-adc8-11c0994b9539'
+ .apps['testdb'].address = '10.152.183.191'
- .apps['testdb'].units['testdb/0'].juju_status.message = 'running install hook'
+ .apps['testdb'].units['testdb/0'].juju_status.message = 'running start hook'
+ .apps['testdb'].units['testdb/0'].address = '10.1.0.220'
22:28:45 INFO wait: status changed:
- .apps['testdb'].app_status.current = 'maintenance'
- .apps['testdb'].app_status.message = 'installing charm software'
- .apps['testdb'].units['testdb/0'].workload_status.current = 'maintenance'
- .apps['testdb'].units['testdb/0'].workload_status.message = 'installing charm software'
- .apps['testdb'].units['testdb/0'].juju_status.current = 'executing'
- .apps['testdb'].units['testdb/0'].juju_status.message = 'running start hook'
+ .apps['testdb'].app_status.current = 'unknown'
+ .apps['testdb'].units['testdb/0'].workload_status.current = 'unknown'
+ .apps['testdb'].units['testdb/0'].juju_status.current = 'idle'
-------------------------------- live log call ---------------------------------
22:28:48 INFO cli: juju config --model jubilant-43279d1a --format json testdb
22:28:48 INFO cli: juju config --model jubilant-43279d1a testdb testoption=foobar
22:28:48 INFO cli: juju config --model jubilant-43279d1a --format json testdb
22:28:48 INFO cli: juju config --model jubilant-43279d1a testdb --reset testoption
22:28:49 INFO cli: juju config --model jubilant-43279d1a --format json testdb
PASSED                                                                   [  7%]
```

jubilant/_juju.py
```diff
@@ -1851,6 +1871,33 @@ def _same_model(a: str | None, b: str | None) -> bool:
     return True
 
 
+def _diff_and_log(name: str, old: StatusEntity | None, new: StatusEntity) -> None:
+    diff = _entity_status_diff(old, new)
+    if not diff:
+        return
+    log = logger_wait.error if new.current == 'error' else logger_wait.info
+    log('[%s] status changed: %s', name, diff)
```
benhoyt: Mightly slightly more natural:

```suggestion
    level = logging.ERROR if new.current == 'error' else logging.INFO
    logger_wait.log(level, '[%s] status changed: %s', name, diff)
```

jubilant/_juju.py
```diff
@@ -1851,6 +1871,33 @@ def _same_model(a: str | None, b: str | None) -> bool:
...
+
+    @property
+    def message(self) -> str: ...
+
+
+def _entity_status_diff(old: StatusEntity | None, new: StatusEntity) -> str | None:
+    old_current, old_message = (old.current, old.message) if old is not None else ('', '')
+    new_current, new_message = new.current, new.message
+
+    if new_current != old_current or new_message != old_message:
```
benhoyt: Let's invert and return None if they're both equal, to put the "main path" or "happy path" at the top level of indentation.

benhoyt (approved):
Couple of minor comments, but approving optimistically.

tests/unit/test_pretty.py
```diff
@@ -152,7 +155,176 @@ def test_large():
...
+    ],
+)
+def test_entity_status_diff(
+    old_current: str,
+    old_message: str,
+    new_current: str,
+    new_message: str,
+    expect: str,
+):
+    # It's simplest to test _entity_status_diff directly, even though it's not public.
```
tromai: This function tests against app status changes. I can add unit status changes, but that would duplicate everything, without extra benefit. `_entity_status_diff` accepts `StatusInfo`, and only read `current` and `message` attributes from it, so I think this is fine. 

But let me know if it's better to add unit status changes here too.

tests/unit/test_pretty.py
```diff
@@ -152,7 +155,176 @@ def test_large():
...
+    assert (
+        jubilant._juju._entity_status_diff(
+            old_status.apps['snappass-test'].app_status,
+            new_status.apps['snappass-test'].app_status,
+        )
+        is None
+    )
+
+
+def test_diff_and_log_no_change(caplog: pytest.LogCaptureFixture):
```
tromai: I had to test `_diff_and_log_entity_status` here because we couldn't test the same thing with public `juju.wait()`. Our `mocks.Run` doesn't handle 2 consecutive `juju status` commands.

jubilant/_juju.py
```diff
@@ -1851,6 +1871,26 @@ def _same_model(a: str | None, b: str | None) -> bool:
...
 
+def _diff_and_log_entity_status(name: str, old: StatusInfo | None, new: StatusInfo) -> None:
+    diff = _entity_status_diff(old, new)
+    if not diff:
+        return
+    level = logging.ERROR if new.current == 'error' else logging.INFO
+    logger_wait.log(level, '[%s] status changed: %s', name, diff)
+
+
+def _entity_status_diff(old: StatusInfo | None, new: StatusInfo) -> str | None:
```
tromai: Alternatives:

#### 1
```
class _EntityStatusInfo(Protocol):
    @property
    def current(self) -> str: ...
    @property
    def message(self) -> str: ...

def _entity_status_diff(old: _EntityStatusInfo | None, new: _EntityStatusInfo) -> str | None:
```
This serves the purpose, but we need to add an extra Protocol, and it doesn't prevent:
```
_entity_status_diff(app_status, unit_status)
```

#### 2
We can use a Generic Type

```
T = TypeVar('T', AppStatus, UnitStatus)

def _entity_status_diff(old: T | None, new: T) -> str | None: 
```
This can prevent:
```
_entity_status_diff(app_status, unit_status)
```

But we need to "widen" the parameter types to `AppStatus` and `UnitStatus`, and make `_entity_status_diff` a bit more involved.

As this is a private function, I don't think 

```
_entity_status_diff(app_status, unit_status)
```

is a problem, so I went with: 

```
def _entity_status_diff(old: StatusInfo | None, new: StatusInfo) -> str | None:
```

@james-garner-canonical What do you think about this?

tests/unit/test_pretty.py
tromai: Ah nevermind. I figured out a way to test this case. This function will go away.

The test case will be in `test_wait.py`

```python
...
    snappass_json = json.loads(SNAPPASS_JSON)
    
    run.handle(['juju', 'status', '--format', 'json'], stdout=json.dumps(snappass_json))
    juju = jubilant.Juju()

    count = 0
    def helper() -> bool:
        # Return False 2 times, then return True.
        nonlocal count
        if count < 1:
            count += 1
            return False

        return True

    caplog.set_level(logging.INFO, logger='jubilant.wait')

    juju.wait(lambda _: helper())

    assert len(caplog.records) == 2  # only log 1 unit + 1 app the first time.
```

jubilant/_juju.py
```diff
@@ -1851,6 +1871,26 @@ def _same_model(a: str | None, b: str | None) -> bool:
...
+    old_current, old_message = (old.current, old.message) if old is not None else ('', '')
+    new_current, new_message = new.current, new.message
+
+    if new_current == old_current and new_message == old_message:
+        return None
+
+    diff_line = f'{old_current} ({old_message}) -> ' if old_current else ''
+    diff_line += f'{new_current} ({new_message})'
+    return diff_line
```
james-garner-canonical: **question:** Can `old_current` be falsey if `old` isn't `None`?

**suggestion:** If not, I think this would be clearer:
```suggestion
    if old is None:
        return f'{new.current} ({new.message})'
    if (old.current, old.message) == (new.current, new.message):
        return None
    return f'{old.current} ({old.message}) -> {new.current} ({new.message})'
```

jubilant/_juju.py
james-garner-canonical: I agree with your judgement that this function signature doesn't need to be responsible for ensuring that the statuses are of the same type. It would be problematic if we mixed them up at the call site, but the call site is clear enough that I'm not worried about this. Simple is good here.

jubilant/_juju.py
```diff
@@ -1851,6 +1871,26 @@ def _same_model(a: str | None, b: str | None) -> bool:
     return True
 
 
+def _diff_and_log_entity_status(name: str, old: StatusInfo | None, new: StatusInfo) -> None:
+    diff = _entity_status_diff(old, new)
+    if not diff:
+        return
+    level = logging.ERROR if new.current == 'error' else logging.INFO
+    logger_wait.log(level, '[%s] status changed: %s', name, diff)
```
james-garner-canonical: **idea:** I wonder if in-lining the `_entity_status_diff` helper would give us more flexibility here without much cost. This would make it easy to use a different message when there's no diff (the new status is brand new).

Assuming my read of the `_entity_status_diff` helper's logic is correct, this might look like:
```suggestion
    level = logging.ERROR if new.current == 'error' else logging.INFO
    if old is None:
        logger_wait.log(level, '[%s] status: %s (%s)', name, new.current, new.message)
        return
    if (old.current, old.message) == (new.current, new.message):
        return
    template = '[%s] status changed: %s (%s) -> %s (%s)'
    logger_wait.log(level, template, name, old.current, old.message, new.current, new.message)
```
As a very minor bonus, this would also defer all string interpolation to the logger.

jubilant/_juju.py
```diff
@@ -1851,6 +1871,26 @@ def _same_model(a: str | None, b: str | None) -> bool:
     return True
 
 
+def _diff_and_log_entity_status(name: str, old: StatusInfo | None, new: StatusInfo) -> None:
```
james-garner-canonical: **minor:** The diff bit is an implementation detail, WDYT about `_log_entity_status_if_needed` or `_log_entity_status_if_changed`?

I also wonder if 'entity' is needed. How about:
```suggestion
def _log_status_if_needed(name: str, old: StatusInfo | None, new: StatusInfo) -> None:
```

jubilant/_juju.py
```diff
@@ -1722,9 +1722,29 @@ def wait(
...
+                        new_app_status.app_status,
+                    )
+                    prev_units = prev_app_status.units if prev_app_status else {}
+                    for unit_name, new_unit_status in sorted(new_app_status.units.items()):
+                        prev_unit = prev_units.get(unit_name)
+                        _diff_and_log_entity_status(
+                            unit_name,
+                            prev_unit.workload_status if prev_unit else None,
+                            new_unit_status.workload_status,
+                        )
```
james-garner-canonical: **minor:** This could be refactored to reflect the fact that we do the same thing with app and unit statuses (sketch below). I think this is largely a matter of taste, totally non-blocking.
```py
                for app_name, new_app in sorted(status.apps.items()):
                    prev_app = prev_apps.get(app_name)
                    prev_app_status = prev_app.app_status if prev_app else None
                    items = [(app_name, prev_app_status, new_app.app_status)]
                    for unit_name, new_unit in sorted(new_app.units.items()):
                        prev_unit = prev_units.get(unit_name)
                        prev_unit_status = prev_unit.workload_status if prev_unit else None
                        items.append((unit_name, prev_units_status, new_unit.workload_status))
                    for name, prev, new in items:
                        _log_status_if_needed(name, prev, new)
```

tests/unit/test_pretty.py
```diff
@@ -152,7 +155,160 @@ def test_large():
     assert eval(status_repr, jubilant.statustypes.__dict__) == status
 
 
-def test_diff():
+@pytest.mark.parametrize(
```
james-garner-canonical, line 158: The testing approach looks good, and matches how we previously tested the wait logging.

james-garner-canonical (commented):
Thanks for this, great work. I have one user-facing / design concern: I think we should use a differently worded message the first time we report a status for an entity (when there's no previous status), since `status changed` feels odd in that case.

When there's a previous status, our message reflects a relatively recent status change, but when there's no previous status, we actually don't know when it changed. For example, it could have been set 10 minutes ago in an earlier test. So `status changed: <status> (<message>)` feels a little misleading. Perhaps a bare `status: <status> (<message>)` is better in that case. WDYT?

With the current `_entity_status_diff` abstraction, this is a little tricky, but I wonder if in-lining it into the logging helper would be a clean way to address this (see [comment](https://github.com/canonical/jubilant/pull/351/changes#r3532886845)).

My other comments/suggestions are non-blocking.

jubilant/_juju.py
tromai: Thank you. That sounds great.

jubilant/_juju.py
tromai: The reason I used `entity` was to not confuse with `def _status_diff(old: Status | None, new: Status) -> str:`, which is for the verbose diff lines. I still prefer a name that can keep them away from each other.

I like your suggestion to remove `diff` from the name. May be we can go with: 

```
def _log_short_status_if_needed(name: str, old: StatusInfo | None, new: StatusInfo) -> None:
```
?

jubilant/_juju.py
tromai: I intentionally had `_entity_status_diff` to make the unit testing easier (similar to how `_status_diff` is being tested). 

I like your suggestion. If we inline it, the logic is much simple. But unit tests need to change. We will need to use `caplog` and extract the log records out of it, instead of plain string assertions.

I slightly lean toward keeping `_entity_status_diff`. But am happy to switch.

If we keep `_entity_status_diff`, the change can be

```
    diff = _entity_status_diff(old, new)
    if not diff:
        return
    level = logging.ERROR if new.current == 'error' else logging.INFO
    if old is None:
        logger_wait.log(level, '[%s] status: %s', name, diff)
        return
    
    logger_wait.log(level, '[%s] status changed: %s', name, diff)
``` 

Because `_entity_status_diff` shows only the new status if old status is None.

jubilant/_juju.py
james-garner-canonical: Sounds reasonable to me.

jubilant/_juju.py
tromai: `StatusInfo.current` and `StatusInfo.message` are of type `str`. So yes `old_current` can be falsy if `old is not None`.
In that case `diff_line = f'{old_current} ({old_message}) -> ' if old_current else ''` will be an empty string -> the final `diff_line` only contains new current and new message. But I can see this pattern is not clear.

I like your suggestion. I will change it.

jubilant/_juju.py
james-garner-canonical: `caplog` is a good general purpose solution, but for unit testing this function it might be even easier to `monkeypatch.setattr(_juju.logger_wait, 'log', mock_log)`.

I'd lean towards prioritising library code simplicity over test simplicity for a case like this, but I don't mind either way.

jubilant/_juju.py
james-garner-canonical: I'd guess that in practice `StatusInfo.current` can only be the real Juju status values, so we don't need to worry about it being the `''` library code default at runtime. But it might be worth checking that.

This does make me think though -- presumably `message` could often be an empty string, in which case we'd log something like `blocked () -> active ()`. I wonder if that's desirable or if we should drop the `()` when there's no message (which complicates the message construction a bit and is probably a point in favour of keeping it as a separately testable unit).

jubilant/_juju.py
tromai: Sounds good. Thanks!

After we updated the signature for `_log_status_if_needed`, it's good to refactor this block too.

jubilant/_juju.py
tromai: I'm convinced. I will go with your suggestion 😁.

jubilant/_juju.py
tromai: I double checked the possible values of `StatusInfo.current`. 

With my limited understanding: from `juju status`, `StatusInfo.current` and `StatusInfo.message` are of type string ([reference](https://github.com/juju/juju/blob/08dd5b61f906bf16a8d078dacf4944c5bfa39c95/cmd/juju/status/formatted.go#L291-L299)). They are all marked as `omitempty`, and will be removed from the JSON result if they are empty strings.
 
In Jubilant, if they are real values, nothing changes. But if they doesn't exist in the JSON, Jubilant turned them into empty strings (for [example](https://github.com/canonical/jubilant/blob/7fa4ba69d8bf595049f6faa45bb3ba69c11de99c/jubilant/statustypes.py#L82-L83)).

Even if we don't need to worry about `old_current` value, I still think it's better to `if-else` on `old` instead of `old.current`, as you suggested.

> presumably message could often be an empty string, in which case we'd log something like blocked () -> active ()

Thanks for pointing that out. I can see an empty `()` being slightly unusual, but I intentionally kept it to maintain the symmetrical format. I think this was discussed in one of the daily sync.

james-garner-canonical: Makes sense, thanks for following up with this.

jubilant/_juju.py
tromai: I made these changes in https://github.com/canonical/jubilant/pull/351/commits/6ed3e11dd4584f78f79ea59c2906de3dc9022cd9:
- Refactor to inline `_entity_status_diff` entirely in this function.
- Update unit tests. I decided to use caplog. I looked at my changes again, and saw that I already used caplog in another test function :sweat_smile:. It seems good to be consistent with `tests/unit/test_wait.py` too. Let me know if it looks okay, I'm more than happy to explore the `monkeypatch` direction.

james-garner-canonical (approved):
Look great!

PR #352, opened 2026-06-27
ci: attest build provenance for released artifacts

tonyandrewmeyer:
This PR adds build provenance attestation to align with the other Charm Tech repos (and to have the attestation for security checks).


tromai (approved):
It looks good.

Is there a reason we go with version 3 in this pipeline, instead of version 4 ? I can see we have version 4 for operator and pytest-jubilant.

https://github.com/canonical/operator/blob/17b0eaf735e705ec09ea0af56a4e4c4e701141e9/.github/workflows/publish.yaml#L31

https://github.com/canonical/pytest-jubilant/blob/c71442eead3fc695d76d2f0b9292adb0b4ebaa3c/.github/workflows/build-and-publish.yaml#L39

tonyandrewmeyer:
> Is there a reason we go with version 3 in this pipeline, instead of version 4 ?

No, other than I took a while to open the PR. Fixed, thanks, avoids a dependabot bump later.

PR #370, opened 2026-07-05
ci: bump minimum python version to 3.10

### LINKED ISSUE #300: Bump required Python version to 3.10
At some point we should drop support or Python 3.8, as we mostly have elsewhere. At that point you would need to use an older Jubilant to run your integration tests with 3.8, which probably isn't much of an issue for most people. If people are using Jubilant in production with older Ubuntu bases, that might be more of an issue.

That would also mean we can use ops 3.x rather than 2.23.x.

There are a few small changes that ruff and pyright will make as a result, but that should easily fit into the same PR.

Does this mean a major version bump as we did with ops?

tromai:
This PR resolves #300 .

Changes included:

- Bumped the minimal Python version in `pyproject.toml`.
- Updated `AGENTS.md`.
- Bumped python version of `pyright` and `ruff`, re-ran the linting tools and fixed errors.
- Re-lock `uv.lock`.
- Removed python 3.8 from `ci.yaml` matrix.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -48,10 +48,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: ["3.8", "3.10", "3.12", "3.14"]
-        exclude:
-          - os: macos-latest
-            python-version: "3.8"
+        python-version: ["3.10", "3.12", "3.14"]
 
     steps:
       - name: Check out repo
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -23,7 +23,7 @@ make help                   # Run to see all available commands
 - **Docstrings**: Google style
 - **Type hints**: Strict pyright mode; `from __future__ import annotations` in every file
 - **Imports**: Import modules, not other objects, unless they are only used for type annotations
-- **Target Python**: 3.8+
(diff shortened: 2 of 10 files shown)


tonyandrewmeyer:
We'll need a canonical-repo-automation PR to drop the required status check before this can merge. After it's merged, we likely want to do a matching one in pytest-jubilant.

tonyandrewmeyer (approved):
Looks good, thanks! I'm still good with this not being a major bump, and it seems like the Python used to run integration tests should be easily updated, unlike unit tests or the charm itself.

tromai:
> We'll need a canonical-repo-automation PR to drop the required status check before this can merge. After it's merged, we likely want to do a matching one in pytest-jubilant.

Thanks @tonyandrewmeyer ! I will create a PR on canonical-repo-automation. Happy to do the same thing for pytest-jubilant if no one has worked on it yet :smile:

tromai:
`ops==3.8.0` that I bumped in this PR is too new for the 7 days cut off from uv :rofl:  . That's why most of our checks failed.

We have a couple of options:
1. Wait another day (ops 3.8.0 was released on 30 of June), and run the CI again.
1. Ignore ops from this cut off `exclude-newer-package = { op = false }`.
1. Use an older version of ops.

I'm happy to wait (option 1).

PR #374, opened 2026-07-15
feat: expose wait to CLI

### LINKED ISSUE #323: Expose a CLI entrypoint providing Jubilant's wait functionality
Add a CLI entrypoint to jubilant, so that after installing `jubilant` you can run `jubilant wait ...` on from the commandline -- or equivallently, `uvx jubilant wait ...`.

The goal is to expose Jubilant's rich wait semantics (which allow arbitrary Python logic to be executed based on a [status object](https://documentation.ubuntu.com/jubilant/reference/statustypes/#jubilant.statustypes.Status)) to be used directly from the commandline when interacting with a live Juju controller, for example:
```bash
jubilant wait 'status.apps["foo"].bar=="baz"'
```

This is currently possible like this:
```bash
uvx --with jubilant \
    python -c 'import jubilant; jubilant.Juju().wait(lambda status: status.apps["foo"].bar=="baz")'
```

---

`jubilant wait` should accept a single positional argument, which is a string interpreted with roughly the following semantics:
```py
jubilant.Juju().wait(lambda status: eval("<input>"))
```

We should probably also expose a way to configure:
- the Juju executable passed to `jubilant.Juju` (env var? top-level CLI option?)
- the wait timeout, polling delay, and required number of successes (CLI options for `jubilant wait`)
- perhaps also the `error` argument for [jubilant.Juju.wait](https://documentation.ubuntu.com/jubilant/reference/jubilant/#jubilant.Juju.wait) (which would be a second string to evaluate)

To avoid the unfortunate implications of Hyrum's Law, the user's string should be evaluated in a strictly defined environment, e.g. `locals={'jubilant': jubilant, 'juju': <the jubilant.Juju instantiated>, 'status': status}`.

---

The user's `ready` (and `error`) strings need to be run with `eval` to return a (boolean) value for `wait` to act on. (We could instead read output from a specific variable, but this complicates the user's input in the simplest cases.) `eval` will evaluate the string as a single Python statement. However note that this doesn't prevent the user from executing arbitrary logic, using a string like `exec('<logic>') or some_value` (`exec` always returns 1). For example:

```py
s = r'''exec("""
def foo():
    print("hello")
    return 1
a = foo()
""") or a'''
b = eval(s)  # prints "hello"
assert b == 1
```

Passing such a string on the CLI is annoying due to quoting issues (probably simplest as `jubilant wait "$(</path/to/file)"`), so we shouldn't design this feature around being able to execute more than a single Python statement, but we should be aware that the user will be able to execute arbitrary code.

---

The idea for this feature came via Sinan, and Charm Tech discussed it at the 26.04 Engineering Sprint and agreed it's a good idea.

tromai:
Resolves #323 

This PR adds a CLI entrypoint that exposes Jubilant's wait method. This CLI entrypoint can be used in two ways:
- After installing Jubilant, you can run `jubilant wait ...`
- With uv: `uvx jubilant wait ...`

The goal is to expose Jubilant's rich wait semantics to be used directly from the command line when interacting with a live Juju controller. See [Examples](#examples).
- The ready condition (a positional argument) and error condition (optional argument) accept Python expressions as strings. These expressions will be evaluated in a strictly defined environment. This environment contains the `jubilant` module, the `jubilant.Juju` instance of the live Juju model, and the `jubilant.Status` object.
- delay, timeout, and successes are optional arguments. Their default values are identical to their corresponding `juju.wait`'s arguments.
- You can specify the Juju executable passed to `jubilant.Juju` with the argument `--juju-cli-bin`.

For more details on these arguments, see [Help](#help).

## Prepare to run the CLI locally

Check out this branch, bootstrap a Juju model.

## Help

<details>
<summary>Click to expand</summary>

```
> jubilant wait --help
usage: jubilant wait [-h] [--delay DELAY] [--error ERROR]
                     [--successes SUCCESSES] [--timeout TIMEOUT]
                     [--juju-cli-bin JUJU_CLI_BIN]
                     [--model MODEL] [--quiet | --verbose]
                     ready

The wait command queries Juju status and checks that the ready condition succeeds
a number of times in a row (default 3).

Both ready and --error accept Python expressions. Those expressions have access to
three variables: "jubilant" (the jubilant module), "juju" (the jubilant.Juju instance),
and "status" (the jubilant.Status object).

Examples:

    jubilant wait 'jubilant.all_active(status)'

    jubilant wait 'jubilant.all_active(status, "app")' --error 'jubilant.any_error(status)'

These are equivalent to the following Python calls:

    juju.wait(jubilant.all_active)

    juju.wait(
        lambda status: jubilant.all_active(status, 'app'),
        error=jubilant.any_error,
    )

positional arguments:
  ready                 Python expression for the ready condition

options:
  -h, --help            show this help message and exit
  --delay DELAY         delay in seconds between status calls
                        (default: 1.0)
  --error ERROR         Python expression for the error condition
                        (default: None)
  --successes SUCCESSES
                        number of times `ready` must evaluate to
                        true for the wait to succeed (default: 3)
  --timeout TIMEOUT     overall timeout in seconds (default:
                        180.0)
  --juju-cli-bin JUJU_CLI_BIN
                        path to the Juju CLI binary
  --model MODEL         the Juju model to operate on, otherwise
                        use the current Juju model
  --quiet               suppress all output except errors
  --verbose             increase verbosity
```

</details>

## Exit codes

| Code | Meaning |
|------|---------|
| `0` | The ready condition succeeded. |
| `1` | The error condition evaluated to true, or an exception was raised while waiting. |
| `124` | The wait timed out. |
| `130` | Keyboard Interrupted received. |

### Deploy snappass-test and wait

On one terminal, run the wait first
```
uv run jubilant wait "jubilant.all_agents_idle(status, 'snappass-test')"
```

On another terminal, deploy `snappass-test`

```
juju deploy snappass-test
```

Output from `jubilant wait ...`

<details>
<summary>Click to expand</summary>

```
2026-07-15 23:16:51,000 [testapp] status: waiting (installing agent)
2026-07-15 23:16:51,000 [testapp/0] status: unknown ()
2026-07-15 23:16:59,000 [snappass-test] status: waiting (installing agent)
2026-07-15 23:16:59,000 [snappass-test/0] status: waiting (installing agent)
2026-07-15 23:17:10,000 [snappass-test] status changed: waiting (installing agent) -> maintenance (installing charm software)
2026-07-15 23:17:10,000 [snappass-test/0] status changed: waiting (installing agent) -> maintenance (installing charm software)
2026-07-15 23:17:12,000 [snappass-test] status changed: maintenance (installing charm software) -> unknown ()
2026-07-15 23:17:12,000 [snappass-test/0] status changed: maintenance (installing charm software) -> unknown ()
2026-07-15 23:17:13,000 [snappass-test] status changed: unknown () -> waiting (installing agent)
2026-07-15 23:17:16,000 [snappass-test/0] status changed: unknown () -> active (redis started)
2026-07-15 23:17:17,000 [snappass-test/0] status changed: active (redis started) -> active (snappass started)
2026-07-15 23:17:19,000 Ready condition succeeded 3 times (jubilant.all_agents_idle(status, 'snappass-test'))
```
</details>

### Wait with --error condition

Assuming `snappass-test` is already deployed.

```
$ uv run jubilant wait False --error 1==1 

2026-07-15 23:21:11,000 [snappass-test] status: active (snappass started)
2026-07-15 23:21:11,000 [snappass-test/0] status: active (snappass started)
2026-07-15 23:21:11,000 [testapp] status: waiting (installing agent)
2026-07-15 23:21:11,000 [testapp/0] status: unknown ()
2026-07-15 23:21:11,000 Error expression evaluated to true (1==1)
```

### Wait triggers timeout 

Assuming `snappass-test` is already deployed. This will timeout before the wait expression evaluate to True 3 times.

```
$ uv run jubilant wait True --timeout 1.0 

2026-07-15 23:22:48,000 [snappass-test] status: active (snappass started)
2026-07-15 23:22:48,000 [snappass-test/0] status: active (snappass started)
2026-07-15 23:22:48,000 [testapp] status: waiting (installing agent)
2026-07-15 23:22:48,000 [testapp/0] status: unknown ()
2026-07-15 23:22:49,000 Wait timed out after 1.0 seconds
```

### The expression raises an exception when evaluated. 

Assuming `snappass-test` is already deployed. All exceptions will be caught and the CLI will gracefully exit.

```
$ uv run jubilant wait 0/0

2026-07-15 23:23:35,000 [snappass-test] status: active (snappass started)
2026-07-15 23:23:35,000 [snappass-test/0] status: active (snappass started)
2026-07-15 23:23:35,000 [testapp] status: waiting (installing agent)
2026-07-15 23:23:35,000 [testapp/0] status: unknown ()
2026-07-15 23:23:35,000 Exception while waiting: ZeroDivisionError('division by zero')
```

If you want the verbose `juju.wait` logs, increase the verbosity:

<details>
<summary>Click to expand</summary>

```
$ uv run jubilant --verbose wait 0/0

2026-07-15 23:34:09,000 INFO jubilant.wait [snappass-test] status: active (snappass started)
2026-07-15 23:34:09,000 INFO jubilant.wait [snappass-test/0] status: active (snappass started)
2026-07-15 23:34:09,000 INFO jubilant.wait [testapp] status: waiting (installing agent)
2026-07-15 23:34:09,000 INFO jubilant.wait [testapp/0] status: unknown ()
2026-07-15 23:34:09,000 DEBUG jubilant.wait wait: status changed:
+ .model.name = 'test-jubilant'
+ .model.type = 'caas'
+ .model.controller = 'concierge-k8s'
+ .model.cloud = 'k8s'
+ .model.version = '3.6.23'
+ .model.model_status.current = 'available'
+ .apps['snappass-test'].charm = 'snappass-test'
+ .apps['snappass-test'].charm_origin = 'charmhub'
+ .apps['snappass-test'].charm_name = 'snappass-test'
+ .apps['snappass-test'].charm_rev = 9
+ .apps['snappass-test'].exposed = False
+ .apps['snappass-test'].base.name = 'ubuntu'
+ .apps['snappass-test'].base.channel = '20.04'
+ .apps['snappass-test'].charm_channel = 'latest/stable'
+ .apps['snappass-test'].scale = 1
+ .apps['snappass-test'].provider_id = 'a3486300-f889-4a78-b059-89d119b34da8'
+ .apps['snappass-test'].address = '10.152.183.109'
+ .apps['snappass-test'].app_status.current = 'active'
+ .apps['snappass-test'].app_status.message = 'snappass started'
+ .apps['snappass-test'].units['snappass-test/0'].workload_status.current = 'active'
+ .apps['snappass-test'].units['snappass-test/0'].workload_status.message = 'snappass started'
+ .apps['snappass-test'].units['snappass-test/0'].juju_status.current ='idle'
+ .apps['snappass-test'].units['snappass-test/0'].juju_status.version ='3.6.23'
+ .apps['snappass-test'].units['snappass-test/0'].leader = True
+ .apps['snappass-test'].units['snappass-test/0'].address = '10.1.0.32'
+ .apps['snappass-test'].units['snappass-test/0'].provider_id = 'snappass-test-0'
+ .apps['testapp'].charm = 'local:testapp-0'
+ .apps['testapp'].charm_origin = 'local'
+ .apps['testapp'].charm_name = 'testapp'
+ .apps['testapp'].charm_rev = 0
+ .apps['testapp'].exposed = False
+ .apps['testapp'].base.name = 'ubuntu'
+ .apps['testapp'].base.channel = '24.04'
+ .apps['testapp'].scale = 1
+ .apps['testapp'].provider_id = 'b9b188b2-755d-4763-a5eb-8ee36c221025'
+ .apps['testapp'].address = '10.152.183.41'
+ .apps['testapp'].app_status.current = 'waiting'
+ .apps['testapp'].app_status.message = 'installing agent'
+ .apps['testapp'].units['testapp/0'].workload_status.current = 'unknown'
+ .apps['testapp'].units['testapp/0'].juju_status.current = 'idle'
+ .apps['testapp'].units['testapp/0'].juju_status.version = '3.6.23'
+ .apps['testapp'].units['testapp/0'].leader = True
+ .apps['testapp'].units['testapp/0'].address = '10.1.0.247'
+ .apps['testapp'].units['testapp/0'].provider_id = 'testapp-0'
+ .apps['testapp'].endpoint_bindings[''] = 'alpha'
+ .apps['testapp'].endpoint_bindings['db'] = 'alpha'
2026-07-15 23:34:09,000 ERROR jubilant.cli Exception while waiting: ZeroDivisionError('division by zero')
```

</details>

### The ready expression tries to access a module not in scope

```
$ uv run jubilant wait "os.name==posix"

2026-07-15 23:25:06,000 [snappass-test] status: active (snappass started)
2026-07-15 23:25:06,000 [snappass-test/0] status: active (snappass started)
2026-07-15 23:25:06,000 [testapp] status: waiting (installing agent)
2026-07-15 23:25:06,000 [testapp/0] status: unknown ()
2026-07-15 23:25:06,000 Exception while waiting: NameError("name 'os' is not defined")
```

### Silent output with --quiet

When `--quiet` is specified, all status output is suppressed. Only error messages are printed.

On success, there is no output at all:

```
$ uv run jubilant wait True --quiet

$
```

On failure, only the error line is printed:

```
$ uv run jubilant wait True --timeout 1.0 --quiet

2026-07-15 23:22:49,000 Wait timed out after 1.0 seconds
```


jubilant/__main__.py
```diff
@@ -0,0 +1,180 @@
...
+        juju.wait(**wait_kwargs)
+    except jubilant.WaitError:
+        logger.error('The --error expression "%s" evaluates to True', args.error)
+        return 1
+    except TimeoutError:
+        logger.error('Wait timed out after %s seconds', args.timeout)
+        return 1
+    except Exception as error:
+        logger.error('Exception raised while evaluating the input expression: %s', str(error))
+        logger.debug('', exc_info=True)
```
benhoyt: Looking at the output, this doesn't add any value, but is just the jubilant stack trace, which is a bit confusing to the user. So let's drop this DEBUG log.

jubilant/__main__.py
```diff
@@ -0,0 +1,180 @@
...
+    try:
+        juju.wait(**wait_kwargs)
+    except jubilant.WaitError:
+        logger.error('The --error expression "%s" evaluates to True', args.error)
+        return 1
+    except TimeoutError:
+        logger.error('Wait timed out after %s seconds', args.timeout)
+        return 1
+    except Exception as error:
+        logger.error('Exception raised while evaluating the input expression: %s', str(error))
```
benhoyt: Let's make this a bit terser, but then also include the exception class name:

```
Exception evaluating "ready" or "error": ZeroDivisionError: division by zero
```

jubilant/__main__.py
```diff
@@ -0,0 +1,180 @@
...
+        'timeout': args.timeout,
+        'successes': args.successes,
+    }
+    if args.error is not None:
+        wait_kwargs['error'] = _helper(args.error)
+
+    try:
+        juju.wait(**wait_kwargs)
+    except jubilant.WaitError:
+        logger.error('The --error expression "%s" evaluates to True', args.error)
```
benhoyt: My suggestion would be something like:

```
Error expression evaluated to true (%s)
```

jubilant/__main__.py
```diff
@@ -0,0 +1,180 @@
...
+        return 1
+    except TimeoutError:
+        logger.error('Wait timed out after %s seconds', args.timeout)
+        return 1
+    except Exception as error:
+        logger.error('Exception raised while evaluating the input expression: %s', str(error))
+        logger.debug('', exc_info=True)
+        return 1
+
+    logger.info('Status condition %s is ready.', args.ready)
```
benhoyt: How about this:

```
Ready condition succeeded %d times (%s)
```

jubilant/__main__.py
```diff
@@ -0,0 +1,180 @@
...
+        'ready': _helper(args.ready),
+        'delay': args.delay,
+        'timeout': args.timeout,
+        'successes': args.successes,
+    }
+    if args.error is not None:
+        wait_kwargs['error'] = _helper(args.error)
+
+    try:
+        juju.wait(**wait_kwargs)
```
benhoyt: Let's make this a bit clearer/simpler and specify all the arguments as named keyword args here. `error` can be None for no error condition.

jubilant/__main__.py
```diff
@@ -0,0 +1,180 @@
...
+        arg_parser.print_usage()
+        return 1
+
+    juju = jubilant.Juju(cli_binary=args.juju_cli_bin)
+
+    def _helper(expression: str) -> Callable[[jubilant.Status], bool]:
+        return lambda status, val=expression: eval(  # noqa: S307
+            val,
+            {'jubilant': jubilant, 'juju': juju, 'status': status},
+        )
```
benhoyt: Please add a comment why if this doesn't work:

```suggestion
        return lambda status: eval(  # noqa: S307
            expression,
            {'jubilant': jubilant, 'juju': juju, 'status': status},
        )
```

jubilant/__main__.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def main(argv: Sequence[str] | None = None) -> int:
+    """The main entrypoint."""
+    arg_parser = argparse.ArgumentParser('jubilant')
+
+    arg_parser.add_argument(
+        '--verbose',
+        action='store_true',
+        help='Increaes verbosity.',
```
benhoyt:
```suggestion
        help='Increase verbosity.',
```

pyproject.toml
```diff
@@ -42,6 +42,9 @@ dev = [
 requires = ["setuptools>=60"]
 build-backend = "setuptools.build_meta"
 
+[project.scripts]
```
tromai, line 45: Unanswered questions:
- Should we ship this if people installed it as a library ? 
- When should we ship this ? How ?

I will look into this.

tonyandrewmeyer:
Drive-by comment: it's probably worth running the cli-standards skill over the branch (or manually evaluating it again the CLI standards).

tromai:
> Drive-by comment: it's probably worth running the cli-standards skill over the branch (or manually evaluating it again the CLI standards).

That's a great idea, I will give the skill a try. Thanks!

jubilant/__main__.py
tromai: This was a leftover from my initial prototype. Back then, I created these lambdas inside a for loop, with `expression` being a loop variable

```
for expression in ["boo", "foo"]:
    # create lambda
    ... = lambda status: eval(  # noqa: S307
        aexpression
        {'jubilant': jubilant, 'juju': juju, 'status': status},
    )
```
`expression` here belong to the outside scope, and lambda accesses `expression` when it's being called, not when it's defined. These lambdas when called will access `expression` from the outside scope -> get the same value. Therefore, `lambda status, val=expression:` saves `expression` as a local variable, at lambda definition time. 

In the current implementation, this pattern is not needed. I have removed it.

See [more](https://docs.python.org/3/faq/programming.html#why-do-lambdas-defined-in-a-loop-with-different-values-all-return-the-same-result).

tromai:
@tonyandrewmeyer I ran the CLI standard skill, got some useful warnings out of it.

**These were the issues I fixed:**
- Remove short hand argument (e.g `-e` for `--error`).
- Added a `version` command.
- Added a `--quiet` flag, which when used only display errors to the terminal.

**There were other warnings:**

- Error messages. Right now our error messages doesn't follow the standards' [examples](https://github.com/canonical/charm-tech/blob/main/skills/engineering/cli-standards/references/examples.md#feedback-error-messages) directly.

Ours:

```
Wait timed out after 180 seconds
```

CLI standard:

```
error: cannot establish the connection
```

If we follow the standard, we might have something like:
```
error: cannot wait: timed out after 180 seconds.
```
I'm not too sure which way is better.

- `--verbosity=trace` and `--verbosity=debug` mentioned in [verbosity-levels](https://github.com/canonical/charm-tech/blob/main/skills/engineering/cli-standards/SKILL.md#verbosity-levels)). I can't shake off the feeling that this is irrelevant, and would clutter the options we already have.

pyproject.toml
tromai: From [uv documentation](https://docs.astral.sh/uv/concepts/projects/config/#command-line-interfaces), this is the way to define CLI entrypoints for the project

```
[project.scripts]
jubilant = "jubilant.__main__:main"
```

Additional examples of this pattern:
- [hyrum](https://github.com/canonical/hyrum/blob/75bb135cd41e15291dcf61ea68b34e53592d4a50/pyproject.toml#L22-L23)
- [Create executable script doc](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#creating-executable-scripts):

> To install a command as part of your package, declare it in the [project.scripts] table.
> ```
> [project.scripts]
> spam-cli = "spam:main_cli"
> ```
> In this example, after installing your project, a spam-cli command will be available. Executing this command will do the equivalent of import sys; from spam import main_cli; sys.exit(main_cli()).

I currently don't see any breaking change. But I will double check with hyrum.

benhoyt:
Regarding CLI standards and error format, I don't think it's mandating that specific format, but rather "Be concise, precise, and clear at all times" -- so I'd recommend sticking with the existing format.

jubilant/_main.py
```diff
@@ -0,0 +1,197 @@
+"""Jubilant CLI."""
```
benhoyt, line 1: Regarding the filename, I'd prefer just `_main` rather than `__main__`, as the latter feels like a special Python thing, whereas I think this is just a private file (and we reference it from pyproject.toml).

jubilant/__main__.py
```diff
@@ -0,0 +1,197 @@
...
+
+import jubilant
+
+logger = logging.getLogger('jubilant.cli')
+
+# ISO 8601.
+DATE_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
+
+
+# ISO 8601 requires UTC.
```
benhoyt: Are we sure we want to log in UTC? This logging output is mostly for human consumption, so maybe local time is better?

jubilant/__main__.py
```diff
@@ -0,0 +1,197 @@
...
+
+    # version subcommand
+    _ = sub_parser.add_parser(
+        name='version',
+        description='Show the version.',
+    )
+
+    wait_description: str = """
+    The wait command queries Juju status and checks that the ready condition succeeds
+    three times in a row.
```
benhoyt: It's not necessarily three if they specify `--successes`:
```suggestion
    a number of times in a row (default 3).
```

jubilant/__main__.py
```diff
@@ -0,0 +1,197 @@
...
+
+    Use --error to terminate the command early with a condition.
+
+    Both ready and --error accept Python expressions. Those expressions have accesses to
+    the jubilant module, the jubilant.Juju instance, and the jubilant.Status object.
```
benhoyt: We should say how these variables are named, something like:
```suggestion
    three variables: "jubilant" (the jubilant module), "juju" (the jubilant.Juju instance), and "status" (the jubilant.Status object).
```

jubilant/__main__.py
```diff
@@ -0,0 +1,197 @@
...
+
+    Set a timeout in seconds for this command with --timeout.
+
+    Set the delay in seconds between each Juju status query with --delay.
+
+    Set the number of consecutive successes the ready condition must have with --successes.
```
benhoyt: I'd drop these lines (and the "Use --error to terminate the command early with a condition." line above) and make the help text of the option specific enough that it gives the meaning there, otherwise we're duplicating information. It's only for the further details about ready/error that we need an extra paragraph here.

jubilant/__main__.py
```diff
@@ -0,0 +1,197 @@
...
+    Examples:
+        juju.wait(jubilant.all_active)
+        juju.wait(
+            lambda status: jubilant.all_active(status, 'snappass-test'),
+            error=jubilant.any_error,
+        )
+
+    can be run from the CLI as:
+        jubilant wait jubilant.all_active
+        jubilant wait "jubilant.all_active(status, 'snappass-test')"
```
benhoyt: I think it's more typical to use single quotes for the outer shell quotes here, like `awk '{ print $1 }'`, so I suggest inverting this example:
```suggestion
        jubilant wait 'jubilant.all_active(status, "snappass-test")'
```

jubilant/__main__.py
```diff
@@ -0,0 +1,197 @@
...
+        '--error',
+        default=None,
+        help='The Python expression for the error condition.',
+    )
+
+    wait_parser.add_argument(
+        '--delay',
+        type=float,
+        default=1.0,
+        help='Delay in seconds between status calls. Default: 1.0',
```
benhoyt: You can use `%(default)s` (I think) to automatically show the default value, to avoid the duplication.

jubilant/__main__.py
```diff
@@ -0,0 +1,197 @@
...
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+
+    wait_parser.add_argument(
+        'ready',
+        help='The Python expression for the ready condition.',
+    )
+
+    wait_parser.add_argument(
+        '--error',
```
benhoyt: I'd specify the `--` args in alphabetical order (delay, error, successes, timeout) just for an easily-explainable defined order in the help.

jubilant/__main__.py
```diff
@@ -0,0 +1,197 @@
...
+        '--successes',
+        type=int,
+        default=3,
+        help='Number of times `ready` must evaluate to True for the wait to succeed. Default: 3',
+    )
+
+    args = arg_parser.parse_args(argv)
+
+    if args.command == 'version':
+        print(f'jubilant {jubilant.__version__}', file=sys.stderr)
```
benhoyt: The `version` subcommand's job is to print the version, so I think this should go to stdout (the default), not stderr. I also prefer dropping the "jubilant " prefix, as the version number is easier to consume if used in a script.

benhoyt (commented):
Nice, I really like this. I just added a few minor comments, mostly about the help output.

Please get another reviewer as well. Also, I only glanced at the tests, so if the other reviewer could review those, that would be great.

jubilant/_main.py
tonyandrewmeyer: Not just feels like, it is a special Python thing. It's what gets run when you do `python -m`. I don't think we intend people to do `python -m jubilant` so I would agree just `_main` is better here.

jubilant/__main__.py
tonyandrewmeyer: If this is coming from the CLI specification, I think those should be taken as "generally do this" rather than "you must always do this".

I think local time is better if you're running this locally, but UTC perhaps better in CI, but I think GitHub runners are all set to UTC anyway, so local==utc there.

jubilant/_main.py
```diff
@@ -0,0 +1,197 @@
...
+        configure_logging(logging.WARNING)
+    elif args.verbose:
+        configure_logging(logging.DEBUG)
+    else:
+        configure_logging(logging.INFO)
+
+    juju = jubilant.Juju(cli_binary=args.juju_cli_bin)
+
+    def _helper(expression: str) -> Callable[[jubilant.Status], bool]:
+        return lambda status: eval(  # noqa: S307
```
tonyandrewmeyer, line 151: As we discussed yesterday, the effective risk here is pretty low, because if you can run Jubilant you can probably just run Python and so do anything anyway.

However, I would still plug the `__builtins__` hole here. To avoid `eval("__builtins__['__import__']('os').system('ls')", {}, {})` and the like. Or if you'd prefer to not, then add a comment explaining why we don't believe it's an issue.

jubilant/__main__.py
tonyandrewmeyer: Stdout is the [gnu standard](https://www.gnu.org/prep/standards/standards.html#g_t_002d_002dversion) for what it's worth. I agree the number only form is nicer to parse, although it does seem very common to have the name as well.

tests/unit/test_main.py
```diff
@@ -0,0 +1,236 @@
...
+
+@pytest.mark.parametrize(
+    'argv_str',
+    [
+        pytest.param(
+            'wait',
+            id='missing_ready',
+        ),
+        pytest.param(
+            'wait --uknown-argument',
```
tonyandrewmeyer:
```suggestion
            'wait --unknown-argument',
```

Works with the typo, but I assume unintended :)

jubilant/__main__.py
```diff
@@ -0,0 +1,197 @@
...
+            delay=args.delay,
+            timeout=args.timeout,
+            successes=args.successes,
+        )
+    except jubilant.WaitError:
+        logger.error('Error expression evaluated to true (%s)', args.error)
+        return 1
+    except TimeoutError:
+        logger.error('Wait timed out after %s seconds', args.timeout)
+        return 1
```
tonyandrewmeyer: Perhaps 124? That's pretty widely used for timing out, not just by the `timeout` command, I think.

jubilant/__main__.py
```diff
@@ -0,0 +1,197 @@
...
+
+    if args.quiet:
+        configure_logging(logging.WARNING)
+    elif args.verbose:
+        configure_logging(logging.DEBUG)
+    else:
+        configure_logging(logging.INFO)
+
+    juju = jubilant.Juju(cli_binary=args.juju_cli_bin)
```
tonyandrewmeyer: Do you think we should have a `--model` arg, or just have people switch, and wait for feature requests?

jubilant/__main__.py
```diff
@@ -0,0 +1,197 @@
...
+    arg_parser.add_argument(
+        '--juju-cli-bin',
+        default='juju',
+        help='Path to the Juju CLI binary. Default: juju',
+    )
+
+    sub_parser = arg_parser.add_subparsers(dest='command')
+
+    # version subcommand
+    _ = sub_parser.add_parser(
```
tonyandrewmeyer: Style nit:

```suggestion
    sub_parser.add_parser(
```

tests/unit/test_main.py
```diff
@@ -0,0 +1,236 @@
...
+    ],
+)
+def test_exception_from_ready_expression(
+    expression: str,
+    mock_wait: MagicMock,
+) -> None:
+    def _helper(*args: Any, **kwargs: Any) -> None:
+        kwargs['ready'](MagicMock())
+
+    mock_wait.side_effect = _helper
```
tonyandrewmeyer: Minor note: this works, but it's testing eval expression errors through the wait mock's side_effect rather than through the actual code path. A more direct test would be to verify that the exception handler in main() catches Exception from the eval'd callable when Juju.wait() raises it.

jubilant/__main__.py
```diff
@@ -0,0 +1,197 @@
...
+            type(error).__name__,
+            str(error),
+        )
+        return 1
+
+    logger.info('Ready condition succeeded %d times (%s)', args.successes, args.ready)
+    return 0
+
+
+if __name__ == '__main__':
```
tonyandrewmeyer:
```suggestion
if __name__ == '__main__':  # pragma: no cover
```

This keeps coverage at 100% which is nice, although not a strict requirement.

jubilant/_main.py
```diff
@@ -0,0 +1,197 @@
...
+    Both ready and --error accept Python expressions. Those expressions have accesses to
+    the jubilant module, the jubilant.Juju instance, and the jubilant.Status object.
+
+    Set a timeout in seconds for this command with --timeout.
+
+    Set the delay in seconds between each Juju status query with --delay.
+
+    Set the number of consecutive successes the ready condition must have with --successes.
+
+    Examples:
```
tonyandrewmeyer, line 84: I didn't understand that this was one long paragraph at first.

```suggestion
    For example, the CLI version of these are:
```

And then a change in line 104.

Or maybe something like:

```
Examples:

    juju.wait(jubilant.all_active)  # CLI: jubilant wait jubilant.all_active
```

Or a small table, Python on the left, CLI on the right?

jubilant/__main__.py
```diff
@@ -0,0 +1,197 @@
...
+    """
+    wait_parser = sub_parser.add_parser(
+        name='wait',
+        description=textwrap.dedent(wait_description).strip(),
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+
+    wait_parser.add_argument(
+        'ready',
+        help='The Python expression for the ready condition.',
```
tonyandrewmeyer: Style nit:

```suggestion
        help='Python expression for the ready condition.',
```

jubilant/__main__.py
```diff
@@ -0,0 +1,197 @@
...
+    )
+
+    wait_parser.add_argument(
+        '--error',
+        default=None,
+        help='The Python expression for the error condition.',
```
tonyandrewmeyer: Style nit:

```suggestion
        help='Python expression for the error condition.',
```

jubilant/__main__.py
```diff
@@ -0,0 +1,197 @@
...
+        type=float,
+        default=180.0,
+        help='Overall timeout in seconds. Default: 180.0',
+    )
+
+    wait_parser.add_argument(
+        '--successes',
+        type=int,
+        default=3,
+        help='Number of times `ready` must evaluate to True for the wait to succeed. Default: 3',
```
tonyandrewmeyer: I assume it's not `True` but truth-y, and if so then our style is to use lower-case "true" for that.

jubilant/__main__.py
tonyandrewmeyer: Or [ArgumentDefaultsHelpFormatter](https://docs.python.org/3/library/argparse.html#argparse.ArgumentDefaultsHelpFormatter).

tests/unit/test_main.py
```diff
@@ -0,0 +1,236 @@
...
+    assert mock_wait.call_count == 1
+
+
+def test_parse_version(mock_wait: MagicMock) -> None:
+    """Test parsing the version command."""
+
+    exit_code = main(['version'])
+
+    assert exit_code == 0
+    assert mock_wait.call_count == 0  # Print version and exit.
```
tonyandrewmeyer, line 44: No need to assert on the actual version, but I think asserting that the output is the right format (int.int.int or name int.int.int or whatever it ends up as) would be worthwhile.

tests/unit/test_main.py
```diff
@@ -0,0 +1,236 @@
...
+        '/snap/juju',
+        'wait',
+        'juju.cli_binary=="/snap/juju"',
+    ])
+    assert exit_code == 0
+
+    wait_kwargs = mock_wait.call_args.kwargs
+    status = Status._from_dict(json.loads(MINIMAL_JSON))
+
+    assert wait_kwargs['ready'](status)
```
tonyandrewmeyer, line 230: Shouldn't this assert that the provided juju executable was used?

tests/unit/test_main.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+def test_juju_cli_bin_default(mock_wait: MagicMock) -> None:
+    exit_code = main([
+        'wait',
+        'juju.cli_binary=="juju"',
```
tonyandrewmeyer, line 236: This tests the default but by providing the default value. Wouldn't it be better if the test didn't pass the value and then asserted it was hat was used anyway?

jubilant/__main__.py
```diff
@@ -0,0 +1,197 @@
...
+        description='Show the version.',
+    )
+
+    wait_description: str = """
+    The wait command queries Juju status and checks that the ready condition succeeds
+    three times in a row.
+
+    Use --error to terminate the command early with a condition.
+
+    Both ready and --error accept Python expressions. Those expressions have accesses to
```
tonyandrewmeyer:
```suggestion
    Both ready and --error accept Python expressions. Those expressions have access to
```

jubilant/_main.py
```diff
@@ -0,0 +1,197 @@
...
+            timeout=args.timeout,
+            successes=args.successes,
+        )
+    except jubilant.WaitError:
+        logger.error('Error expression evaluated to true (%s)', args.error)
+        return 1
+    except TimeoutError:
+        logger.error('Wait timed out after %s seconds', args.timeout)
+        return 1
+    except Exception as error:
```
tonyandrewmeyer, line 170: I wonder if we should catch KeyboardInterrupt and provide a clean exit? Not sure on this one, go with your gut.

jubilant/_main.py
```diff
@@ -0,0 +1,197 @@
...
+    handler.setFormatter(formatter)
+    root_logger.addHandler(handler)
+    root_logger.setLevel(level)
+
+
+def main(argv: Sequence[str] | None = None) -> int:
+    """The main entrypoint."""
+    arg_parser = argparse.ArgumentParser('jubilant')
+
+    group = arg_parser.add_mutually_exclusive_group()
```
tonyandrewmeyer, line 45: I wonder whether these ought to be on the wait command. They don't impact `version`, although I guess they could. It was initially surprising that I could tack `--verbose` on to the end of my command, although quickly obvious why not.

If we add some third command in the future, possibly it would also support logging levels, and maybe they are really a "jubilant" config not a "jubilant wait" config, and it just feels like that because one of the main things wait does is log.

I'm not sure here - I think I probably lean towards what you already have, but bringing it up just in case.

tonyandrewmeyer (approved):
Really nice! I tried this out with a few commands and everything went as expected, and it was pretty convenient to use.

A few comments / questions / suggestions, but approving optimistically.

jubilant/_main.py
benhoyt: I'm not sure what is the specific hack Tony had in mind, but Python is very hard to sandbox. For example, you can access the builtins from basically any class, like this:

```
>>> jubilant.Juju.__init__.__globals__['__builtins__']['dir']
<built-in function dir>
```

So I wonder if trying to lock things down "a little" is basically security theatre.

jubilant/__main__.py
benhoyt: I think `--model` is reasonable, actually. +1

jubilant/__main__.py
tromai: Ah thanks. I have removed UTC time formatting, and rely on local time.

jubilant/__main__.py
tromai: Thanks. I agree we should print to stdout. I change the content to include only the version number too.

tests/unit/test_main.py
tromai: Thanks for the catch. It's indeed unintended, I have fixed it.

`--uknown-argument` is also an "unknown" argument :rofl:

jubilant/_main.py
tromai: I agree, keeping them separated is quite confusing. I went with your first suggestion:

```
        juju.wait(jubilant.all_active)  # CLI: jubilant wait jubilant.all_active
        juju.wait(
            lambda status: jubilant.all_active(status, 'snappass-test'),
            error=jubilant.any_error,
        )  # CLI: jubilant wait 'jubilant.all_active(status, "snappass-test")'
```

tests/unit/test_main.py
tromai: That's a good idea. I have added this in

```
    # Assert only the version string exists in stdout.
    version_pattern = r'^\d+\.\d+\.\d+$'
    assert re.match(version_pattern, captured.out.strip())  # strip the newline from print
```

jubilant/_main.py
tromai: Thanks for the suggestion. I decided to not catch `KeyboardInterrupt` here for now.

tests/unit/test_main.py
tromai: This also answers https://github.com/canonical/jubilant/pull/374#discussion_r3642566950.

This test asserts the provided juju executable was used, but in a 'round-trip' approach.

The test runs

```
main([
    '--juju-cli-bin',
    '/snap/juju',
    'wait',
    'juju.cli_binary=="/snap/juju"',
])
```

The `juju` instance is initialized with cli binary `/snap/juju`. 

```
juju = jubilant.Juju(cli_binary=args.juju_cli_bin, model=args.model)
```

This `juju` instance is then included in the context of `ready`.

```
lambda status: eval(  # noqa: S307
    expression,
    {'jubilant': jubilant, 'juju': juju, 'status': status},
```

In the test, I obtained this `ready` lambda and evaluated it. `'juju.cli_binary=="/snap/juju"'` gave us whether the `juju` instance used the provided juju executable. 

This is the same for the `test_juju_cli_bin_default`.

I agree that this approach can be confusing, I could also create a mock for `Juju`'s constructor, and assert the passed in parameters. Let me know if that is okay too.

jubilant/_main.py
tromai: Thanks for the discussion.

I agree that:
- The risk here is very low.
- It's very hard to sandbox in Python.

With this feature, the context (e.g. juju model, the status object, etc.) provided to `eval()` were designed to communicate with users about what they can depend on, rather than security concerns. We expect users not to rely on other modules.

I understand that we can go a bit further, and plug the `__builtins__` hole so that they couldn't even try to import extra modules. But I think it wouldn't make sense for a user to have such workaround like `eval("__builtins__['__import__']('os').system('ls')", {}, {})`. We will advise against it, but cannot ban someone from overcoming the `eval()` "sandbox" we have (they can just easily run another Python script in the same environment for example) :thinking:. 

With that, I will add a comment explaining why this is not an issue: it was added in https://github.com/canonical/jubilant/pull/374/commits/663c456b20e0d76dec7ed1065cac7d17b44e1011

jubilant/__main__.py
tromai: The `--model` flag has been added in the latest version :rocket:

jubilant/_main.py
tromai: Thanks for bringing this up. 

I purposely put `--verbose` here because I intended log verbosity to be a Jubilant CLI thing. So if we add a third command, then that command can inherit the logging configuration, and only need to decide on what level they are logging from.

> It was initially surprising that I could tack --verbose on to the end of my command, although quickly obvious why not.

Interesting, do you mean something like `jubilant wait True --verbose` ?

The `version` command seems like an exception as you say, as it doesn't depend on logging configuration. I could make it depend on the "global" logging configuration, but I don't think it would bring much benefit. `version` prints to stdout, instead of stderr, and doesn't include timestamps. I'm happy to keep `version` as it is.

tests/unit/test_main.py
tromai: > A more direct test would be to verify that the exception handler in main() catches Exception from the eval'd callable when Juju.wait() raises it.

I attempted this approach in my latest changes

```
def test_exception_from_ready_expression(
    expression: str,
    monkeypatch: pytest.MonkeyPatch,
) -> None:
    def call_ready(*args: Any, **kwargs: Any) -> None:
        kwargs['ready'](MagicMock())

    monkeypatch.setattr('jubilant.Juju.wait', call_ready)

    assert main(['wait', expression]) != 0
```

I hope it aligns with what you had in mind. I removed `side_effect`, and instead patch `jubilant.Juju.wait` directly with a function that just evaluate the `ready` lambda. This tests what I wanted to:
- The Exception is raised within the ready expression.
- The exception handler in `main()` catches that expression.

jubilant/_main.py
```diff
@@ -49,14 +38,20 @@ def configure_logging(level: int) -> None:
...
-        action='store_true',
-        help='Increase verbosity.',
+    arg_parser.add_argument(
+        '--model',
+        help='The Juju model to operate on, otherwise use the current Juju model.',
+    )
+
+    arg_parser.add_argument(
+        '--juju-cli-bin',
+        help='Path to the Juju CLI binary.',
```
benhoyt: Let's do two things on all of these argument help, for consistency and because they're not actually sentences: lowercase the first letter (consistent with `--help` help) and no full stop at the end (because they're not full sentences).

jubilant/_main.py
benhoyt: Sorry to be a pain, but I find the example quite confusing now. It's help/examples for the CLI tool, so it's weird to have the Python syntax first and the CLI syntax as a comment on the right (which easily gets lost). In addition, the `--error` flag seems to be missing on the second example. How about:

```
Examples:

    jubilant wait jubilant.all_active

    jubilant wait 'jubilant.all_active(status, "snappass-test")' --error jubilant.any_error

These are equivalent to the following Python calls:

    juju.wait(jubilant.all_active)

    juju.wait(
        lambda status: jubilant.all_active(status, 'snappass-test'),
        error=jubilant.any_error,
    )
```

benhoyt (approved):
Approving optimistically, but two small comments on the help output.

jubilant/_main.py
tromai: Thanks! 

I have fixed them. There are exceptions: `Python expression for the ready condition` and `Python expression for the error condition...` still keep the capitalized `P`.

jubilant/_main.py
benhoyt: @tromai Thanks for fixing the `help` messages. Just making sure you saw this comment ^ as it was marked resolved earlier.

tromai: Thanks for the reminder! I indeed missed this, sorry.

@benhoyt I really like the idea of adding new lines between the examples, and putting the CLI version first.

Although I originally prefer a setup where we can view both versions of the example side by side, `These are equivalent to the following Python calls:` already indicates that clearly. I'm happy to go with your suggestion.

@tonyandrewmeyer I think this version aligns with your first suggestion.

**EDIT**: I modify the first example a bit

```
    Examples:

        jubilant wait 'jubilant.all_active(status)'

        ...

    These are equivalent to the following Python calls:

        juju.wait(jubilant.all_active)

        ...
```

This is because `jubilant wait jubilant.all_active` would evaluate `jubilant.all_active` (as a function object, not an invocation) and jubilant wait will succeed.

For example, these 2 commands are run right after another:

```
> jubilant wait jubilant.all_active
2026-08-12 08:50:11,070 [fastapi-demo] status: active ()
2026-08-12 08:50:11,070 [fastapi-demo/0] status: active ()
2026-08-12 08:50:13,419 Ready condition succeeded 3 times (jubilant.all_active)

> jubilant wait jubilant.all_blocked
2026-08-12 08:50:17,112 [fastapi-demo] status: active ()
2026-08-12 08:50:17,112 [fastapi-demo/0] status: active ()
2026-08-12 08:50:19,499 Ready condition succeeded 3 times (jubilant.all_blocked)
``` 

For the second example, I rename `snappass-test` to `app` to fit the 99 characters limit. I couldn't use `# noqa` since it is in a multi-line string. Let me know if that looks good.

tests/unit/test_main.py
benhoyt: Addressed by comment above.

jubilant/_main.py
```diff
@@ -0,0 +1,201 @@
...
+    root_logger.setLevel(level)
+
+
+def main(argv: Sequence[str] | None = None) -> int:
+    """The main entrypoint."""
+    arg_parser = argparse.ArgumentParser(
+        'jubilant',
+    )
+
+    group = arg_parser.add_mutually_exclusive_group()
```
james-garner-canonical: Nit: `group` is just the type of thing, and doesn't tell us what's being grouped. How about `logging_group` or `verbosity_group`? I'd also put the declaration of a parser/group and its args together, like:
```py
parser = argparse.ArgumentParser()
parser.add_argument(...)

some_group = parser.add_mutually_exclusive_group()
some_group.add_argument(...)
```

jubilant/_main.py
```diff
@@ -0,0 +1,201 @@
...
+        help='suppress all output except errors',
+    )
+
+    group.add_argument(
+        '--verbose',
+        action='store_true',
+        help='increase verbosity',
+    )
+
+    sub_parser = arg_parser.add_subparsers(dest='command')
```
james-garner-canonical: Likewise here, the variable name only tells us what kind of thing this is, not what it's for. But also it's not a subparser, is an object that manages multiple subparsers. How about `cmd_parsers` or `cmd_subparsers`?

jubilant/_main.py
```diff
@@ -0,0 +1,201 @@
...
+    )
+
+    args = arg_parser.parse_args(argv)
+
+    if args.command == 'version':
+        print(jubilant.__version__)
+        return 0
+    elif args.command != 'wait':
+        arg_parser.print_usage(file=sys.stderr)
+        return 1
```
james-garner-canonical: How about using `add_subparsers(dest='command', required=True)`?

jubilant/_main.py
```diff
@@ -0,0 +1,201 @@
...
+            timeout=args.timeout,
+            successes=args.successes,
+        )
+    except jubilant.WaitError:
+        logger.error('Error expression evaluated to true (%s)', args.error)
+        return 1
+    except TimeoutError:
+        logger.error('Wait timed out after %s seconds', args.timeout)
+        return 124
+    except Exception as error:
```
james-garner-canonical: Worth gracefully handling `KeyboardInterrupt` with a clean error logging message and exit code?

jubilant/_main.py
```diff
@@ -0,0 +1,201 @@
...
+        logger.error(
+            'Exception evaluating "ready" or "error": %s: %s',
+            type(error).__name__,
+            str(error),
```
james-garner-canonical: I wonder if we should show the traceback here? For example `logger.exception` instead of `logger.error`.

My thinking is that this branch is really for "something went wrong, we don't know what", so it would be nice to give the user the information they need to figure out what went wrong and fix their error (if it was their error) or open an issue elsewhere. Our current error message assumes it's something wrong with their `ready` or `error` argument, but it could be any other funky error state (e.g. something goes wrong when we get `juju.status`).

If we really don't want to show the traceback, maybe we should at least use `%r` for the exception instead of showing only the `type` and `str`.

jubilant/_main.py
```diff
@@ -0,0 +1,201 @@
...
+        )
+        return 1
+
```
james-garner-canonical, line 187: Worth also gracefully handling `KeyboardInterrupt` (which inherits from `BaseException` rather than `Exception`) with a clean error logging message and exit code? This would currently get you the full traceback, which is arguably undesirable in the case of an intentional "stop waiting now".

jubilant/_main.py
```diff
@@ -0,0 +1,201 @@
...
+        juju.wait(
+            ready=_helper(args.ready),
+            error=_helper(args.error) if args.error else None,
+            delay=args.delay,
+            timeout=args.timeout,
+            successes=args.successes,
+        )
+    except jubilant.WaitError:
+        logger.error('Error expression evaluated to true (%s)', args.error)
+        return 1
```
james-garner-canonical, line 177: I wonder if we should use different exit codes to distinguish between: 1) your error condition was reached, 2) something mysterious went wrong.

james-garner-canonical (commented):
Driveby review:
1) A couple of argparse nits.
2) A couple of error logging / exit behaviour questions.

jubilant/_main.py
tromai: Oh, good point. Thanks for pointing this out.

jubilant/_main.py
tromai: Thanks for the detailed suggestion. 

A while ago we decided to remove the stacktrace from the output (https://github.com/canonical/jubilant/pull/374#discussion_r3592265461). 

This block

```
except Exception as error:
    logger.error(
        'Exception evaluating "ready" or "error": %r',
        error,
    )
    return 1
```

Is meant to capture any exception raised while "waiting", in a generic manner. May be the error message is a bit ambiguous, I can change it to `Exception while evaluating "ready" or "error"`.

I agree that it would be great to clearly show where the exception comes from (the Python expression, or elsewhere in the code base). We cannot scope down what Exception will be raised from the `error` or `ready` expression. But we can somewhat distinguish exception raised from Jubilant `wait` method.
- `WaitError` and `TimeoutError`: are already handled
- `CLIError` from `self._cli('status', '--format', 'json', log=False)`: is not handled. I think it's good to add its own catch block
- JSON encoding and decoding exception from `result = json.loads(stdout)`: is not handled. Also good to add its own catch block, since `ready` and `error` expression cannot use the `json` module.

Let me know what you think.

tromai: And good idea on using `%r`. Thanks!

jubilant/_main.py
tromai: This was actually raised by Tony in (https://github.com/canonical/jubilant/pull/374#discussion_r3642589228), but I decided back then not to do it (I was on a fence too).

Since you also suggest, it's time to do it :smile:

jubilant/_main.py
tromai: I didn't have a strong argument to have different exit codes, so I didn't put them here originally :thinking: 

If you strongly suggest, I'm happy to change it.

jubilant/_main.py
james-garner-canonical: Thanks for explaining. If we've already decided not to show the traceback in this case, then that's all good. We can probably get away without error branches for all the different cases for the initial release and see what comes up in practice.

For the message, making it more generic sounds good. Maybe something like "Exception while running juju.wait" or "Exception while waiting for <model>". Mentioning "ready" or "error" seems to point the reader to those specific user-supplied statements, when that might not be the case -- and hopefully the error message points them in that direction if it is the case (e.g. `NameError`, `SyntaxError`, etc).

jubilant/_main.py
james-garner-canonical: Currently we're using 124 for `TimeoutError`. Unsure why. Doesn't seem like there's a clear standard for exit codes. Maybe we should just use 1 for everything initially and reserve the right in the documentation to change exit code 1 to a more specific version in a feature release (that is, users should test `!= 0` or `> 0`, not `== 1` to avoid breakage).

tromai: `124` for `TimeoutError` was suggested here (https://github.com/canonical/jubilant/pull/374#discussion_r3642321295). It follows the status code of `timeout` command. I thought that it's okay to follow as it's part of coreutils. 

```
$ timeout 1s sleep 10
$ echo $?
124
```

I totally agree that we should advise users to test for `!= 0` or `> 0`. The error message would be sufficient to describe the error it encountered. There is also one unit test `test_wait_error_exits_nonzero` where I asserted `== 1`, that should go away too.

james-garner-canonical: Thanks for explaining, sounds reasonable. Maybe:
```py
        return 124  # The same exit code coreutils timeout uses.
```

james-garner-canonical (approved):
LGTM, thanks for addressing my unsolicited feedback! Will be very cool to have this feature released.

jubilant/_main.py
tonyandrewmeyer: > I purposely put `--verbose` here because I intended log verbosity to be a Jubilant CLI thing. So if we add a third command, then that command can inherit the logging configuration, and only need to decide on what level they are logging from.

Yeah, seems reasonable. It's hard to know if future commands will make sense for a verbose setting, but it's a common enough pattern for the level to be at the top, so doing that here is fine.

> > It was initially surprising that I could tack --verbose on to the end of my command, although quickly obvious why not.
> 
> Interesting, do you mean something like `jubilant wait True --verbose` ?

Sorry! A critical "not" got left out of my sentence... I meant to say "it was initially surprising that I could **not** tack --verbose`".

tromai: Thanks @tonyandrewmeyer . That makes sense.

I had a relevant discussion with Ben about this too. We were thinking whether supporting `--verbose` both before and after the `wait` command is easy enough:

- `jubilant --verbose wait expr`
- `jubilant wait expr --verbose`

However, it is not a trivial change to support this. `argparse` doesn't play nicely when you have the same argument name both in the root parser, and in the `wait` subparser. I have tried a couple of approaches but they all needed extra workaround, and not as simple as I hope. I will merge the PR as is. If having `--verbose` before `wait` proves to be inconvenient, I can move it to be part of `wait` :thinking: :

- `jubilant --verbose wait expr` - No
- `jubilant wait expr --verbose` - Yes
- `jubilant wait  --verbose expr`- Yes

@benhoyt Please let me know what you think.

james-garner-canonical: I'd lean towards defining the verbosity arguments for each subparser that it makes sense for rather than having it on the base parser, as that gives more flexibility as to where the argument can be provided. You could have an add_verbosity_args(subparser) helper to point future subparsers in the right direction.

tromai: True. Thinking about it again, it is obvious that: we instinctively expect `--verbose` to work at the end of the command, after `wait`. It's certainly more user-friendly to go with what's natural to us.

Thanks everyone for helping me comes to realise this in the end :sweat_smile: .

jubilant/_main.py
```diff
@@ -36,6 +36,20 @@ def configure_logging(level: int) -> None:
     root_logger.setLevel(level)
 
 
+def _add_verbosity_args(subparser: argparse.ArgumentParser) -> None:
+    verbosity_group = subparser.add_mutually_exclusive_group()
```
tromai, line 40: The other arguments in `wait` subparser are defined in alphabetical order. `--verbose` and `--quiet` won't be in the alphabetical order, but I see them as an acceptable exception.

james-garner-canonical (commented):
I wonder if ` --model` and `--juju-cli-bin` should get the same treatment.

james-garner-canonical:
Current help output:
```
> uv run jubilant -h
usage: jubilant [-h] [--model MODEL] [--juju-cli-bin JUJU_CLI_BIN] {version,wait} ...

positional arguments:
  {version,wait}

options:
  -h, --help            show this help message and exit
  --model MODEL         the Juju model to operate on, otherwise use the current Juju model
  --juju-cli-bin JUJU_CLI_BIN
                        path to the Juju CLI binary
```
```
> uv run jubilant wait -h
usage: jubilant wait [-h] [--delay DELAY] [--error ERROR] [--successes SUCCESSES] [--timeout TIMEOUT] [--quiet | --verbose] ready

...

positional arguments:
  ready                 Python expression for the ready condition

options:
  -h, --help            show this help message and exit
  --delay DELAY         delay in seconds between status calls (default: 1.0)
  --error ERROR         Python expression for the error condition (default: None)
  --successes SUCCESSES
                        number of times `ready` must evaluate to true for the wait to succeed (default: 3)
  --timeout TIMEOUT     overall timeout in seconds (default: 180.0)
  --quiet               suppress all output except errors
  --verbose             increase verbosity
```

tromai:
> I wonder if ` --model` and `--juju-cli-bin` should get the same treatment.

These arguments will be used to initialize the juju object in Jubilant

```
juju = jubilant.Juju(cli_binary=args.juju_cli_bin, model=args.model)
```

I originally viewed `juju` as a "root" thing, where each command like `wait` relies on. However, this is no longer true. `wait` relies on them, `version` doesn't. And since we agreed to put `--verbose` and `--quiet` to `wait`, ` --model` and `--juju-cli-bin` should get the same treatment.

I will move it.

EDIT: and that would help with the help output too. That's a good point. Thanks!

tromai:
@james-garner-canonical  There is one concern, sorry. I wanted the CLI to thinly wrap around `juju.wait()` API. Does having `--model` and `--juju-cli-bin` part of `wait` defeat that purpose and/or confuse users ? Because `juju.wait` doesn't have any arguments to set model or juju cli binary.

```
jubilant wait --juju-cli-bin /snap/juju True
```

jubilant/_main.py
```diff
@@ -0,0 +1,193 @@
...
+    _add_juju_args(wait_parser)
+    _add_verbosity_args(wait_parser)
+
+    args = arg_parser.parse_args(argv)
+
+    if args.command == 'version':
+        print(jubilant.__version__)
+        return 0
+
+    if args.quiet:
```
james-garner-canonical, line 142: I might factor from here down into a separate function for wait since args like quiet or juju are now explicitly part of that subparser. But perhaps not necessary while there are only the two commands.

james-garner-canonical (approved):
Looks great!

jubilant/_main.py
tromai: I'm not against it. 

I think we could re-visit this when we add another command to the CLI :thinking: (`version` is a proper command but it doesn't do much compared to `wait`, edit: so keeping wait in `main()` looks okay to me for now).

PR #383, opened 2026-07-23
docs: extract howto guides from the tutorial

tromai:
Closes #279.

This PR extracts the `fall back to cli commands` and `custom wait condition` from the tutorial page into its own howto guides. I also added `html_metadata` into the newly created pages, and updated the indexes.

There is no change to the original contents.

PR #396, opened 2026-08-16
docs: update juju.wait doc string; update title of run-juju-cli-command how-to page; update juju.cli example

tromai:
This PR made minor doc related updates:
- In `juju.wait` doc string, the link to `How to use custom wait condition` is updated to the new location `https://canonical.com/juju/docs/jubilant/how-to/use-a-custom-wait-condition` after #383.
- Rename the title of `docs/how-to/run-juju-cli-commands.md` to `How to run other Juju CLI commands`

The current example in `How to run other Juju CLI commands` uses `model-config`, which was added in #106 . This PR replaces that example:
- With a juju command relevant to charm integration test (for example, found in a real charm).
- "model-scoped". `show-controller` is great, but it needs `include_model=False`, which is covered in the second example of the page.

I went with `scale-application`. This is used in some real charms, but we don't have a useful returned JSON data:
- [oauth2-proxy-k8s-operator](https://github.com/canonical/oauth2-proxy-k8s-operator/blob/0052696dc1858fdd0a5b29d409eaeb8ef402cd86/tests/integration/test_charm.py#L166)
- [loki-k8s-operator](https://github.com/canonical/loki-k8s-operator/blob/e87a761693c5d38b8bd4abb48d2ea7ea83807d00/tests/integration/test_alert_rules_fire.py#L70)

Please let me know if this works.

There is also a "paradox": a juju command not in Jubilant -> it is widely use as `juju.cli` -> more chance will be added to Jubilant :laughing:

### DIFF
--- a/docs/how-to/run-juju-cli-commands.md
+++ b/docs/how-to/run-juju-cli-commands.md
@@ -9,16 +9,13 @@ myst:
 
 Many common Juju commands are already defined on the `Juju` class, such as [`deploy`](jubilant.Juju.deploy) and [`integrate`](jubilant.Juju.integrate). However, if you want to run a Juju command that's not yet defined in Jubilant, you can fall back to calling the [`Juju.cli`](jubilant.Juju.cli) method.
 
-For example, to fetch a model configuration value using `juju model-config`:
+For example, to scale a Kubernetes application using `juju scale-application`:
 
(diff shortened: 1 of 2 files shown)


benhoyt (approved):
Looks good to me -- thanks.

PR #399, opened 2026-08-26
docs: cover the remaining SEC0030 sub-requirements

tonyandrewmeyer:
A re-read of the security page against SEC0030 v1.3 turned up two sub-requirements not done: logging and monitoring item 4 (alerts, opting out, masking), and security lifecycle item 3 (scheduling or postponing an update, with a warning about the risk of delaying).

Lifecycle item 3 is the main gap: the page covered receiving, detecting and verifying an update, but not scheduling or postponing one, and had no risk warning.

Most of what item 4 it asks for doesn't apply here (there's no monitoring agent and no alert to configure). What the paragraph is actually for is the two things that don't go through the `jubilant` logger, `CLIError` tracebacks and `debug_log()` output, and scoping the bit about handlers to library use, since the `jubilant` command attaches its own to the root logger.

### DIFF
--- a/docs/explanation/security.md
+++ b/docs/explanation/security.md
@@ -36,7 +36,7 @@ The library does not introduce any new security risks beyond directly running Ju
 
 Jubilant does not use any cryptographic technology, hashing, or digital signatures. All cryptographic operations (TLS, credential storage, API authentication) are handled by the Juju CLI and the Juju controller.
 
-There is correspondingly nothing to configure for data in transit or at rest. Jubilant opens no network connections, so the TLS protecting a controller connection is the CLI's to enable and configure. It stores nothing persistently either; the one thing it writes to disk is a temporary file used to pass a secret to the CLI without exposing it in the process arguments, deleted when the call returns. That file goes in the system temporary directory, or in `~/snap/juju/common` when the Juju CLI is a snap, since a confined snap cannot read `/tmp`.
+There is correspondingly nothing to configure for data in transit or at rest. Jubilant opens no network connections, so the TLS protecting a controller connection is the CLI's to enable and configure. Nothing is stored persistently: the temporary files described under [Product architecture](#product-architecture) are the only disk writes, and each is removed when the call returns.
 
 ## Configuring and operating
 
--- a/docs/explanation/security.md
+++ b/docs/explanation/security.md
@@ -36,8 +36,6 @@ The library does not introduce any new security risks beyond directly running Ju
 
 Jubilant does not use any cryptographic technology, hashing, or digital signatures. All cryptographic operations (TLS, credential storage, API authentication) are handled by the Juju CLI and the Juju controller.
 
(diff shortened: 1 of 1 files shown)


docs/explanation/security.md
```diff
@@ -59,6 +59,8 @@ For security-relevant events (login failures, credential errors, controller acce
 
 Secret and credential values are written to temporary files rather than CLI arguments, so they do not appear in Jubilant's logs. The exception is `CLIError`, raised on a Juju CLI failure: its traceback includes the CLI's `stdout` and `stderr`, so any value the Juju CLI itself echoes there would be exposed.
 
+In a test suite, everything Jubilant emits goes through the `jubilant` logger, so that handler is the only control point: raise its level to capture less, or configure no handler at all to opt out entirely. The `jubilant` command-line tool is different, and attaches its own handler to the root logger at a level set by `-q`/`-v`. There are no alerts to configure and no masking beyond the temporary files above, so a test is responsible for keeping secrets out of its own log messages, and for not printing a `CLIError` traceback or `debug_log()` output somewhere the output is kept: neither passes through the logger.
```
dwilding: Some suggestions for readability. I'm thinking it's good to split the paragraph because otherwise the "There are no alerts ..." could be read as only applying to the `jubilant` CLI tool.

```suggestion
In a test suite, everything Jubilant emits goes through the `jubilant` logger. Raise its level to capture less, or remove the handler to opt out. The `jubilant` command-line tool is different, and attaches its own handler to the root logger at a level set by `-q`/`-v`.

There are no alerts to configure and no masking beyond the temporary files described above. Each test is responsible for keeping secrets out of log messages, and for not printing a `CLIError` traceback or `debug_log()` output somewhere the output is kept — both bypass the `jubilant` logger.
```

docs/explanation/security.md
```diff
@@ -71,6 +73,8 @@ Jubilant follows [semantic versioning](https://semver.org/). Security updates ar
 
 **Detecting available updates.** Configure Dependabot or Renovate in your charm repository so that security updates are surfaced automatically as pull requests. Re-lock dependencies when prompted so that the charm tests run with the latest version.
 
+**Scheduling and postponing updates.** An update reaches you when you re-lock, so re-lock on a cadence rather than when someone happens to notice. Pinning to an exact version, or turning off the automated proposals above, postpones updates indefinitely: a published fix will not reach you until the pin is lifted. Jubilant runs with the credentials of whoever runs the tests, and against real controllers, so a machine or CI runner on a known-vulnerable version stays exposed for the whole of that delay.
```
dwilding:
```suggestion
**Scheduling and postponing updates.** An update reaches you when you re-lock, so re-lock regularly rather than when someone notices a new version. Pinning to an exact version, or turning off the automated proposals above, postpones updates indefinitely. Jubilant runs with the credentials of whoever runs the tests, and against real controllers, so a machine or CI runner on a known-vulnerable version stays exposed until you re-lock.
```

dwilding (approved):
A few small suggestions, but looks good, thanks

PR #413, opened 2026-09-07
docs: add how-to use wait from CLI

### LINKED ISSUE #382: Add a howto guide for Jubilant wait CLI
Following https://github.com/canonical/jubilant/issues/323, we need to add a howto guide for this feature.

The guide should focus on how to use the CLI:
- Used from a project that has Jubilant as a dev dependency. Used as a standalone CLI.
- Ready and Error conditions are Python expressions.
- The modules available in `eval()`'s scope.
- How to set the juju binary and the model name.
- Successes and timeout.
- A practical example of deploying a charm, and use the CLI to wait for a condition.

tromai:
Closes #382.

This PR adds a new how-to page for the `Juju.wait` CLI. The feature was added in #374 and available from Jubilant v1.13.0.

[Preview](https://canonical-juju-jubilant--413.com.readthedocs.build/how-to/use-wait-from-cli/).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -299,6 +299,7 @@ html_js_files = [
 intersphinx_mapping = {
     "juju": ("https://documentation.ubuntu.com/juju/3.6/", None),
     "operator": ("https://canonical.com/juju/docs/ops/latest/", None),
+    "python": ("https://docs.python.org/3", None),
 }
 
 
--- a/docs/how-to/index.md
+++ b/docs/how-to/index.md
@@ -28,3 +28,13 @@ Run Juju commands that aren't yet defined in Jubilant.
 
 Run Juju CLI commands <run-juju-cli-commands>
 ```
+
+## Using `Juju.wait` from the command line
+
+Run `Juju.wait` from the command line without code.
+
+```{toctree}
+:maxdepth: 1
+
+Use wait from CLI <use-wait-from-cli>
+```
--- a/docs/how-to/use-wait-from-cli.md
+++ b/docs/how-to/use-wait-from-cli.md
@@ -30,13 +30,13 @@ integration = [
 
 You can run it from the root of that project with `uv`:
 
-```
+```text
 uv run --group integration jubilant wait --help
 ```
 
 You can also run it as a standalone CLI:
 
-```
+```text
 uvx --from 'jubilant>=1.13.0' jubilant wait --help
 ```
 
@@ -48,8 +48,9 @@ The `ready` and `--error` CLI arguments are passed as Python expressions, differ
 
 For example, this CLI invocation:
 
-```
-jubilant wait 'jubilant.all_active(status, "myapp")' --error 'jubilant.any_error(status)'
+```text
+jubilant wait 'jubilant.all_active(status, "myapp")' \
+    --error 'jubilant.any_error(status)'
 ```
 
 is equivalent to the following Python call:
@@ -65,7 +66,7 @@ juju.wait(
 
 By default, the CLI uses the `juju` binary on your `PATH` and operates on the current Juju model. To override either of these, use `--juju-cli-bin` and `--model`:
 
-```
+```text
 jubilant wait 'jubilant.all_active(status)' \
     --juju-cli-bin /snap/bin/juju \
     --model mymodel
@@ -77,7 +78,7 @@ See first: {external+operator:ref}`Configure Jubilant logs <write-integration-te
 
 By default, `uvx jubilant wait` follows brief logging mode:
 
-```
+```text
 $ jubilant wait 'jubilant.all_active(status)'
 
 2026-09-08 01:31:43,040 [snappass-test] status: active (snappass started)
@@ -86,7 +87,8 @@ $ jubilant wait 'jubilant.all_active(status)'
 ```
 
 Use `--quiet` to suppress all output except errors:
-```
+
+```text
 $ jubilant wait 'jubilant.all_blocked(status)' --timeout 2.0 --quiet
 
 2026-09-08 01:33:47,550 Wait timed out after 2.0 seconds
@@ -94,7 +96,7 @@ $ jubilant wait 'jubilant.all_blocked(status)' --timeout 2.0 --quiet
 
 Or `--verbose` to enable verbose logging mode:
 
-```
+```text
 $ jubilant wait 'jubilant.all_active(status)' --verbose
 
 2026-09-08 01:34:25,406 INFO jubilant.wait [snappass-test] status: active (snappass started)
@@ -131,7 +133,7 @@ $ jubilant wait 'jubilant.all_active(status)' --verbose
 
 The `--delay`, `--timeout`, and `--successes` arguments match the corresponding parameters on [`Juju.wait`](jubilant.Juju.wait). For example:
 
-```
+```text
 jubilant wait 'jubilant.all_active(status)' \
     --delay 2 \
     --timeout 600 \
--- a/docs/how-to/use-wait-from-cli.md
+++ b/docs/how-to/use-wait-from-cli.md
@@ -1,0 +1,159 @@
+---
+myst:
+  html_meta:
+    description: Use Juju.wait from the Command-Line Interface.
+---
+
+(use_wait_from_cli)=
+# How to use `Juju.wait` from CLI
+
+See first: {ref}`use_a_custom_wait_condition`
+
+Jubilant provides an entrypoint to run [`wait`](jubilant.Juju.wait) from the CLI.
+
+```{note}
+This is only available from Jubilant v1.13.0 onwards.
+```
+
+## Install and run the CLI
+
+If your project defines Jubilant in a dependency group:
+
+```toml
+# pyproject.toml
+[dependency-groups]
+# Dependencies of integration tests
+integration = [
+    "jubilant>=1.13,<2"
+]
+```
+
+You can run it from the root of that project with `uv`:
+
+```
+uv run --group integration jubilant wait --help
+```
+
+You can also run it as a standalone CLI:
+
+```
+uvx --from 'jubilant>=1.13.0' jubilant wait --help
+```
+
+The rest of this guide will not show the `uv` parts in the sample commands. Please use the suitable one for your setup.
+
+## Ready and error expressions
+
+The `ready` and `--error` CLI arguments are passed as Python expressions, different from how [`Juju.wait`](jubilant.Juju.wait) is used in code. Those expressions have access to three variables: `jubilant` (the [`jubilant`](jubilant) module), `juju` (the [`jubilant.Juju`](jubilant.Juju) instance), and `status` (the [`jubilant.Status`](jubilant.Status) object).
+
+For example, this CLI invocation:
+
+```
+jubilant wait 'jubilant.all_active(status, "myapp")' --error 'jubilant.any_error(status)'
+```
+
+is equivalent to the following Python call:
+
+```python
+juju.wait(
+    lambda status: jubilant.all_active(status, 'myapp'),
+    error=jubilant.any_error,
+)
+```
+
+## Setting the Juju binary and model
+
+By default, the CLI uses the `juju` binary on your `PATH` and operates on the current Juju model. To override either of these, use `--juju-cli-bin` and `--model`:
+
+```
+jubilant wait 'jubilant.all_active(status)' \
+    --juju-cli-bin /snap/bin/juju \
+    --model mymodel
+```
+
+## Configure logging modes
+
+See first: {external+operator:ref}`Configure Jubilant logs <write-integration-tests-for-a-charm-configure-jubilant-logs>`
+
+By default, `uvx jubilant wait` follows brief logging mode:
+
+```
+$ jubilant wait 'jubilant.all_active(status)'
+
+2026-09-08 01:31:43,040 [snappass-test] status: active (snappass started)
+2026-09-08 01:31:43,040 [snappass-test/0] status: active (snappass started)
+2026-09-08 01:31:45,454 Ready condition succeeded 3 times (jubilant.all_active(status))
+```
+
+Use `--quiet` to suppress all output except errors:
+```
+$ jubilant wait 'jubilant.all_blocked(status)' --timeout 2.0 --quiet
+
+2026-09-08 01:33:47,550 Wait timed out after 2.0 seconds
+```
+
+Or `--verbose` to enable verbose logging mode:
+
+```
+$ jubilant wait 'jubilant.all_active(status)' --verbose
+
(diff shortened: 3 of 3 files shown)


docs/how-to/use-wait-from-cli.md
```diff
@@ -0,0 +1,161 @@
...
+
+(use_wait_from_cli)=
+# How to use `Juju.wait` from CLI
+
+See first: {ref}`use_a_custom_wait_condition`
+
+Jubilant provides an entrypoint to run [`wait`](jubilant.Juju.wait) from the CLI.
+
+```{note}
+This is only available from Jubilant v1.13.0 onwards.
```
benhoyt, line 15: In the rest of our docs we haven't done "this method is only available from version X", as it adds a lot of noise. They can just upgrade to the latest. So I would say drop this.

docs/how-to/use-wait-from-cli.md
```diff
@@ -0,0 +1,161 @@
...
+You can run it from the root of that project with `uv`:
+
+```text
+uv run --group integration jubilant wait --help
+```
+
+You can also run it as a standalone CLI:
+
+```text
+uvx --from 'jubilant>=1.13.0' jubilant wait --help
```
benhoyt, line 40: Two suggestions:

1) `uvx` is the simplest way to run this tool, so I'd say move this to the top of this section, like "The simplest way to run the Jubilant wait CLI is: ..." and then follow with the pyproject-based option.
2) I think we should drop the awkward `--from 'jubilant>=1.13.0'`. They're almost certain to get the latest version, and even if they have on older one cached they can easily update. It just makes the doc here look awkward / not simple.

docs/how-to/use-wait-from-cli.md
```diff
@@ -0,0 +1,161 @@
...
+juju.wait(
+    lambda status: jubilant.all_active(status),
+    delay=2.0,
+    timeout=600.0,
+    successes=5,
+)
+```
+
+## Exit codes
+
```
benhoyt, line 155: Add a one sentence intro here, something like "The wait CLI returns the following exit codes:"

benhoyt (approved):
Approving optimistically, but a few suggestions.

docs/how-to/use-wait-from-cli.md
james-garner-canonical: I'd suggest leading with `uv tool install jubilant` (perhaps mentioning `uv tool upgrade` or linking to https://docs.astral.sh/uv/concepts/tools), since the guide uses a bare `jubilant wait` throughout; then `uvx` as a tip; and drop the pyproject.toml project venv stuff entirely.

dwilding: I agree with leading with `uv tool install jubilant`, no need for `--from ...`, uvx as a tip, and dropping the `pyproject.toml` part.

How about making the "Install and run the CLI" section cover four things:

1. How to install - `uv tool install jubilant`
2. The general command format with `ready` and `--error`. That isn't mentioned explicitly at the moment
3. A usage example, and what the equivalent Python code would be. Essentially fold in the "Ready and error expressions" section and remove that section
4. The uvx tip (and possibly talk about upgrading?)

docs/how-to/use-wait-from-cli.md
```diff
@@ -0,0 +1,161 @@
...
+jubilant wait 'jubilant.all_active(status)' \
+    --juju-cli-bin /snap/bin/juju \
+    --model mymodel
+```
+
+## Configure logging modes
+
+See first: {external+operator:ref}`Configure Jubilant logs <write-integration-tests-for-a-charm-configure-jubilant-logs>`
+
+By default, `uvx jubilant wait` follows brief logging mode:
```
dwilding, line 79:
```suggestion
By default, `jubilant wait` follows brief logging mode:
```

docs/how-to/use-wait-from-cli.md
```diff
@@ -0,0 +1,161 @@
...
+
+2026-09-08 01:31:43,040 [snappass-test] status: active (snappass started)
+2026-09-08 01:31:43,040 [snappass-test/0] status: active (snappass started)
+2026-09-08 01:31:45,454 Ready condition succeeded 3 times (jubilant.all_active(status))
+```
+
+Use `--quiet` to suppress all output except errors:
+
+```text
+$ jubilant wait 'jubilant.all_blocked(status)' --timeout 2.0 --quiet
```
dwilding, line 92: Is the use of `all_blocked` important here? If `all_active` works too, I'd use that, to keep focus on the comparison between different verbosity levels.

docs/how-to/use-wait-from-cli.md
```diff
@@ -0,0 +1,161 @@
...
+is equivalent to the following Python call:
+
+```python
+juju.wait(
+    lambda status: jubilant.all_active(status, 'myapp'),
+    error=jubilant.any_error,
+)
+```
+
+## Setting the Juju binary and model
```
dwilding, line 65: I think I'd combine this section, "other arguments", and "exit codes" into a single section called "`wait` command reference". Containing a list of args and brief description of each arg - no need for additional examples, or show what the equivalent Python code would be, or link to Juju.wait again. Then the table of exit codes as you currently have it.

I think that would help keep the majority of the article focused on how to use the CLI in the (I assume) most common situations. Having a little block of reference at the end of the guide is fine, given that our Reference pages are reserved for the Python library reference.

docs/how-to/use-wait-from-cli.md
dwilding: Yeah, seems fine to drop this, especially if we focus on installed-tool usage.

docs/how-to/index.md
```diff
@@ -28,3 +28,13 @@ Run Juju commands that aren't yet defined in Jubilant.
 
 Run Juju CLI commands <run-juju-cli-commands>
 ```
+
+## Using `Juju.wait` from the command line
+
+Run `Juju.wait` from the command line without code.
```
dwilding, line 34: Except the code you have to pass to the CLI 😉 

In seriousness, I think this is fine. One alternative to consider:

```suggestion
Run `Juju.wait` using a standalone tool.
```

docs/how-to/index.md
```diff
@@ -28,3 +28,13 @@ Run Juju commands that aren't yet defined in Jubilant.
 
 Run Juju CLI commands <run-juju-cli-commands>
 ```
+
+## Using `Juju.wait` from the command line
```
dwilding, line 32: I like this title slightly better than the how-to guide title.

How about adjusting the how-to guide title to:

- Long form: How to use `Juju.wait` from the command line
- Short form: Use `Juju.wait` from the command line

Or an alternative I have in mind:

- Long form: How to use the Jubilant CLI to wait for conditions
- Short form: Use the Jubilant CLI to wait for conditions

But I kind of like yours better.

docs/conf.py
```diff
@@ -299,6 +299,7 @@
 intersphinx_mapping = {
     "juju": ("https://documentation.ubuntu.com/juju/3.6/", None),
     "operator": ("https://canonical.com/juju/docs/ops/latest/", None),
+    "python": ("https://docs.python.org/3", None),
```
dwilding, line 302: This doesn't look to be used.

```suggestion
```

dwilding (commented):
Nice, thanks for adding this!

My main recommendation is to introduce `ready` and `--error` as the main args before we start using them in examples.

docs/how-to/use-wait-from-cli.md
tromai: It is important if the reader is running this command right after the previous one in the guide:

```
$ jubilant wait 'jubilant.all_active(status)'

2026-09-08 01:31:43,040 [snappass-test] status: active (snappass started)
2026-09-08 01:31:43,040 [snappass-test/0] status: active (snappass started)
2026-09-08 01:31:45,454 Ready condition succeeded 3 times (jubilant.all_active(status))
```

I thought it might be confusing if `jubilant wait 'jubilant.all_active(status)'` would time out, while it previously worked fine.

However, I agree that we should use the same `all_active`. This is a sample to show the verbosity levels, so the confusion I tried to avoid doesn't exist.

docs/how-to/index.md
tromai: Thanks for the suggestions. 

I think I will go with `How to use Juju.wait from the command line`, to keep it consistent with other how-to titles:
- `How to use a custom wait condition`
- `How to run other Juju CLI commands`

docs/conf.py
tromai: Yes. It was a leftover from a previous version. Thanks for the catch.

docs/how-to/use-wait-from-cli.md
tromai: Thanks everyone for the feedback. I:
- Removed the `pyproject.toml project venv` part
- Start this section with `uv tool install jubilant`
- Mentioned upgrading and linking to uv's Tools page
- Collapse the `Ready and error expressions`
- Add `uvx` as a tip section.

docs/how-to/use-wait-from-cli.md
tromai: Agree. I have brought them to the final `jubilant wait CLI reference` section. I reused the help message itself since it describes all arguments of `jubilant wait` CLI. A `reference` section is a good idea!

docs/how-to/use-wait-from-cli.md
```diff
@@ -0,0 +1,160 @@
+---
+myst:
+  html_meta:
+    description: Use Juju.wait from the Command-Line Interface.
+---
+
```
tromai, line 6: I will need a paragraph explaining why this CLI exists.

docs/how-to/use-wait-from-cli.md
```diff
@@ -0,0 +1,160 @@
...
+General usage:
+
+```text
+jubilant wait
+    [--error ERROR]
+    ready
+```
+
+The `ready` and `--error` CLI arguments are passed as Python expressions, different from how [`Juju.wait`](jubilant.Juju.wait) is used in code. Those expressions have access to three variables: `jubilant` (the [`jubilant`](jubilant) module), `juju` (the [`jubilant.Juju`](jubilant.Juju) instance), and `status` (the [`jubilant.Status`](jubilant.Status) object).
```
james-garner-canonical, line 30: I think we should style user-supplied arguments (1) consistently, and (2) differently from literal stuff you type like `jubilant wait`. For example, `[--error <ERROR>] <READY>`.

And then "The `<READY>` and `<ERROR>` CLI arguments ..." or `The `READY` and `ERROR` CLI arguments ...".

Sorry if this was already litigated.

docs/how-to/use-wait-from-cli.md
```diff
@@ -0,0 +1,160 @@
...
+jubilant wait 'jubilant.all_active(status, "myapp")' \
+    --error 'jubilant.any_error(status)'
+```
+
+is equivalent to the following Python call:
+
+```python
+juju.wait(
+    lambda status: jubilant.all_active(status, 'myapp'),
+    error=jubilant.any_error,
```
james-garner-canonical, line 44: This is true but I think kind of confuses things coming as the first example of `error`.

```suggestion
    error=lambda status: jubilant.any_error(status),
```

If we wanted to, we could include a follow-up sentence or tip highlighting that `error` (and `ready`) are just callables accepting `status`, so you can write `error=jubilant.any_error` instead in the code.

docs/how-to/use-wait-from-cli.md
```diff
@@ -0,0 +1,160 @@
...
+    uvx jubilant wait --help
+```
+
+To upgrade the CLI:
+
+```
+uv tool upgrade jubilant
+```
+
+See more: [uv | Tools](https://docs.astral.sh/uv/concepts/tools)
```
james-garner-canonical, line 60: I'd be tempted to move this next to install rather than having install and <upgrade or run ephemerally> bracketing the usage instructions.

docs/how-to/use-wait-from-cli.md
```diff
@@ -0,0 +1,160 @@
...
++ .apps['snappass-test'].app_status.message = 'snappass started'
++ .apps['snappass-test'].units['snappass-test/0'].workload_status.current = 'active'
++ .apps['snappass-test'].units['snappass-test/0'].workload_status.message = 'snappass started'
++ .apps['snappass-test'].units['snappass-test/0'].juju_status.current = 'idle'
++ .apps['snappass-test'].units['snappass-test/0'].juju_status.version = '3.6.28'
++ .apps['snappass-test'].units['snappass-test/0'].leader = True
++ .apps['snappass-test'].units['snappass-test/0'].address = '10.1.0.215'
++ .apps['snappass-test'].units['snappass-test/0'].provider_id = 'snappass-test-0'
+2026-09-08 01:34:27,856 INFO jubilant.cli Ready condition succeeded 3 times (jubilant.all_active(status))
+```
```
james-garner-canonical, line 117: I'd suggest redacting a bunch of the middle lines with a single `...`.

docs/how-to/use-wait-from-cli.md
```diff
@@ -0,0 +1,160 @@
...
+jubilant wait
+    [-h]
+    [--delay DELAY]
+    [--error ERROR]
+    [--successes SUCCESSES]
+    [--timeout TIMEOUT]
+    [--juju-cli-bin JUJU_CLI_BIN]
+    [--model MODEL]
+    [--quiet | --verbose]
+    ready
```
james-garner-canonical, line 133: I see this is where the `ready` styling comes from. I think the CLI help gets away with it because of colour coding, so `ready` is clearly styled differently from `jubilant wait`.

Image

It would be nice if we could get styling like this in our docs, but since we go for more of a monochrome look, I wonder if we should style this differently.

james-garner-canonical (approved):
I did a read through and left some comments since my review was requested. Up to you which (if any) you want to address before merging.

docs/how-to/use-wait-from-cli.md
tromai: Thanks for this. I like your idea of `<ERROR>` and `<READY>` here. I have updated them. 

As for the `CLI reference` section at the bottom, I lean toward keeping it identical to the CLI usage message:

```
jubilant wait --help

<...>
```

which I think it's okay.

docs/how-to/use-wait-from-cli.md
tromai: Thanks for the suggestion. I lean towards keeping it similar to what we have [here](https://canonical.com/juju/docs/ops/latest/howto/write-integration-tests-for-a-charm/#example-of-verbose-logging).

docs/how-to/use-wait-from-cli.md
tromai: Answered in https://github.com/canonical/jubilant/pull/413#discussion_r4068288302

PR #416, opened 2026-09-14
feat: export ConstraintValue type alias publicly

### LINKED ISSUE #412: Expose `ConstraintValue` in `jubilant` module
This value is not exposed hence cannot be reused in helpers functions.
Interestingly `ConfigValue` is exposed so I guess this is a tiny miss.

I'll bring a PR for that soon.

Gu1nness:
Can be useful for methods wrapping deploys or set/update model constraints

fixes: #412


benhoyt (approved):
Looks reasonable to me, thanks.

jubilant/__init__.py
```diff
@@ -55,4 +56,4 @@
     'unittypes',
 ]
 
-__version__ = '1.13.0'
+__version__ = '1.14.0'
```
james-garner-canonical: With Jubilant we handle the version bump as part of the [release process](https://github.com/canonical/jubilant/blob/main/CONTRIBUTING.md#doing-a-release), and release monthly, so let's leave this at `1.13.0` for now.
```suggestion
__version__ = '1.13.0'
```

james-garner-canonical (approved):
This is reasonable: we already export the `ConfigValue` type alias, and `ConstraintValue` is part of the public API, being exposed in several public methods (like `add_machine`).

